PR #29, opened 2019-08-08
Added function that returns all wire names in wire diagram

mehalter:


epatters:
Thanks for the suggestion @mehalter.

As I understand it, this function returns the set of input and output port types for all boxes in the diagram. Because a port need not have an incident wire (representing a creation or deletion operation), this set is not necessarily the set of all wire types. Thus, I think the name `wirenames` is not quite right.

What is the use case for this function? My initial reaction is that this function may not be generally useful enough to include in the standard library, especially given how easy it is to implement.

mehalter:
What name do you think would be appropriate?

Also, I am working on a project that can convert scientific models represented as wiring diagrams into different types of model implementations (differential equations, agent based modeling, etc.) and this function kept coming up as useful. I don't know if it is generally useful, but figured I would suggest it since it kept popping up in my work.

epatters:
I appreciate the suggestion, but for now I see this function as not being generally useful enough.

That said, I'm pleased to hear you're using the package, and for such a cool project! I'd be interested to know how it's working out and any areas where improvement is needed (I know there are many).

Issue #33, opened 2019-08-15
Generate HTML documentation


epatters:
We don't have much long-form documentation yet, but the API docs, in the form of docstrings, aren't bad. I should use [https://github.com/JuliaDocs/Documenter.jl](Documenter.jl) to generate nice HTML documentation.

PR #34, opened 2019-08-15
Started creating Documenter.jl based documentation

mehalter:
I noticed you opened an issue about creating Documenter.jl documentation (#33). I figured I could help pick up some of the leg work to get started. All it contains is the Module APIs and the index contains the contents of the README so far.

I also added a travis job to build the documentation on pushes, and there is commented out code in the `doc/make.jl` that will enable pushing the generated HTML to a `gh_pages` branch.


epatters:
Thank you! This is very helpful.

One comment. Regular users of the package shouldn't have to install Documenter. We should be able to do this without adding Documenter to the project or manifest files. For example, LightGraphs seems to have done this ([.travis.yml](https://github.com/JuliaGraphs/LightGraphs.jl/blob/master/.travis.yml) and [Project.toml](https://github.com/JuliaGraphs/LightGraphs.jl/blob/master/Project.toml)).

mehalter:
Yeah that is a good point, I can make these changes.

mehalter:
Ok, these changes have been made

epatters:
Perfect, thanks. I will set up deployment of the docs later.

PR #114, opened 2020-03-05
Updated standard library, tests, and documentation to adopt the new where clause notation

mehalter:
All tests still pass when running `Pkg.test("Catlab")`


mehalter:
I also added the ability to store the axioms in the `Signature` struct to further pave the way to making them useful in the future. Also updated the relevant tests.

mehalter:
Fixed `TODO` to add `replace_types` functionality to the `AxiomConstructor` type.

epatters:
This is fantastic. Thanks!

PR #115, opened 2020-03-05
Allowed for local compilation of documentation and fixed broken links

bosonbaas:


docs/make.jl
```diff
@@ -8,12 +8,24 @@ const generated_dir = joinpath(@__DIR__, "src", "generated")
...
 @info "Building Literate.jl docs"
+
+# Define config if not being compiled on recognized service
+config = Dict{String,String}()
+env_set = haskey(ENV, "HAS_JOSH_K_SEAL_OF_APPROVAL") ||
+   haskey(ENV, "GITHUB_ACTIONS") ||
+   haskey(ENV, "GITLAB_CI")
+
+if !env_set
+    config["nbviewer_root_url"] = "https://nbviewer.jupyter.org/github/epatters/Catlab.jl/blob/gh-pages/v0.5.2"
```
epatters: Can this link point to a tag where the version is not hard-coded, something like `gh-pages` or `gh-pages/master`? Not sure whether these work.

bosonbaas: Replaced the static version number with dev

epatters:
Thanks! This is great.

PR #118, opened 2020-03-06
Updated signature definitions throughout the source, tests, and documentation

mehalter:


src/doctrines/Monoidal.jl
```diff
@@ -78,7 +75,8 @@ show_latex(io::IO, expr::ObExpr{:munit}; kw...) = print(io, "I")
 The signature (but not the axioms) is the same as a braided monoidal category.
 """
 @signature MonoidalCategory(Ob,Hom) => SymmetricMonoidalCategory(Ob,Hom) begin
-  braid(A::Ob, B::Ob)::Hom(otimes(A,B),otimes(B,A))
+  braid(A::Ob, B::Ob)::((A ⊗ B) → (B ⊗ A))
+  @op braid :σ
```
epatters: We need to export the Unicode operator for braiding. It would also be nice to add a unit test.

src/doctrines/Monoidal.jl
```diff
@@ -217,16 +213,12 @@ FIXME: This signature should extend both `MonoidalCategoryWithDiagonals` and
 supported.
 """
 @signature SymmetricMonoidalCategory(Ob,Hom) => MonoidalCategoryWithBidiagonals(Ob,Hom) begin
-  mcopy(A::Ob)::Hom(A,otimes(A,A))
-  mmerge(A::Ob)::Hom(otimes(A,A),A)
+  mcopy(A::Ob)::(A → (A ⊗ A))
```
epatters: This looks like an example of the problem I mentioned on Slack: the notation `@op mcopy Δ` should be defined (it is not inherited) but I guess that yields a warning.

src/doctrines/Monoidal.jl
mehalter: Fixed in commit 560bb3e

src/doctrines/Monoidal.jl
mehalter: I can show you when you get in, but this appears to be working as we want it to

Here is the code that I used to test this:
```
using Catlab
using Catlab.Doctrines
import Catlab.Doctrines: Ob, Hom

A = Ob(FreeCartesianCategory, :A)
B = Ob(FreeCartesianCategory, :B)
C = Ob(FreeCartesianCategory, :C)
f = Hom(:f, A,B)
g = Hom(:g, B,C)

f ⋅ g

Δ(A)

A′ = Ob(FreeBiproductCategory, :A)
B′ = Ob(FreeBiproductCategory, :B)
C′ = Ob(FreeBiproductCategory, :C)
f′ = Hom(:f, A′,B′)
g′ = Hom(:g, B′,C′)

f′ ⋅ g′

Δ(A′)

println(typeof(Δ(A)))
println(typeof(Δ(A′)))
```

epatters: It works, but only because the Unicode notation is omitted in MonoidalCategoryWithBidiagonals, when it should really be there! We can discuss.

epatters:
Thanks! I'll merge this now and we can resolve the issue with Julia functions for Unicode operators separately.

PR #122, opened 2020-03-06
Fixed the generation of alias functions

mehalter:
Alias functions now fully respect inheritance and are correctly typed.


epatters:
Thank you!

PR #123, opened 2020-03-06
GLA solvers

jpfairbanks:
add GLA iterative solver tests
fix adjoint in GLA doctrine


epatters:
The unit tests passed in the second commit, but I think that is a coincidence!

epatters:
Superseded by #124.

PR #124, opened 2020-03-07
GLA gmres test

jpfairbanks:
Added more tests and now they pass. Avoids the stochastic failure by only doing power method on a larger matrices where the P(failure) is smaller.


epatters:
Great, thanks!

Issue #62, opened 2019-12-13
From signatures to theories: allow specification of axioms


epatters:
Quote from the docs:

> At present, Catlab supports the specification of signatures, but not of axioms, reflecting its status as a programming library, not a proof assistant. It is the programmer's responsibility to ensure any declared instances of an algebraic structure satisfy its axioms.

It would be nice to allow users to write down the axioms. For now, we have no intention of enforcing them, but they would be helpful for documentation and knowledge representations purposes.

Following this, the macro `@signature` should probably be renamed to `@theory`.

epatters:
Thanks to @mehalter, we have a very good start on this issue. To finish it, we need to:

- Rename the `Signature` data type to `Theory` since it now stores axioms
- Create a new macro `@theory` that does what `@signature` now does
- Retain a macro called `@signature, both for backwards compatibility and to indicate that the axioms are not being recorded. IMO, this macro should raise an error if axioms are specified.
- Update the documentation to reflect these changes

epatters:
Resolved by #127. Thanks @mehalter!

PR #127, opened 2020-03-09
Move from @signature to @theory, and add @import to @instance macro

mehalter:
Relevant Issue (#62)

This PR moves the `@signature` macro to `@theory` now that axiom definition is supported. `@signature` is left to help with backwards compatibility, and also doesn't allow axiom definitions.

Also a `@import` call was added to the `@instance` macro to allow you to import external functions.

### DIFF
--- a/src/core/GAT.jl
+++ b/src/core/GAT.jl
@@ -590,18 +590,18 @@ the Typeclass.jl library for Julia.
 macro instance(head, body)
   # Parse the instance definition.
   head = parse_signature_binding(head)
-  functions, imports = parse_instance_body(body)
+  functions, ext_functions = parse_instance_body(body)
 
   # We must generate and evaluate the code at *run time* because the signature
   # module is not defined at *parse time*.
   # Also, we "throw away" any docstring.
   # FIXME: Is there a better place to put the docstring?
-  expr = :(instance_code($(esc(head.name)), $(esc(head.params)), $functions))
+  expr = :(instance_code($(esc(head.name)), $(esc(head.params)), $functions, $ext_functions))
   Expr(:block,
     Expr(:call, esc(:eval), expr),
     :(Core.@__doc__ abstract type $(esc(gensym(:instance_doc))) end)) # /dev/null
 end
-function instance_code(mod, instance_types, instance_fns)
+function instance_code(mod, instance_types, instance_fns, external_fns)
   code = Expr(:block)
   class = mod.class()
   bindings = Dict(zip(class.type_params, instance_types))
@@ -628,20 +628,20 @@ end
 function parse_instance_body(expr::Expr)
   @assert expr.head == :block
   funs = JuliaFunction[]
-  imports = Symbol[]
+  ext_funs = Symbol[]
   for elem in strip_lines(expr).args
     elem = strip_lines(elem)
     head = elem.head
     if head == :macrocall && elem.args[1] == Symbol("@import")
       @match elem.args[2] begin
-        sym::Symbol => push!(imports, sym)
-        Expr(:tuple, arr) => (imports = [imports; Symbol[arr...]])
+        sym::Symbol => push!(ext_funs, sym)
+        Expr(:tuple, arr) => (ext_funs = [ext_funs; Symbol[arr...]])
       end
     else
       push!(funs, parse_function(elem))
     end
   end
-  return (funs, imports)
+  return (funs, ext_funs)
 end
 
 """ Invoke a term constructor by name on an instance.
--- a/src/core/GAT.jl
+++ b/src/core/GAT.jl
@@ -613,10 +613,10 @@ function instance_code(mod, instance_types, instance_fns, external_fns)
       f_impl = instance_fns[sig]
     elseif !isnothing(f.impl)
       f_impl = f
-    else
-      # FIXME: Can we check whether the method has already been defined?
-      @warn "Method $(f.call_expr) not implemented in $(class.name) instance"
+    elseif f.call_expr.args[1] in external_fns
       continue
+    else
+      @error "Method $(f.call_expr) not implemented in $(class.name) instance"
     end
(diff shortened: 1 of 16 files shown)


src/core/GAT.jl
```diff
@@ -69,109 +69,147 @@ end
...
-1. Type constructors, indicated by the special type `TYPE`, e.g.
+1. Type constructors, indicated by the special type `TYPE`, e.g.,
    `Hom(X::Ob,Y::Ob)::TYPE`
 2. Term constructors, e.g.,
    `id(X::Ob)::Hom(X,X)`
 3. Julia functions operating on the term constructors to provide additional
    functionality
+4. Function aliases, e.g.,
+   `@op Hom :→`
+5. Equivalency axioms, e.g.,
```
epatters: Let's call these "equality axioms" rather than "equivalency axioms". That would be the more standard term.

src/core/GAT.jl
```diff
@@ -69,109 +69,147 @@ end
...
+
+A theory can extend existing theories (at present only one).
+"""
+macro theory(head, body)
+  theory_builder(head, body)
+end
+
+""" Define a signature for a generalized algebraic theory (GAT).
+
+A signature is the same as a theory, except may not contain axioms, and
```
epatters: Small typo: "except it may not"

src/core/GAT.jl
```diff
@@ -589,34 +627,34 @@ the Typeclass.jl library for Julia.
...
+      f_impl = instance_fns[theory]
     elseif !isnothing(f.impl)
       f_impl = f
-    else
-      # FIXME: Can we check whether the method has already been defined?
-      @warn "Method $(f.call_expr) not implemented in $(class.name) instance"
+    elseif f.call_expr.args[1] in external_fns
       continue
+    else
+      @error "Method $(f.call_expr) not implemented in $(class.name) instance"
```
epatters: Rather than logging an error using the `Logging` module, I think we should actually raise an exception. This was the original behavior, which I changed to a warning as a temporary workaround.

epatters:
This looks great. I made a few minor comments above.

Regarding when to use `@theory` vs `@signature`, I propose that we adopt the policy of using `@theory` only when we actually write down the axioms. Then writing `@signature` will serve as a warning that the theory hasn't been fully specified and as a reminder to eventually write down the axioms. What do you think? Any thoughts @jpfairbanks?

src/core/GAT.jl
mehalter: Fixed in commit 1a3a27c

src/core/GAT.jl
mehalter: Fixed in commit 1a3a27c

src/core/GAT.jl
mehalter: Fixed in commit 1a3a27c

mehalter:
How do you want to enforce this? Or just a policy for best practices? This is what I was wondering when I asked you if we should check if a theory definition doesn't have any axioms defined.

mehalter:
For now, I haven't added this error throwing, but have moved all of our doctrine definitions that don't have axioms to `@signature`

src/core/GAT.jl
```diff
@@ -647,12 +647,12 @@ function instance_code(mod, instance_types, instance_fns, external_fns)
   bound_fns = OrderedDict(parse_function_sig(f) => f for f in bound_fns)
   instance_fns = Dict(parse_function_sig(f) => f for f in instance_fns)
   for (theory, f) in bound_fns
```
epatters: I didn't catch this the first time, but `theory` here should have been kept as `sig` because the object is the signature of a Julia function not the signature of a GAT.

src/linear_algebra/GLA.jl
```diff
@@ -29,25 +29,48 @@ Functional fragment of graphical linear algebra.
...
+  Δ(A) == Δ(A) ⋅ σ(A, A) ⊣ (A::Ob)
+  Δ(A) ⋅ (Δ(A) ⊗ id(A)) == Δ(A) ⋅ (id(A) ⊗ Δ(A)) ⊣ (A::Ob)
+  Δ(A) ⋅ (◊(A) ⊗ id(A)) == id(A) ⊣ (A::Ob)
+  ⊞(A) == σ(A, A) ⋅ ⊞(A) ⊣ (A::Ob)
+  (⊞(A) ⊗ id(A)) ⋅ ⊞(A) == (id(A) ⊗ ⊞(A)) ⋅ ⊞(A) ⊣ (A::Ob)
+  (mzero(A) ⊗ id(A)) ⋅ ⊞(A) == id(A) ⊣ (A::Ob)
+  ⊞(A) ⋅ Δ(A) == ((Δ(A) ⊗ Δ(A)) ⋅ (id(A) ⊗ (σ(A, A) ⊗ id(A)))) ⋅ (⊞(A) ⊗ ⊞(A)) ⊣ (A::Ob)
+  ⊞(A) ⋅ ◊(A) == ◊(A) ⊗ ◊(A) ⊣ (A::Ob)
+  mzero(A) ⋅ Δ(A) == mzero(A) ⊗ mzero(A) ⊣ (A::Ob)
+  mzero(A) ⋅ ◊(A) == munit() ⊣ (A::Ob)
```
epatters: RHS should be `◊(mzero())`.

src/core/GAT.jl
mehalter: Fixed

src/linear_algebra/GLA.jl
mehalter: Fixed in commit 011e9d2

epatters:
Looks great, thanks. I'll merge.

I don't think we should throw an error when a theory has no axioms. This is valid in rare cases and, more importantly, there is no way to check the more common but more subtle error of listing axioms but forgetting one.

PR #128, opened 2020-03-10
Overload `plus` function in GLA

epatters:
This PR does two things:

1. Renames `ozero` to `mzero` in doctrine for additive monoidal categories. So now we have (`otimes` and `munit`) and (`oplus` and `mzero`), which should be easy to remember.

2. Renames `mplus` to `plus` and `mzero` to `zero` in abelian bicategories of relations and in GLA, overloading `plus` in GLA and `zero` in base Julia.


jpfairbanks:
I like using multiple dispatch to harmonize our notation with regular math notation, but the use of +(A) to mean mplus(A) is really disturbing as a unary operator. It’s because mplus(A) is the operator (x,y) maps to x+y. Seeing +(A) is just weird. But I guess I could get used to it

epatters:
OK, how about I revert the second commit and we just write `plus(A)` instead of `+(A)`?

mehalter:
Yeah I like this, and agree we should do `plus(A)` and not `+(A)`

PR #129, opened 2020-03-10
Added ability to define operation aliases in block form

mehalter:
Added a second notation for `@op` macro inside of the `@theory`

```
@op begin
  Hom :→
  compose :⋅
end
```

Also updated tests to make sure both notations work, and found a typo/bug in one of the ParseError definitions

### DIFF
--- a/docs/src/apis/core.md
+++ b/docs/src/apis/core.md
@@ -91,7 +91,7 @@ discussion, see Cartmell, 1986, Sec 10: Informal syntax).
 
 Notice the `@op` call where we can create method aliases that can then be used
 throughout the rest of the theory and outside of definition. We can either use
-this block notation, or a single line notation such as `@op compose :·` to
+this block notation, or a single line notation such as `@op (·) := compose` to
 define a single alias. Here we utilize this functionality by replacing the `Hom`
 and `compose` methods with their equivalent unicode characters, `→` and `·`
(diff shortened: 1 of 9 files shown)


jpfairbanks:
Cool! What about doing the syntax like 
```
@op begin
  (→) := Hom
  (⋅) := compose
end
```

That makes more sense from a math notation perspective.

mehalter:
I just made a change that will currently allow you to define them in the notation that James brought up, or the same notation that is used for the single line definitions. Once we decide on a specific version to use, I can remove the other in the `if` statement. We can agree to use both, but I think I would prefer to settle on a single notation.

epatters:
I would also strongly prefer to settle on a single notation. I think I slightly prefer James' notation, but here I don't feel as strongly.

jpfairbanks:
Yeah I think the `Hom :+` is ooogly. let's just do the `:=` notation. and then we just forbid using `:=` as a unicode name for a Gat term.

mehalter:
Cool I think this PR is ready once the tests pass.

epatters:
For consistency, shouldn't  we use the same syntax in the short form of `@op`, e.g., `@op (→) := Hom` instead of `@op Hom :→`?

jpfairbanks:
Yeah I think that is more consistent

jpfairbanks:
New code looks good to me. I like this interface a lot

Issue #112, opened 2020-03-04
No Julia functions inside GAT signatures


epatters:
Since I originally thought of signatures like Haskell type classes, arbitrary Julia function definitions are allowed in signatures of GATs, e.g., for defining Unicode syntax. I would like to ban this and draw a clear distinction between the algebraic theories and the Julia code that instantiates them. This would be a first step towards a better internal representation of GATs (cf. #15).

jpfairbanks:
Is this closed by the recent work on the Unicode syntax such as #129?

epatters:
Not yet, although the groundwork has been laid. At this point, I think @mehalter could resolve this issue fairly quickly.

epatters:
Here is some background info and a suggested approach, based on issues that Micah noticed in his PR #130.

The original intent behind allowing function definitions in signatures, besides supporting Unicode operators, was to allow certain term constructors to have default implementations via other terms. For example, in the definition of `CompactClosedCategory`, we currently see this:

```julia
hom(A::Ob, B::Ob) = B ⊗ dual(A)
ev(A::Ob, B::Ob) = id(B) ⊗ (σ(dual(A), A) ⋅ dcounit(A))
curry(A::Ob, B::Ob, f::Hom) = ((id(A) ⊗ (dunit(B) ⋅ σ(dual(B), B))) ⋅ (f ⊗ id(dual(B))))
```

This expresses the fact that the closed monoidal structure of a compact closed category is derived from its dual objects. However, as argued above, this conflates the algebraic theory with its Julia implementation. Thanks to Micah, we can now record axioms, so the function definitions above can be replaced with equations:

```julia
(hom(A,B) == B ⊗ dual(A)) where (A::Ob, B::Ob)
(ev(A,B) == id(B) ⊗ (σ(dual(A), A) ⋅ dcounit(A))) where (A::Ob, B::Ob)
(curry(A, B, f) == ((id(A) ⊗ (dunit(B) ⋅ σ(dual(B), B))) ⋅ (f ⊗ id(dual(B))))) where (A::Ob, B::Ob, C::Ob, f::Hom(A⊗B,C))
```

Another reason that such definitions should be handled by axioms is that sometimes there are sets of terms that can each be defined via the others. For example, in `CartesianCategory`, we have

```julia
pair(f,g) == compose(mcopy(A), otimes(f,g)) where (A::Ob, B::Ob, C::Ob, f::Hom(A,B), g::Hom(A,C))
[...]

PR #130, opened 2020-03-11
Remove functions from signature/theory definitions

mehalter:
Solves issue #112

### DIFF
--- a/src/doctrines/Monoidal.jl
+++ b/src/doctrines/Monoidal.jl
@@ -262,12 +262,6 @@ end
 
   # Adjoint mate of morphism f.
   mate(f::(A → B))::(dual(B) → dual(A)) ⊣ (A::Ob, B::Ob)
-
-  # Closed monoidal category
-  hom(A::Ob, B::Ob) = B ⊗ dual(A)
-  ev(A::Ob, B::Ob) = id(B) ⊗ (σ(dual(A), A) ⋅ dcounit(A))
-  curry(A::Ob, B::Ob, f::Hom) =
-    ((id(A) ⊗ (dunit(B) ⋅ σ(dual(B), B))) ⋅ (f ⊗ id(dual(B))))
 end
 
 @syntax FreeCompactClosedCategory(ObExpr,HomExpr) CompactClosedCategory begin
--- a/src/doctrines/Monoidal.jl
+++ b/src/doctrines/Monoidal.jl
@@ -115,10 +115,14 @@ end
 Actually, this is a cartesian *symmetric monoidal* category but we omit these
 qualifiers for brevity.
 """
-@signature MonoidalCategoryWithDiagonals(Ob,Hom) => CartesianCategory(Ob,Hom) begin
+@theory MonoidalCategoryWithDiagonals(Ob,Hom) => CartesianCategory(Ob,Hom) begin
   pair(f::(A → B), g::(A → C))::(A → (B ⊗ C)) ⊣ (A::Ob, B::Ob, C::Ob)
   proj1(A::Ob, B::Ob)::((A ⊗ B) → A)
   proj2(A::Ob, B::Ob)::((A ⊗ B) → B)
+
+  pair(f, g) == Δ(A) ⋅ (f ⊗ g) ⊣ (A::Ob, B::Ob, C::Ob, f::(A → B), g::(A → C))
+  proj1(A, B) == id(A) ⊗ ◊(B) ⊣ (A::Ob, B::Ob)
+  proj2(A, B) == ◊(A) ⊗ id(B) ⊣ (A::Ob, B::Ob)
 end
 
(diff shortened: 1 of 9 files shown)


src/experimental/AlgebraicNets.jl
```diff
@@ -34,7 +34,6 @@ TODO: Explain
 """
 @signature MonoidalCategoryWithBidiagonals(Ob,Hom) => AlgebraicNetTheory(Ob,Hom) begin
   linear(x::Any, A::Ob, B::Ob)::(A → B)
-  constant(x::Any, A::Ob) = (x → munit(Ob) → A)
```
jpfairbanks: This seems really important, we need the ability to add constants into our morphisms. Should this be a definition outside the `@signiture`?

mehalter: Is constant here a type or a term? I don't know what this is supposed to be exactly. I think it needs to be rewritten as a term constructor

jpfairbanks: Yeah I think there should be a generally better way to handle constants. the whole constant(A) = Hom(munit(Ob), A) is a real pain when declaring the axioms. Especially if we have more complex constants like the Vector thing in structured matrices. @epatters, can we do `constant(A)::Hom(munit(Ob), A)` in the theory as a term constructor and then handle them in the syntax and instances?

epatters: The code in that lines seems to have gotten mangled during our earlier Unicode refactors. I guess I didn't have any tests to catch it :(

Anyway, yes, there should be a term constructor ``constant(A::Ob, x)::Hom(munit(), A)`` where `x` is the constant value. It can then be implemented in instances and syntax systems. We're already using this device for `scalar` terms in GLA.

epatters: Actually, here we have ``constant(x::Any, A::Ob)::Hom(munit(), A)`` but the order isn't so important.

epatters:
I wrote a long note in #112 about how to think about `hom`/`ev`/`curry` and related situations.

I think the only remaining problem is how to handle the Unicode `\circ`. Unfortunately, as Micah noticed, Base defines an untyped function, despite it only working for Julia callables (`Function` and I guess `Type`). So, unless we get rid of it, our definition will have to be typed.

Only possible solution is to allow a variant of `@op` or argument to `@op`, called something like `reverse`, that reverses the arguments. Thoughts on this?

jpfairbanks:
Could we handle the `\circ` by typing the function as `\circ(args::Varargs{GAT.Stub}) = compose(reverse(args)...)`? That would allow it to work on any combination of GAT terms.

For example:
```julia
julia> supertype(LinearFunctions.Ob)
Catlab.GAT.Stub

julia> supertype(LinearFunctions.Hom)
Catlab.GAT.Stub
```

src/experimental/AlgebraicNets.jl
mehalter: Ah, that's what I thought. I have fixed it. For this entry, is there a function definition we could put in the instance `AlgebraicNetTheory`, because now it is saying there needs to be a defintion for the function `constant`

epatters:
With a tweak to the type hierarchy for GAT expressions, that would work for signatures and syntax, but it won't work for instances, whose type hierarchy we cannot control.

src/experimental/AlgebraicNets.jl
epatters: Try something like:

```julia
function constant(x::Any, A::NFormulas)::Formulas
  @assert A.n == 1
  Formulas([x], Symbol[])
end
```

I think this only ever worked for "scalar" constants.

src/doctrines/Monoidal.jl
```diff
@@ -263,11 +267,10 @@ end
   # Adjoint mate of morphism f.
   mate(f::(A → B))::(dual(B) → dual(A)) ⊣ (A::Ob, B::Ob)
 
-  # Closed monoidal category
```
epatters: In order for these axioms to make sense, we need corresponding term constructors.

test/doctrines/Monoidal.jl
```diff
@@ -134,12 +134,6 @@ f, g = Hom(:f, A, B), Hom(:g, B, A)
 @test dom(mate(f)) == dual(B)
 @test codom(mate(f)) == dual(A)
 
-f = Hom(:f, otimes(A,B), C)
```
epatters: With appropriate term constructors added, some version of these tests should still pass.

mehalter: fixed

src/doctrines/Monoidal.jl
mehalter: fixed

epatters:
Looks great, thanks @mehalter! The failing doc action looks like a GitHub bug.

Issue #84, opened 2020-01-16
Thin categories and preorders


epatters:
On Slack, @kris-brown asks how to do preorders in Catlab. In fact, it would be good to add signatures and syntax systems for preorders and also thin categories to the standard library.

epatters:
Ideally, [thin categories](https://ncatlab.org/nlab/show/thin+category) would be expressed as:

```julia
@theory Category(Ob,Hom) => ThinCategory(Ob,Hom) begin
  (f == g) <= (A::Ob, B::Ob, f::Hom(A,B), g::Hom(A,B))
end
```

However, we cannot yet write this because `@theory` does not exist and `@signature` does not accept axioms (#62). So for now we'd have to write:

```julia
@signature Category(Ob,Hom) => ThinCategory(Ob,Hom) begin
  # (f == g) <= (A::Ob, B::Ob, f::Hom(A,B), g::Hom(A,B))
end
```

epatters:
Preorders should be expressed in the usual terminology, something like:

```julia
@signature Preorder(Ob,Leq) begin
  Ob::TYPE
  Leq(lhs::Ob, rhs::Ob)::TYPE

  reflexive(A::Ob)::Leq(A,A)
  transitive(::Leq(A,B), ::Leq(B,C))::Leq(A,C) <= (A::Ob, B::Ob, C::Ob)
end
```

Note the level shift here: what are considered *axioms* of preorders become *operations* in the generalized algebraic theory.

We should also add syntactic sugar for `\leq` and implement a syntax system `FreePreorder` that normalizes all `Leq`s on the same objects to be equal.

jpfairbanks:
That makes sense to me, but can you explain the part about the level shift? Why do axioms of the preorders become operations in the GAT?

epatters:
Usually, we think of preorders as being a set equipped with a binary relation `<=` (a boolean-valued operation, if you like) satisfying two axioms: reflexivity and transitivity.

But if we want to represent preorders as generalized or essentially algebraic theories, everything has to shift a level downwards. The binary relation `<=` (the "operation") becomes a dependent type (here `Leq`), the reflexivity and transitivity axioms becomes operations, and a new axiom is introduced saying that any two terms of type `Leq(A,B)` are equal. This agrees with the understanding of thin categories as preorders: the composition operation is transitivity and the identity operation is reflexivity. (The associativity and unitality axioms are automatically satisfied due to the thinness axiom, so we don't even have to think about them.)

[...]

PR #132, opened 2020-03-12
Add Preorder and Thin Category Implementation

jpfairbanks:
Addresses #84 by defining a Preorder GAT and Thin Categories GAT. Provides MonoidalThinCategory and appropriate syntaxes

### DIFF
--- a/src/doctrines/Preorders.jl
+++ b/src/doctrines/Preorders.jl
@@ -1,0 +1,60 @@
+export Preorder, ThinCategory, MonoidalThinCategory, FreePreorder,
+  FreeThinCategory, FreeMononoidalThinCategory
+
+import Catlab.Doctrines: Ob, Hom, otimes, compose, ⋅, ⊗
+
+""" Doctrine of *Thin category*
+
+Thin categories have at most one Hom between any two objects.
+"""
+@theory Category(Ob,Hom) => ThinCategory(Ob,Hom) begin
+  f == g ⊣ (A::Ob, B::Ob, f::Hom(A,B), g::Hom(A,B))
+end
+
+@syntax FreeThinCategory(ObExpr,HomExpr) ThinCategory begin
+  compose(f::Hom, g::Hom) = associate(new(f,g; strict=true))
+end
+
+""" Doctrine of *Preorder*
+
+Preorders encode the axioms of reflexivity and transitivity as term constructors.
+"""
+@theory Preorder(Ob,Leq) begin
+  # METATYPES of the GAT
+  Ob::TYPE
+  Leq(lhs::Ob, rhs::Ob)::TYPE
+  @op (≤) := Leq
+  # Axioms of the mathematical object are lifted to term constructors in the GAT
+  reflexive(A::Ob)::(A≤A) # ∀ A there is a term reflexive(A) which implies A≤A
+  transitive(f::(A≤B), g::(B≤C))::(A≤C) ⊣ (A::Ob, B::Ob, C::Ob)
+
+  # axioms of the GAT are equivalences on terms or simplification rules in the logic
+  f == g ⊣ (A::Ob, B::Ob, f::(A≤B), g::(A≤B))
+  # read as (f⟹ A≤B ∧ g⟹ A≤B) ⟹ f ≡ g
+end
+
+# a GAT-homomorphism between the Preorder GAT and the ThinCategory GAT
+# this is a morphism is *Gat* the category whose objects are doctrines
+# and whose morphisms are algebraic structure preserving maps
+
+# @functor F(Preorder(Ob, Leq))::ThinCategory(Ob, Hom) begin
+#   Ob ↦ Ob
+#   Leq ↦ Hom
+#   reflexive ↦ id
+#   transitive ↦ compose
+# end
+
+
+@syntax FreePreorder(ObExpr,HomExpr) Preorder begin
+  transitive(f::Leq, g::Leq) = associate(new(f,g; strict=true))
+end
+
+@theory MonoidalCategory(Ob,Hom) => MonoidalThinCategory(Ob,Hom) begin
+  f == g ⊣ (A::Ob, B::Ob, f::Hom(A,B), g::Hom(A,B))
+end
+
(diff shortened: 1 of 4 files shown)


test/doctrines/Preorders.jl
```diff
@@ -0,0 +1,52 @@
...
+#   # axioms of the GAT are equivalences on terms or simplification rules in the logic
+#   f == g ⊣ (A::Ob, B::Ob, f::(A≤B), g::(A≤B))
+#   # read as (f⟹ A≤B ∧ g⟹ A≤B) ⟹ f ≡ g
+# end
+#
+# @syntax FreePreorder(ObExpr,HomExpr) Preorder begin
+#   transitive(f::Leq, g::Leq) = associate(new(f,g; strict=true))
+# end
+#
+# x,y,z = Elt(FreePreorder, [:x, :y, :z])
```
epatters: I agree that `Elt` would be better.

The reason this doesn't work is that the generated code only allows you to write `Elt(FreePreorder.Elt, :x)`. Because that gets hold, I added special convenience methods for `Ob` [here](https://github.com/epatters/Catlab.jl/blob/e500fa203d7b3dbd57fc545afdea8c7f91f01606/src/doctrines/Doctrines.jl#L13). We could do something similar for Elt.

src/doctrines/Preorders.jl
```diff
@@ -0,0 +1,60 @@
+export Preorder, ThinCategory, MonoidalThinCategory, FreePreorder,
+  FreeThinCategory, FreeMononoidalThinCategory
+
+import Catlab.Doctrines: Ob, Hom, otimes, compose, ⋅, ⊗
```
epatters: For once we don't need these imports, because all the doctrine files get included into a single module.

src/doctrines/Preorders.jl
```diff
@@ -0,0 +1,60 @@
+export Preorder, ThinCategory, MonoidalThinCategory, FreePreorder,
+  FreeThinCategory, FreeMononoidalThinCategory
```
epatters: We should probably export `Elt`, `Leq`, `≤`, `reflexive`, `transitive`.

src/doctrines/Preorders.jl
```diff
@@ -0,0 +1,60 @@
...
+  # axioms of the GAT are equivalences on terms or simplification rules in the logic
+  f == g ⊣ (A::Ob, B::Ob, f::(A≤B), g::(A≤B))
+  # read as (f⟹ A≤B ∧ g⟹ A≤B) ⟹ f ≡ g
+end
+
+# a GAT-homomorphism between the Preorder GAT and the ThinCategory GAT
+# this is a morphism is *Gat* the category whose objects are doctrines
+# and whose morphisms are algebraic structure preserving maps
+
+# @functor F(Preorder(Ob, Leq))::ThinCategory(Ob, Hom) begin
```
epatters: It would be very cool to write this!

src/doctrines/Preorders.jl
```diff
@@ -0,0 +1,60 @@
...
+#   reflexive ↦ id
+#   transitive ↦ compose
+# end
+
+
+@syntax FreePreorder(ObExpr,HomExpr) Preorder begin
+  transitive(f::Leq, g::Leq) = associate(new(f,g; strict=true))
+end
+
+@theory MonoidalCategory(Ob,Hom) => MonoidalThinCategory(Ob,Hom) begin
```
epatters: Extend `SymmetricMonoidalCategory` instead of `MonoidalCategory`? That's usually what we want.

test/doctrines/Preorders.jl
jpfairbanks: Should the theory macro generate those functions for every metatype in the GAT?

src/doctrines/Preorders.jl
jpfairbanks: Should it be called SymmetricMonoidalThinCategory or just MonoidalThinCategory?

test/doctrines/Preorders.jl
epatters: Arguably, but note that it only makes sense for nullary type constructors like `Ob` or `Elt`. It wouldn't make sense for `Hom` or `Leq`.

src/doctrines/Preorders.jl
epatters: If we're being precise, "symmetric" should probably be in there. Maybe `ThinSymmetricMonoidalCategory`?

test/doctrines/Preorders.jl
```diff
@@ -0,0 +1,52 @@
...
+x,y,z = Ob(FreeThinCategory, :x, :y, :z)
+f, g = Hom(:f, x, y), Hom(:g, y, z)
+@test dom(compose(f,g)) == x
+@test codom(compose(f,g)) == z
+
+
+x,y,z = Ob(FreePreorder, :x, :y, :z)
+@test_throws UndefVarError f, g = Leq(:f, x, y), Leq(:g, y, z)
+# Preorder doesn't have domain and codomain defined.
+@test_throws UndefVarError dom(transitive(f,g)) == x
```
epatters: Because `Leq` is defined as `Leq(lhs::Elt, rhs::Elt)`, the accessors are `lhs` and `rhs`, not `dom` and `codom`. So `lhs(transitive(f,g)) == x` and `rhs(transitive(f,g)) == z` should work.

That reminds me: we should also export `lhs` and `rhs`.

test/doctrines/Preorders.jl
jpfairbanks: Why don't you need to do this for `Hom` or `Leq`?

epatters: A constructor like `Hom(FreeCategory, :f, :g, :h)` doesn't seem to make sense. What would the domain and codomain be? I suppose we could introduce a shorthand like `Hom([:f, :g, :h], A, B)` to make three generators all with the same domain `A` and codomain `B`.

epatters:
I think this PR is quite close. Would you like me to finish it up? I could probably do it pretty quickly.

epatters:
Thanks for doing most of this. I've taken the liberty of finishing it in #148.

PR #137, opened 2020-03-14
Update index.md with history of GATs

jpfairbanks:
add snippet from slack explaining the history of GATs

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -54,6 +54,19 @@ or other graphical user interface. It is primarily a programming library, not a
 user-facing application. However, it could be used as the backend for such an
 application.
 
+### What is a GAT?
+Generalized Algebraic Theories (GATs) are the backbone of Catlab so let's expand a bit on GATs and how they fit into the bigger picture of algebra.
+An algebraic structure, like a group or category, is a mathematical object whose axioms all take the form of equations that are universally quantified (the equations have no exceptions). That’s not a formal definition but it’s a good heuristic. There are different ways to make this precise. The oldest, going back to universal algebra in the early 20th centrury, are algebraic theories.
+
+[Universal Algebra](https://en.wikipedia.org/wiki/Universal_algebra) (sometimes called general algebra) is the field of mathematics that studies algebraic structures themselves, not examples ("models") of algebraic structures.
+For instance, rather than take particular groups as the object of study, in universal algebra one takes the class of groups as an object of study. In an algebraic theory, you have a collection of (total) operations and they obey a set of equational axioms. Classically, there is only a single generating type, but there are also typed or multi-sorted versions of algebraic theories. Most of the classical structures of abstract algebra, such as groups, rings, and modules, can be defined as algebraic theories.
+
+Importantly, a category cannot be defined as a algebraic theory. The reason is that the operation of composition is partial, since you can only compose morphisms with compatible (co)domains. Now, categories sure feel like algebraic structures, so people have come up with generalizations of algebraic theories that accomodate categories and related structures.
+
+The first of these was Freyd’s essentially algebraic theories. In an essentially algebraic theory, you can have partially defined operations; however, to maintain the equational character of the system, the domains of operations must themselves be defined equationally. For example, the theory of category would be defined as having two types, Ob and Hom, and the composition operation `compose(f::Hom,g::Hom)::Hom` would have domain given by the equation `codom(f) == dom(g)`. As your theories get more elaborate, the sets of equations defining the domains get more complicated and reasoning about the structure is overwhelming.
+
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -54,6 +54,19 @@ or other graphical user interface. It is primarily a programming library, not a
...
+
+[Universal Algebra](https://en.wikipedia.org/wiki/Universal_algebra) (sometimes called general algebra) is the field of mathematics that studies algebraic structures themselves, not examples ("models") of algebraic structures.
+For instance, rather than take particular groups as the object of study, in universal algebra one takes the class of groups as an object of study. In an algebraic theory, you have a collection of (total) operations and they obey a set of equational axioms. Classically, there is only a single generating type, but there are also typed or multi-sorted versions of algebraic theories. Most of the classical structures of abstract algebra, such as groups, rings, and modules, can be defined as algebraic theories.
+
+Importantly, a category cannot be defined as a algebraic theory. The reason is that the operation of composition is partial, since you can only compose morphisms with compatible (co)domains. Now, categories sure feel like algebraic structures, so people have come up with generalizations of algebraic theories that accomodate categories and related structures.
```
varkor: > Importantly, a category cannot be defined as a algebraic theory.

Is this known? It might be better to phrase this as "Importantly, there is no known way to define a category as algebraic theory." (or similar). You can often encode a surprising amount of structure into simpler systems (c.f. the simply-typed lambda calculus and the typed SK-calculus). It's not entirely implausible that some dependent algebraic structures could be encoded as (multisorted) algebraic structures.

docs/src/index.md
```diff
@@ -54,6 +54,19 @@ or other graphical user interface. It is primarily a programming library, not a
...
+
+Later, Cartmell proposed generalized algebraic theories, which solves the same problem but in a different way. Rather than having partial operations, you have total operations but on dependent types (types that are parameterized by values). So now the composition operation has signature `compose(f::Hom(A,B), g::Hom(B,C))::Hom(A,C) where (A::Ob, B::Ob, C::Ob)`  exactly as appears in Catlab. This is closer to the way that mathematicians actually think and write about categories. For example, if you look at the definitions of category, functor, and natural transformation in Emily Riehl’s textbook, you will see that they are already essentially in the form of a GAT, whereas they require translation into an essentially algebraic theory. Nevertheless, GATs and essentially algebraic theories have the same expressive power, at least in their standard set-based semantics. GATs provide a version of the computer scientist's type theory that plays well with the mathematician's algebra, thus, providing a perfect opportunity for computer algebra systems.
```
varkor:
```suggestion
Nevertheless, GATs and essentially algebraic theories have the same expressive power, at least in their standard set-based semantics.
```
Although this claim (which is originally from Cartmell's thesis) is widely accepted as folklore, it's actually a dubious one. See, for instance, [Voevodsky's comments on the subject](https://www.math.ias.edu/Voevodsky/files/files-annotated/Dropbox/Unfinished_papers/Dynamic_logic/Stage_current/2013_03_05_TTS.pdf). I think it'd be worth simply excluding this comment.

varkor:
Other than those comments, this seems like a nice introduction :)

docs/src/index.md
jpfairbanks: Since this is not a research article, I think this is fine. I'll defer to @epatters for his judgement.

docs/src/index.md
jpfairbanks: Since this is not a research article, I think this is fine. I'll defer to @epatters for his judgement.

docs/src/index.md
epatters: I'm pretty sure this is true. I know I've seen it claimed in various places. For example, in their introduction to [this paper](https://www.cs.bham.ac.uk/~sjv/partialalgebras.pdf), Vickers and Palmgren state that "the theory of categories is not algebraic", by which they mean that it cannot be construed as a multi-sorted algebraic theory. Unfortunately they don't give a proof or a reference. The way you would prove such a thing is to show that any category of models of an algebraic theory has a certain property that the category of categories fails to possess. For example, this is how you prove that the [theory of fields is not algebraic](https://mathoverflow.net/questions/3003/in-what-sense-are-fields-an-algebraic-theory) (!).

Anyway, if you find a good reference on this, please let us know.

docs/src/index.md
epatters: The claim also appears in Cartmell's 1986 paper with a sketch of a proof. I wasn't aware of Voevodsky's worry, but if I'm reading him correctly, he is worried about whether generalized algebraic and essentially algebraic theories are in 1-1 correspondence, not whether they generate the same class of set-theoretic models (which is the claim asserted above). Thanks for the reference.

epatters:
Thanks a lot James. I may follow up later with references for readers who really want to dig into this stuff, but I think this is a great start.

docs/src/index.md
varkor: Ah, you're right; I had misremembered the concern. Cartmell's sketch is a little too unrigorous for my liking, but I should probably just take that as a prompt to write it out in a bit more detail to convince myself :) Thanks.

docs/src/index.md
varkor: I looked around for a reference, but couldn't find one, so I asked on mathoverflow and received [an elegant explanation](https://mathoverflow.net/a/354926/152679).

epatters: Very nice! I'll update the docs with a link to this answer.

Issue #125, opened 2020-03-07
LinearMaps isomorphic to LinearOperators


jpfairbanks:
We could use LinearOperators.jl as the instance type for linear maps. This would enable the methods in Krylov.jl instead of IterativeSolvers.jl. Showing compatibility between them would be cool. We have a backend agnostic representations of linear operators (frontend language) with two instances that play the role of backend execution engines. We would get a win in just being able to show GATs as a runtime independent language for modeling linear systems. 

The solvers in Krylov are different being targeted at the Optimization community rather than the NLA community. 

[1] https://github.com/JuliaSmoothOptimizers/LinearOperators.jl 
[2] https://github.com/JuliaSmoothOptimizers/Krylov.jl

jpfairbanks:
A reason to do this is methods like LSMR for solving Least Squares Problems via the normal equations. 

[1] https://juliasmoothoptimizers.github.io/Krylov.jl/dev/solvers/#Krylov.lsmr

epatters:
I really like this idea, both for its practical utility and as a demonstration of multiple instances for a single GAT. I bet it would be easy to implement too, based on how easy it was to make an instance for LinearMaps.jl.

jpfairbanks:
This would be a good first issue, to translate the `@instance` for LinearMaps to an `@instance` for LinearOperators.

epatters:
Resolved by #140. Thanks @bosonbaas!

PR #140, opened 2020-03-17
Added LinearOperators instance of GLA

bosonbaas:
This pull request is designed to resolve issue #125


src/linear_algebra/GLA.jl
```diff
@@ -11,6 +11,9 @@ using AutoHashEquals
 using LinearMaps
 import LinearMaps: adjoint
 const LMs = LinearMaps
+using LinearOperators
+import LinearOperators: adjoint
+const LOs = LinearOperators
```
jpfairbanks: I think that we can add the opEye to the import line above and then eliminate this module alias.

test/linear_algebra/GLA.jl
```diff
@@ -48,6 +49,28 @@ x, y = [2, 1], [7, 3, 5]
...
+@test mcopy(dom(f))*x == [x;x]
+@test delete(dom(f))*x == []
+@test plus(dom(f))*[x;x] == 2x
+@test zero(dom(f))*Float64[] == zero(x)
+
+@test (h+k)*x == h*x + k*x
+@test scalar(dom(f),3)*x == 3*x
+@test antipode(dom(f))*x == -1*x
+@test adjoint(g)*y == g'*y
+
```
jpfairbanks: We should add some tests that use the `functor` function to plug these maps into an expression like:
```julia
A, B = Ob(FreeLinearFunctions, :A, :B)
f, g = Hom(:f, A, B), Hom(:g, A, B)
h, k = Hom(:h, B, A), Hom(:k, B, A)
fop, gop = LinearOperator([0 1; 1 0]), LinearOperator([1 2 3; 4 5 6; 7 8 9])
hop, kop = LinearOperator([1 -1; -1 1; 0 1]), LinearOperator([1 1; 2 2; 3 3])

@test functor(compose(f,h), Dict(f=>fop, h=>hop)) * [1 1] == hop*fop*[1 1]
```

Project.toml
```diff
@@ -15,6 +15,7 @@ LightGraphs = "093fc24a-ae57-5d10-9952-331d41423f4d"
 LightXML = "9c8b4983-aa76-5018-a973-4c85ecc9e179"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LinearMaps = "7a12625a-238d-50fd-b39a-03d52299707e"
+LinearOperators = "5c8ed15e-5a4c-59e4-a42b-c7e8811fb125"
```
epatters: To appease the Julia package bot, we need to add a corresponding entry to the `[compat]` section.

src/linear_algebra/GLA.jl
```diff
@@ -156,6 +159,56 @@ plus_lm(x::AbstractVector) = begin
...
+
+  compose(f::LinearOperator, g::LinearOperator) = g*f
+  id(V::LinearOpDom) = LOs.opEye(V.N)
+
+  oplus(V::LinearOpDom, W::LinearOpDom) = LinearOpDom(V.N + W.N)
+  oplus(f::LinearOperator, g::LinearOperator) = oplus_lo(f, g)
+  mzero(::Type{LinearOpDom}) = LinearOpDom(0)
+  braid(V::LinearOpDom, W::LinearOpDom) = braid_lo(V.N, W.N)
+  mcopy(V::LinearOpDom) =
+    LOs.opExtension(1:V.N, V.N*2)+LOs.opExtension((V.N+1):(V.N*2), V.N*2)
```
epatters: I know this is nit-picky, but my brain wants to parse `V.N*2` as `V.N^2` or `V.N**2` (Python style). Maybe write this as `2*V.N`?

epatters:
Thanks Andrew, this looks great. Incidentally, I like the `opExtension` and `opRestriction` functions that LinearOperators provides.

src/linear_algebra/GLA.jl
jpfairbanks: Oh yeah, I never thought about it, but I agree with that nitpick. Constants before variables just like in hand written math notation.

src/linear_algebra/GLA.jl
```diff
@@ -156,6 +159,56 @@ plus_lm(x::AbstractVector) = begin
...
+  plus(V::LinearOpDom) =
+    LOs.opRestriction(1:V.N, V.N*2)+LOs.opRestriction((V.N+1):(V.N*2), V.N*2)
+  zero(V::LinearOpDom) = LOs.opZeros(V.N, 0)
+
+  plus(f::LinearOperator, g::LinearOperator) = f+g
+  scalar(V::LinearOpDom, c::Number) = LOs.opEye(typeof(c),V.N)*c
+  antipode(V::LinearOpDom) = scalar(V,-1)
+end
+
+oplus_lo(f::LinearOperator, g::LinearOperator) = begin
```
jpfairbanks: we don't need to suffix the function names here because julia will use multiple dispatch to pick the right method based on the types of the arguments

src/linear_algebra/GLA.jl
```diff
@@ -156,6 +159,56 @@ plus_lm(x::AbstractVector) = begin
...
+  codom_total = size(f,1) + size(g,1)
+  dom_f   = 1:size(f,2)
+  codom_f = 1:size(f,1)
+  dom_g   = (size(f,2)+1):dom_total
+  codom_g = (size(f,1)+1):codom_total
+  fOp = LOs.opExtension(codom_f, codom_total)*f*LOs.opRestriction(dom_f,dom_total)
+  gOp = LOs.opExtension(codom_g, codom_total)*g*LOs.opRestriction(dom_g,dom_total)
+  fOp + gOp
+end
+braid_lo(v::Int, w::Int) = begin
```
jpfairbanks: Unlike the `oplus_lo` case above, this function does need a suffix or an additional input argument. One idiomatic julia way to handle this is to pass the return type as an argument to the function. Like

```julia
function braid(::Type{LinearOperator}, v::Int, w::Int)
  upper = LOs.opExtension(1:w, v+w) * LOs.opRestriction((v+1):(v+w),v+w)
  lower = LOs.opExtension((w+1):(v+w), v+w) * LOs.opRestriction(1:v,v+w)
  upper + lower
end
```

which you would then call as `braid(LinearOperator, 5,2)`. When you pass the return type as an argument it allows multiple dispatch to find the right method of that function. This is how `zeros(Int , 16)` or `ones(Float64, 32)` know what type to fill the array with. When you do this idiom, the type gets passed as the first argument, that is just a convention they picked in the julia standard library, so people use it everywhere.

src/linear_algebra/GLA.jl
bosonbaas: Good call, I'll make those edits. Definitely makes more sense stylistically.

src/linear_algebra/GLA.jl
bosonbaas: Thanks, that's good to know! I'll get those updated.

src/linear_algebra/GLA.jl
epatters: Right, this definition can be moved into the body of the `@instance` with just the name `oplus`.

src/linear_algebra/GLA.jl
```diff
@@ -156,6 +161,56 @@ plus_lm(x::AbstractVector) = begin
...
+  codom_total = size(f,1) + size(g,1)
+  dom_f   = 1:size(f,2)
+  codom_f = 1:size(f,1)
+  dom_g   = (size(f,2)+1):dom_total
+  codom_g = (size(f,1)+1):codom_total
+  fOp = opExtension(codom_f, codom_total)*f*opRestriction(dom_f,dom_total)
+  gOp = opExtension(codom_g, codom_total)*g*opRestriction(dom_g,dom_total)
+  fOp + gOp
+end
+braid_lo(::Type{LinearOperator}, v::Int, w::Int) = begin
```
epatters: Likewise, this can be moved into the body of `@instance` with signature `braid(V::LinearOpDom, W::LinearOpDom)`.

src/linear_algebra/GLA.jl
epatters: Looks like this is resolved now, except for the lingering alias `LOs`.

src/linear_algebra/GLA.jl
epatters: Just saw your comment above @jpfairbanks. What is the rationale for `braid(::Type{LinearOperatior}, ...)` instead of a dispatch on object type as in `braid(V::LinearOpDom, W::LinearOpDom)`?

bosonbaas: The definition of `braid_lo` is meant to just be a helper function. Since `braid(V::LinearOpDom, W::LinearOpDom)` is already present in the body of `@instance`, I think @jpfairbanks was trying to keep from defining two functions with the same signature (he was referring to the helper function, not the in-instance function).
You're right, though, that a helper function can be avoided here and the function can be moved into the body of `@instance`, so I'll make that edit!

jpfairbanks: Yeah I was just explaining the Julia idiom.  I think you should just eliminate that helper function entirely and inline it into the instance definition.

src/linear_algebra/GLA.jl
bosonbaas: Okay! I was trying to follow the pattern set up by the LinearMaps instance, where there seem to be helper functions (eg delete_lm, plus_lm) defined after the `@instance` block to keep the `@instance` block itself clean. Is it preferred to just have the entire implementation inside the `@instance` block?

src/linear_algebra/GLA.jl
bosonbaas: Just removed the lingering alias

src/linear_algebra/GLA.jl
epatters: IMO, it's good to put the implementation inside the `@instance` block when feasible. The reason I had the helper functions in the other case is that I actually had to use them at multiple spots inside the `@instance`, due to the way that `LinearMaps` handles adjoints.

epatters:
This shaped up very nicely. Thanks @bosonbaas!

src/linear_algebra/GLA.jl
bosonbaas: Ah, that makes sense! Thanks for the explanation!

Issue #144, opened 2020-03-24
Transfer of Instances to Separate Experiment Folder


bosonbaas:
As we begin to add more `@instances` of `@theories`, we were wondering if it would make sense to have a separate folder in root (possibly called "experiments") to contain these implementations. Inside this folder, we could have a separate directory for each `@instance` with it's own Project.toml and Manifest.toml specification. 

This would keep the Catlab.jl package from obtaining more dependencies as more `@instances` are being developed. @jpfairbanks may also have additional thoughts on this.

epatters:
Oh, interesting. Is the idea to be "mono-repo, multi-package," at least for smaller experiments? That could definitely make sense as way to avoid dependency explosion while still keeping development manageable.

bosonbaas:
Yeah, that would be the idea for smaller experiments! @jpfairbanks made the comment that they will be attempting a similar transition in [SemanticModels.jl](https://github.com/jpfairbanks/SemanticModels.jl) to deal with dependency bloat.

jpfairbanks:
Yeah I think we should manage all the AlgebraicJulia packages that way. A folder for examples/experiments that we aren’t committed to supporting with SemVer and have dependencies that we don’t want Catlab to require. Then if they become useful enough that people want to use them, they get integrated into `src` or a new repo in the AlgebraicJulia org for that example.

epatters:
I am on board with this plan. As a start, the stuff currently in [Catlab.Experimental](https://github.com/epatters/Catlab.jl/tree/master/src/experimental) could be moved out.

jpfairbanks:
Yeah I think each of those files 

-  AlgebraicNets.jl 
-  Experimental.jl 
-  Markov.jl 
-  MathFormulas.jl 

Could be their own subproject. The first step is to make them a project in the `Catlab/examples/` folder with their own Project.toml file and package structure, then when an example matures so that people would want to install it separately, that code becomes a baby repo in the AlgebraicJulia organization. That way we strike a balance between tiny half baked repos and dependency management in Catlab.

epatters:
Agreed. Note that AlgebraicNets.jl + MathFormulas.jl should go together, since they depend on each other, and Experimental.jl is just the parent module. So we only need two sub-projects for these.

bosonbaas:
Thanks for the input! I'll try to get this together and submitted in a PR soon.

bosonbaas:
Question for how testing will work with this new structure:
[...]

PR #147, opened 2020-03-27
Dagger Instance

jpfairbanks:
This PR starts work on parallel computing with Catlab by making an instance of GLA for Dagger DAGs. This allows us to execute linear maps in parallel using the Dagger scheduler.


experiments/DaggerInstanceGLA/Manifest.toml
```diff
@@ -0,0 +1,341 @@
+# This file is machine-generated - editing it directly is not advised
```
jpfairbanks: This Manifest shouldn't get committed to the Repo examples/*/Manifest.toml should go in the gitignore

experiments/DaggerInstanceGLA/src/DaggerInstanceGLA.jl
```diff
@@ -0,0 +1,78 @@
...
+# LinearFunctions from having to think the thunk each time the dom or codom
+# is queried
+
+struct MatrixThunk
+  thunk::Thunk
+  dom::Int
+  codom::Int
+end
+
+matrixToThunk(A::LinearMap) = begin
```
jpfairbanks: you can do `identity` instead of `x->x`

jpfairbanks: This function should be called `Thunk(A::LinearMap)` and you add `import Dagger: Thunk` to the top of the file

experiments/DaggerInstanceGLA/src/DaggerInstanceGLA.jl
```diff
@@ -0,0 +1,78 @@
...
+@instance LinearFunctions(DagDom, MatrixThunk) begin
+
+  adjoint(f::MatrixThunk) = MatrixThunk(delayed(adjoint)(f.thunk), f.codom, f.dom)
+  +(f::MatrixThunk, g::MatrixThunk) = MatrixThunk(delayed(+)(f.thunk, g.thunk), f.dom, f.codom)
+
+  dom(f::MatrixThunk) = f.dom
+  codom(f::MatrixThunk) = f.codom
+
+  compose(f::MatrixThunk, g::MatrixThunk) = MatrixThunk(delayed(*)(g.thunk,f.thunk), g.dom, f.codom)
+  id(V::DagDom) = MatrixThunk(delayed(x->x)(LMs.UniformScalingMap(1, V.N)), V.N, V.N)
```
jpfairbanks: `identity` should work here too.

experiments/DaggerInstanceGLA/src/DaggerInstanceGLA.jl
```diff
@@ -0,0 +1,78 @@
...
+
+  oplus(V::DagDom, W::DagDom) = DagDom(V.N + W.N)
+  oplus(f::MatrixThunk, g::MatrixThunk) = MatrixThunk(delayed(LMs.BlockDiagonalMap)(f.thunk, g.thunk), f.dom+g.dom, f.codom+g.codom)
+  mzero(::Type{DagDom}) = DagDom(0)
+  braid(V::DagDom, W::DagDom) =
+  MatrixThunk(delayed(x->x)(LinearMap(braid_lm(V.N), braid_lm(W.N), W.N+V.N, V.N+W.N)),V.N+W.N, W.N+V.N)
+
+  mcopy(V::DagDom)  = MatrixThunk(delayed(x->x)(LinearMap(mcopy_lm, plus_lm, 2*V.N, V.N)), V.N, 2*V.N)
```
jpfairbanks: I think these should be `MatrixThunk(delayed(x->LinearMap(mcopy_lm, plus_lm, 2*V.N, V.N)*x), V.N, 2*V.N)` etc.

experiments/DaggerInstanceGLA/src/test.jl
```diff
@@ -0,0 +1,23 @@
...
+
+
+d = Dict(f=>fop, g=>gop, h => hop, V => DagDom(4), W=>DagDom(3))
+F(ex) = functor((DagDom, MatrixThunk), ex, generators=d)
+#d = Dict(f=>fop, g=>gop, h => hop, V => LinearMapDom(4), W=>LinearMapDom(3))
+#F(ex) = functor((LinearMapDom, LinearMap), ex, generators=d)
+
+M = plus(f,g)⋅h
+
+print(Matrix(collect(F(M).thunk)))
```
jpfairbanks: Need some real tests to make sure we get the numerically correct answer.

jpfairbanks:
@bosonbaas, I think this is a good start. I noted some stuff that can be simplified. And we need some tests before this is ready to merge.

bosonbaas:
Thanks for the comments! It might be useful to wait on pulling this request until we transfer `Catlab.Experiments` over to it's own dedicated folder ( #144 ) so that this can be placed into that structure.

epatters:
Very cool. Dagger is new to me, although I once used dask, which looks very similar.

Thanks for getting this started, Andrew.

experiments/DaggerInstanceGLA/src/DaggerInstanceGLA.jl
bosonbaas: Yeah, I see what you're getting at with that, I had a hard time with this one (as well as braid and delete). I think as-is will cause a typeError, since `delayed(x->LinearMap(mcopy_lm, plus_lm, 2*V.N, V.N)*x)` is not of type `Thunk`.

From the way it's used in the LinearMaps instance, `mcopy` returns a morphism (LinearMap object) which is then applied to other morphisms through composition. If we're keeping with that usage (applying `mcopy` by composition), I think we'll want this to return a MatrixThunk which has an identity Thunk for the LinearMap `mcopy`, which can then be used as an argument in a `compose` call.

bosonbaas:
Currently the only tests I have are comparing the results of expressions evaluated as `LinearMap`s and as `MatrixThunk`s in an attempt to show equivalence between the two instances. What other tests would be good to have in here?
I'll hold off on asking about the structure of the `tests` directory here, as I'll put that in #144 for discussion.

jpfairbanks:
Here are the notes from our conversation today, I think we might want to have both the OpenDAG approach you implemented here and a "CT-native" approach like the following Thunk implementation.

```julia
AbstractThunk

struct ThunkDom
  types::Vector{DataType}
end
  
struct GeneratorThunk <: AbstractThunk
  dom
  codom
  t::Thunk
end

CompositeThunk <: AbstractThunk
  fs::Vector{AbstractThunk}
end
  
ProductThunk <: AbstractThunk
  fs::Vector{AbstractThunk}
end
  
CopyThunk <: AbstractThunk
  x::ThunkDom
end

BraidThunk <: AbstractThunk
  x::ThunkDom
  y::ThunkDom
end

# copy(X::Ob) → (x)↦CopyThunk(x,y)
# braid(X::Ob, Y::Ob) → (x,y)↦BraidThunk(x,y)

# f⊗g⋅h⋅Δ(x)... ↦ CompositeThunk(ProductThunk(f,g), h, CopyThunk(x))
```

epatters:
I haven't looked at this carefully yet, but is the comment I made in #157 relevant here as well?

jpfairbanks:
Yes it is.

epatters:
@jpfairbanks and @bosonbaas, what are your plans for this experiment? Looks like a solid amount of work has gone into it. One possibility is to just merge it into the experiments. For that, I think we would need to make sure all the tests still pass and update the [experiments GitHub CI action](https://github.com/AlgebraicJulia/Catlab.jl/blob/7cc720ff002d8f6e049ce44865c2828a07616f09/.github/workflows/experiments.yml#L9) to run the tests (otherwise, bit rot is inevitable).

jpfairbanks:
I think we would want to update it to use something like `oapply(d::DWD, args::Vector{Thunk})::Thunk` then you would get that Dagger Thunks are an operad algebra for DWDs.

bosonbaas:
I've made an `evaluate` function in Algebraic Relations PR [17](https://github.com/AlgebraicJulia/AlgebraicRelations.jl/pull/17) which implements the conversion from DWD to Thunk and then evaluates that Thunk.
Would this make sense as something to put in the `experiments` folder (after being converted to simply an `oapply` for Thunks)?

jpfairbanks:
I think it makes sense to pull the workflow parts of algebraic relations out into a new package and put this in that package as a DAG construction tool. Algebraic relations could still have the code that generated a schema from a Petri net presentation of an SMC, but the creation and execution of Workflows would go in the new package. Then when you use both Workflows and Relations, you would get checkpointing into SQL for free.

epatters:
Closing this PR, as it will be better suited to new LinAlg package if we ever pick it up again.

PR #149, opened 2020-03-31
Migration from Experimental to Examples

bosonbaas:
This PR is made in response to issue #144. It transitions the Experimental projects into their own Julia Packages, along with their testing structures.


bosonbaas:
What should be done with the Experimental.jl file? Should there still be some single file which points to all example packages, or can Experimental.jl be removed?

epatters:
Thanks for starting this re-org. The file `Experimental.jl` can simply be removed.

Also, what do you think about renaming the directory `examples` to `experiments`? These packages are mainly for experimentation, not for documentation. I don't want to confuse users on this point.

bosonbaas:
Yeah, I definitely agree with that name change. Also, any thoughts on documentation integration with these? Currently the PR is failing the docs generation because the documentation refers to Experimental.

epatters:
For the experiments, I am more concerned about the tests than the docs because the docs are very sparse. Currently, I think there is only one document (`algebraic_nets.jl`, a notebook in Literate form). For now, it can just be dumped into a `docs` folder in the subproject and left there. That should get the main docs building again.

Two more points:

1. I don't think we need the nesting provided by the `core` folder in the subprojects. We can just keep everything at the toplevel in `src`, which is what small Julia projects usually do.

2. Can we add a new GitHub action to run the experiment tests? IMO, this is important. Otherwise, we will forget to update this code when we make breaking changes to Catlab.

bosonbaas:
I noticed that the Experimental folder was actually already removed from the `runtests.jl` function for Catlab tests. When I added Experimental back in, it was running into some errors with `AlgebraicNets.jl`, so that will need to be dealt with when the GitHub Actions is extended for testing `experiments`.

epatters:
Thanks, I think this is almost ready. I confirmed that the tests pass, apart from two errors with the algebraic nets. I can resolve this issue later. We can add the GitHub actions integration later, if you want.

I noticed that the `CompAlgebra/docs` folder has two copies of a notebook and a Markdown. I am no longer committing Jupyter notebooks to this repo since they diff so badly. Can you please remove everything in this folder except `algebraic_nets.jl`?

epatters:
Excellent, thanks!

PR #157, opened 2020-04-08
StochMaps

jpfairbanks:
@epatters, here is a first crack at `StochMaps` to implement an Instance of Markov Kernels as an SMC + Diagonals. The next step is to actually associate the `StochDom`, `StochMaps` types with an `@instance` which should be straight forward. 

The CT approach gives you an obvious roadmap for an implementation. Once you have done the math to formalize a mathematical modeling framework into a GAT, you have a roadmap. You know what algebraic operations you need to support on your types and can just make types for each of those operations. I guess when you design software this way you are going to get AST-like data structures that mimic the algebraic formulas of the GAT you are working with. Then the algorithms for doing stuff with those structures are going to look like "walk the tree in preorder traversal order". 

When we go to extend `StochMap` to support CCC structure, we just need to add the `StochEv` and `StochCurry` types and make sure that `compose, otimes, mcopy, braid` are defined for them. And then we will have a nice little PPL


epatters:
Cool, thanks!

A thought about the implementation: it seems like there is a fair amount of boiler plate code involved in creating structs for composites, products, etc. Since these structs basically mirror the structure of an expression tree, an alternative strategy would be to just use regular old expressions. You are allowed to put anything you want inside the generators. So, you could put actual Julia types or functions or distributions inside the generators, and then have `crand` dispatch on the GAT term, e.g., `HomExpr{:compose}` and `Hom{:otimes}`.

jpfairbanks:
Oh that’s awesome, I was thinking of the GATExprs as purely syntactic and never would have thought to put “real” Julia values in them. This makes it really easy to define a datastructure that is mathematically a Functor from C to D you can just put the generators in D into GAT expressions and use dispatch

jpfairbanks:
The more I think about it, the more I think this is the right way to think about Algebraic Theories as DSLs. You get a language for expressing the systems, then a "natural" datastructure for representing them (the GATExprs modulo equivalence dictated by the axioms of the theory). Properties of the system that can be defined functorially can be computed with the "functor" algorithm, ie. "tell me what to do with the generators and term constructors and I'll just walk the tree and inductively compute the property."

Since models of a theory are functors in the categorical logic sense, "any property of a system that is defined functorially" is equivalent to "any model of the theory evaluated at a particular term in the syntax category". 

The fact that SMCs have string diagrams is the sense in which "modeling languages are graphs + mathematical interpretations of those graphs"

jpfairbanks:
Ok, that was waaaay easier!

```julia
using Catlab
using Catlab.Doctrines
import Base: length
length(::FreeCartesianCategory.Ob{:generator}) = 1
length(f::FreeCartesianCategory.Ob{:otimes}) = sum(length.(f.args))

# How do you give semantics to a stochastic map? You call it.
function crand(f::FreeCartesianCategory.Hom{:generator}, args...)
    f.args[1](args...)
end

# Compositional structure
crand(f::FreeCartesianCategory.Hom{:id}, args...) = (args)
function crand(f::FreeCartesianCategory.Hom{:compose}, args...)
    if length(f.args) > 2
        return crand(f.args[end], crand(compose(f.args[1:end-1]...), args...)...)
    end
    return crand(f.args[end], crand(f.args[1], args...)...)
end

# Monoidal Structure
function crand(f::FreeCartesianCategory.Hom{:otimes}, args...)
    map(1:length(f.args)) do i
        crand(f.args[i], args[i]...)
    end |> xs->filter(xs) do x # handle the () you get from deletes
        x != ()
    end
end
function crand(f::FreeCartesianCategory.Hom{:braid}, args...)
    y = args[1:length(f.args[1])]
    x = args[(length(f.args[1])+1):end]
    return (x...,y...)
end

# Diagonal Comonoid
crand(f::FreeCartesianCategory.Hom{:mcopy}, args...) = (args..., args...)
crand(f::FreeCartesianCategory.Hom{:delete}, args...) = ()

X = Ob(FreeCartesianCategory, :Float64)
u = Hom(x->x*rand(), X, X)
crand(((u⊗u)⋅braid(X,X)), 1,2)
crand(mcopy(X),2)
crand((mcopy(X)⋅(u⊗u)),2)
crand(delete(X), 1)
crand(u⋅delete(X), 1)
crand(u⋅mcopy(X)⋅(id(X)⊗delete(X)), 1)
```

epatters:
Nice, that is much shorter!

I agree that GAT expressions are the natural or "default" data structure for categorical DSLs. You get them for free, in a sense, just from the description of the GAT. Other data structures, such as wiring diagrams, may have better computer-algebraic properties, but you don't get them for free. Someone has to figure out what is the right graphical syntax for each doctrine and there is no algorithm for doing that.

jpfairbanks:
Yeah, this connects to your idea that Wiring Diagrams are "partial normal form" for SMCs. I think the fact a WD in Stoch is basically a bayesian network is connected to that. I don't know how to formally state it, but I think there is a theorem that "every time a computer scientist uses a graph, they are doing monoidal category theory without admitting it."

jpfairbanks:
So it looks like to get this merged, we should just

- [ ] Delete the StochMaps implementation
- [ ] Repurpose the StochMaps tests for the Stoch Exper implementation

Anything else?

epatters:
That sounds right. Thanks for working on it.

It looks like you accidentally committed a PDF file :) Can you please edit the commit and force push so this doesn't end up in version control?

jpfairbanks:
haha, yeah will rm. I added that for docs to myself.

jpfairbanks:
Due to my lack of confidence in my git skills, I made a new branch and rebased #158

epatters:
Thanks.

PR #158, opened 2020-04-20
ENH: Implement Stochastic Maps as GATExprs

jpfairbanks:
Rebase of PR closes #157

### DIFF
--- a/experiments/Markov/src/Markov.jl
+++ b/experiments/Markov/src/Markov.jl
@@ -1,5 +1,7 @@
 module Markov
 
 include("MarkovCategories.jl")
+include("StochMaps.jl")
+include("StochExpr.jl")
 
 end
--- a/experiments/Markov/src/Markov.jl
+++ b/experiments/Markov/src/Markov.jl
@@ -1,7 +1,6 @@
 module Markov
 
 include("MarkovCategories.jl")
-include("StochMaps.jl")
 include("StochExpr.jl")
 
 end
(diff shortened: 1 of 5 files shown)


experiments/Markov/src/StochExpr.jl
```diff
@@ -0,0 +1,46 @@
+using Catlab
+using Catlab.Doctrines
+import Base: length
+
+export crand
+
+length(::FreeCartesianCategory.Ob{:generator}) = 1
```
epatters, line 7: This convenient function is also ready implemented in `Catlab.Doctrines` as `ndims`. Maybe it should be `length`? Depends on how you think about the object.

experiments/Markov/src/Markov.jl
```diff
@@ -1,5 +1,7 @@
 module Markov
 
 include("MarkovCategories.jl")
+include("StochMaps.jl")
```
epatters, line 4: Should this line be here? I only see `StochExpr.jl`.

epatters:
Maybe a few small comments. Otherwise, LGTM.

experiments/Markov/src/Markov.jl
jpfairbanks: Leftover from the last deletion.

jpfairbanks:
🚢 ?

epatters:
🚢-ed!

PR #162, opened 2020-05-10
add example recipes to the literate folder

jpfairbanks:
@epatters just a first cut at implementing the recipes category. I think that we need to reckon with the fact that cooking has generic operations like `mix(A,B)::Hom(otimes(A,B), mixture(A,B))` for all pairs of objects `A,B` or `cook(A)::Hom(A, Cooked(A))`.

### DIFF
--- a/docs/literate/recipes/recipes.jl
+++ b/docs/literate/recipes/recipes.jl
@@ -1,0 +1,104 @@
+using Catlab
+using Catlab.Doctrines
+using Catlab.Programs
+using Catlab.WiringDiagrams
+using Catlab.Graphics
+using Catlab.Graphics.Graphviz
+
+import Catlab.Doctrines: Ob, Hom, dom, codom, compose,
+  ⋅, ∘, id, otimes, ⊗, munit, braid,
+  mcopy, Δ, delete, ◊, mmerge, ∇, create, □
+
+draw(d::WiringDiagram) = to_graphviz(d, orientation=LeftToRight, labels=true)
+
+@present Bakery(FreeBiproductCategory) begin
+  egg::Ob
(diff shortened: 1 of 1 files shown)


epatters:
Thanks for starting this example!

Do we need a shorthand for declaring multiple objects in one line in a presentation? Those lists look pretty long.

jpfairbanks:
It was a pain to type for sure! Maybe `X,Y,Z::Ob...` would be nice or `(X,Y,Z)::Ob`? Maybe just using a constructor instead of a macro?

`P = Presentation(FreeBiprodyctCategory, obs, homs)` and then use broadcasting to make the list of Obs and Homs? The probable here is that I have to refer to the objects that I create to create the homs which means you have to make a Julia variable for each object or refer to them as obs[i] when creating the homs. I guess this is why the present macro exists. 

The next step here is to implement a recipe calculator that takes an amount of ingredients and tells you the amount of product it will make. It should support a class of functions for each hom that are closed under scaling up the amounts by positive integer multiples. You can’t have a recipe that calls for 1.5 eggs. 

The calculator could tell you the amount of product you will get based on the amount of ingredients. But it can also tell you the time it would take to make it if you associated a time to each hom and then did an optimal scheduling approach. That would be the work-depth model of parallel computing we had talked about.

epatters:
Since today we're going through old PRs, I'll note that the abbreviated syntax was finally implemented in #292. @jpfairbanks, do you plan to finish this? Or does someone else want to take it over?

epatters:
Going to close this PR since it's now very old and we have plans to take cooking more seriously! But feel free to reopen if you want to make this into a sketch.

PR #170, opened 2020-06-18
New logo for Catlab

mehalter:
Would we want to migrate the logo from SemanticModels to Catlab? I updated and added the assets as well as updated the README and the docs.


mehalter:
Oh I didn't mean to commit those, they aren't being used anywhere

epatters:
Perfect, thanks!

Issue #165, opened 2020-06-15
Roadmap: GAT of GATs


jpfairbanks:
@epatters @mehalter @bosonbaas, here are some tasks I think we should address for the next push on Catlab (aka Gatlab).

- [ ] CT constructions
    - [ ] Product/Sum
    - [ ] Span/Cospan
    - [ ] Functor/Instance
    - [ ] Category of Elements
- [ ] GAT Functors, injective, surjective to talk about relationships between algebraic theories
    - [ ] Causal Theories
    - [ ] Statistical Theories
    - [ ] Physical Theories
- [ ] Compute pushout/pullback of GAT functors
    - [ ] Cospan category for open GATs
- [ ] Pushout/pullback within a Cat, can we infer the rules for composition of FinSet cospans from the definition of FinSet morphisms?
- [ ] Relation orient computing: functors from an arbitrary GAT into Rel
- [ ] Generic Operads
  - [ ] Syntax
  - [ ] Drawing
  - [ ] Substitution based on term matching
- [ ] Axiom Support
  - [ ] Generating term rewriting rules from the specification of the axioms of a GAT

mehalter:
- [ ] Task 0: Refactor SemanticModels -> (Catlab.jl, AlgebraicX.jl)
- [ ] Task 1: Refactor Catlab.jl implementation of GAT to provide more efficient and easier to work with data structures

PR #175, opened 2020-06-22
Begin refactoring of GATs

epatters:
Some first steps towards refactoring the GAT system (see #165):

- The `Typeclass` wrapper struct, a holdover from the days when GATs mixed with Julia code, has been removed.
- The `@theory`/`@signature` macros no longer generate a whole module with stub Julia types. Instead, they generate a single abstract type representing the theory, e.g. an abstract type `Category`. The theory of categories can then be retrieved by calling `theory(Category)`. So we are now abusing the module system somewhat less.
- The stub methods generated by `@theory`/`@signature` have been replaced with a mere generic function declaration, reducing the size of the method tables for `compose`, `id`, etc by about a half. (I don't know why I didn't do this before. Probably because when I first wrote that code I barely knew Julia.)

Most of these changes will be invisible to most users. The main breaking change is that instead of writing `Port{CartesianCategory.Hom}` and `WiringDiagram{CartesianCategory.Hom}`, we will now write `Port{CartesianCategory}` and `WiringDiagram{CartesianCategory}`, which is cleaner anyway.

### DIFF
--- a/experiments/Markov/src/MarkovCategories.jl
+++ b/experiments/Markov/src/MarkovCategories.jl
@@ -33,13 +33,13 @@ end
 # Wiring diagrams
 #################
 
-mcopy(A::Ports{MarkovCategory.Hom}, n::Int) = implicit_mcopy(A, n)
+mcopy(A::Ports{MarkovCategory}, n::Int) = implicit_mcopy(A, n)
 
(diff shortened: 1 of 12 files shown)


epatters:
Thanks! It's a start.

PR #176, opened 2020-06-23
Proper typing of presentations

epatters:
Clean up to the presentation system, enabled by #175:

- Type parameterize presentations by theory, so that a presentation of a category will have type `Presentation{Category}`. This fixes a longstanding known issue.
- Index generators by GAT type, so that, for example, it is fast to retrieve all the object generators, or all the morphism generators, in a presentation of a category.


epatters:
Thanks! I guess I forgot to run the full test suite before I made the PR.

PR #179, opened 2020-06-26
[WIP] Continued GAT Refactoring

mehalter:
So far I have:

- Removed Expr types from GAT defintions 
- Began separating code geration and theory creation
- Made code generation only dependent on a theory object

### DIFF
--- a/src/core/GAT.jl
+++ b/src/core/GAT.jl
@@ -11,6 +11,8 @@ using Match
 
 using ..Meta
 
+import Base: ==, hash
+
 # Data types
 ############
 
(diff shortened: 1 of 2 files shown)


mehalter:
To remove dependency of any of the raw macro head/body objects for the code generation, I added the theory name to the theory struct. I'm not sure if this is the best way, I was also thinking it might make sense to not have it store a name and the code generation struct would just also take in a name possibly. Having anonymous theories I think will be valuable down the line when we want to start doing pushouts and things with GAT of GATs

epatters:
I agree, eventually we'll want to be able to manipulate theories anonymously, but I think this is fine for now if it simplifies the refactor.

jpfairbanks:
@epatters, there has been a ton of work since this was originally written, do we still want this refactoring code?

epatters:
Unfortunately this got stale because Micah had higher-priority work to do. I'll close for now, but I may refer to this when I finally get to the GAT refactoring myself.

PR #182, opened 2020-06-29
ENH: n-ary (co)products

olynch:
I don't want to merge this yet, but I would like some feedback about my current approach to n-ary (co)products. My plan is to build arbitrary limits on top of n-ary (co)products and n-ary (co)equalizers.

### DIFF
--- a/src/categorical_algebra/FinSets.jl
+++ b/src/categorical_algebra/FinSets.jl
@@ -2,7 +2,7 @@
 """
 module FinSets
 export FinOrd, FinOrdFunction, force, terminal, product, equalizer, pullback,
-  initial, coproduct, coequalizer, pushout
+  limit, initial, coproduct, coequalizer, pushout, colimit
 
 using AutoHashEquals
 using DataStructures: IntDisjointSets, union!, find_root
@@ -104,10 +104,27 @@ function product(A::FinOrd, B::FinOrd)
   Span(π1, π2)
 end
 
+function product(Xs::Vector{<:FinOrd})
+  ns = Int[X.n for X in Xs]
+  indices = CartesianIndices(tuple(ns...))
+  apex = prod(ns)
+  πs = [FinOrdFunction(i -> indices[i][j],apex,ns[j]) for j in 1:length(ns)]
+  Cone(FinOrd(apex),πs)
+end
+
 function equalizer(f::FinOrdFunction, g::FinOrdFunction)
   @assert dom(f) == dom(g) && codom(f) == codom(g)
   m = dom(f).n
   FinOrdFunction(filter(i -> f(i) == g(i), 1:m), m)
+end
+
+function equalizer(fs::Vector{<:FinOrdFunction})
+  @assert length(fs) >= 1
+  f1 = fs[1]
+  frest = fs[2:end]
+  @assert all(dom(f) == dom(f1) && codom(f) == codom(f1) for f in frest)
+  m = dom(f1).n
+  FinOrdFunction(filter(i -> all(f1(i) == f(i) for f in frest), 1:m),m)
 end
 
 """ Pullback of cospan of functions between finite ordinals.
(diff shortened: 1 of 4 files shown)


epatters:
Hi Owen, I think this is a great start. The approach makes sense to me.

olynch:
Generic (co)limits are implemented, I'd love to merge after feedback.

olynch:
Should I rebase on top of the C-Sets implementation of graphs?

epatters:
Great, thanks! I'll take a look soon. @jpfairbanks and @mehalter have also had thoughts about the (co)limits interface, so hopefully they will chime in too.

Don't worry about rebasing off the C-sets branch. We can sort that out later. I think you will need to rebase off master to resolve some minor conflicts with stuff Micah added recently.

jpfairbanks:
@olynch, @epatters, these look good to me, getting master merged in and then test passing and we should be ready to merge.

src/categorical_algebra/FinSets.jl
```diff
@@ -104,12 +104,29 @@ function product(A::FinOrd, B::FinOrd)
   Span(π1, π2)
 end
 
+function product(Xs::Vector{FinOrd{T}}) where {T}
```
epatters, line 107: Is it worth reducing all the binary products, coproducts, etc to the general n-ary case, to avoid redundant code? The only reason not to do it would be performance, but I suspect that the difference will be negligible.

epatters: Also, do `product` and `coproduct` work in the special case `length(Xs) == 0`, where they should give terminal and initial objects, respectively?

src/categorical_algebra/ShapeDiagrams.jl
```diff
@@ -33,6 +35,25 @@ apex(span::Span) = dom(span.left) # == dom(span.right)
 left(span::Span) = span.left
 right(span::Span) = span.right
 
+""" Cone of morphisms in a category.
+"""
+struct Cone{S,T}
```
epatters, line 40: With apologies for being picky, for the sake of consistency, let's stick to indentation of width 2 spaces throughout the codebase.

src/categorical_algebra/FinSets.jl
olynch: I think that the question about the binary coproducts depend on whether you want to use (Co)Span as a datastructure. It seems to me like there are cases where we want to use (Co)Span, and not (Co)Cone, so it might make sense to keep around the special case.

Also, something that I thought of is that if we care about efficiency, then, for instance, in a limit, any object in the diagram which has an arrow pointing to it does not need to be in the big product that is then filtered, because its value is always determined by some other thing. If we implemented this optimization, then the performance of the general case would be more similar to the performance of the specialized case in pushouts/pullbacks. Maybe that would be good for another PR though, as that would complicate the implementation.

src/categorical_algebra/ShapeDiagrams.jl
olynch: 2 spaces are objectively correct :P

src/categorical_algebra/FinSets.jl
epatters: OK, then let's keep as is for now. Even if you follow up with a more complicated but more efficient implementation for the general case, it might be good to keep the simple special case just for clarity.

jpfairbanks: I'm worried about the performance implications of using n-ary instead of binary, because it not only involves the for-loop overhead, but also the memory allocation. If these were tuples it wouldn't matter because julia can use stack allocation for tuples, but with Vector{T} it heap allocates. If we relaxed the type constraint here, you could pass a (f,g)::Tuple{FinOrd, FinOrd} and get the best of both worlds. In that case we should just assume that `Xs` is iterable and use that interface for both binary and n-ary (co)products

epatters:
Great, I think this is ready! Your commit message says you're planning to squash commits. I'll merge once that's done.

epatters:
Thanks again @olynch!

PR #183, opened 2020-06-30
Implementation of Decorated Cospans

mehalter:
As I am working on [AlgebraicPetri.jl](https://github.com/AlgebraicJulia/AlgebraicPetri.jl), I am putting together an implementation of decorated cospans in Catlab to help work with open systems.

### DIFF
--- a/src/categorical_algebra/ShapeDiagrams.jl
+++ b/src/categorical_algebra/ShapeDiagrams.jl
@@ -41,4 +41,19 @@ base(cospan::Cospan) = codom(cospan.left) # == codom(cospan.right)
 left(cospan::Cospan) = cospan.left
 right(cospan::Cospan) = cospan.right
 
+""" Decorate Cospan of morphisms for representing open networks.
+"""
+struct DecoratedCospan{Left,Right,Decoration}
+  cospan::Cospan{Left,Right}
+  decoration::Decoration
+end
+
+decoration(m::DecoratedCospan) = m.decoration
+undecorate(m::DecoratedCospan) = m.cospan
+dom(m::DecoratedCospan) = dom(m.cospan)
+codom(m::DecoratedCospan) = codom(m.cospan)
+base(m::DecoratedCospan) = base(m.cospan)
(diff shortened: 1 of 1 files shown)


mehalter:
I added some abstract types for Functor, Laxitor, and LaxMonoidalFunctor for more general use in the future. Let me know if you want to move this to a different place or not include it in there.

mehalter:
I finished the implementation of AlgebraicPetri and this is the stuff I added to get Decorated Cospans working with Catlab. Let me know if any of this should change before merging

src/categorical_algebra/ShapeDiagrams.jl
```diff
@@ -1,10 +1,20 @@
 """ Diagrams of a given shape.
 """
 module ShapeDiagrams
-export Span, Cospan, left, right, apex, base
+export AbstractFunctor, AbstractLaxitor, LaxMonoidalFunctor,
```
epatters, line 4: `AbstractLaxitor` should be spelled `AbstractLaxator`.

src/categorical_algebra/ShapeDiagrams.jl
```diff
@@ -1,10 +1,20 @@
...
+       Span, Cospan, DecoratedCospan, left, right, apex, base,
+       decorator, decoration, undecorate
 
 using ...Theories: dom, codom
 
+abstract type AbstractFunctor end
```
epatters, line 10: Strictly speaking, stuff about lax monoidal functors doesn't belong in this module, but I think that's OK for now. Once we have a fully generic system for decorated cospans, we can create a separate module for decorated cospans and put everything there.

mehalter: Yeah I agree it doesn't really fit in this module, so we can definitely move it once it becomes sufficiently sophisticated.

src/categorical_algebra/ShapeDiagrams.jl
mehalter: Fixed in latest commit.

PR #187, opened 2020-07-05
C-sets as a generic data structure

epatters:
This PR introduces a generic data structure for C-sets, aka presheaves. For any category C given by finite presentation, `@generated` code is used to achieve performance that should be comparable to a specialized data structure.

In general, C-sets can be small or large along two axes: the theory/schema C and the instance data. This data structure is intended for C-sets that are small along both axes: the schema C should be tiny enough to be reasonably encoded as type parameters, and the data should fit comfortably in memory. (At the other end of the scale are knowledge bases and relational databases, which tend to have large schemas and large data, respectively.) The implementation here effectively includes both graphs and data frames as special cases.

Features:
- Toggleable support for indexing to enable fast lookup of preimages (e.g. edges incident to a vertex)
- Support for "data attribute" morphisms, to which arbitrary data may be attached
- Graphs (aka multidigraphs) and symmetric graphs as C-sets
- Property graphs and symmetric property graphs as C-sets
- As further examples of C-sets (in the unit tests), discrete dynamical systems and dendrograms

Limitations/missing features:
- Indexing of data attributes, unique or otherwise
- Deleting parts from C-sets. This is a bit subtle. The first step is to decide on the expected behavior.
- Deriving `AbstractCSet` instances from functors between indexing categories

Eventually this will be used to eliminate Catlab's dependency on MetaGraphs.jl. The dependency on LightGraphs.jl will be retained for its many useful graph algorithms, such as topological sort.


jpfairbanks:
So cool! LGTM

epatters:
Great, glad to hear it!

PR #197, opened 2020-07-14
Rename concrete types for FinOrd and FinOrdRel

### LINKED ISSUE #184: Names of FinOrdFunctionLazy and FinOrdFunctionMap
I'm having a hard time remembering these names and I think that the problem is that Lazy and Map aren't semantically related. I think we should use `FinOrdFunc` and `FinOrdVec` which are both subtypes of `FinOrdFunction` or we should use `FinOrdLazy` and `FinOrdEager`. Func vs Vec tells you about the storage mechanism and Lazy vs Eager tells you about the evaluation strategy. I'm ok encoding either one, but I think we should be consistent. What do you think?

epatters:
Renames the concrete subtypes of `FinOrdFunction` to `FinOrdFunc` and `FinOrdVector` and the subtypes of `FinOrdRelation` to `FinOrdPredicate` and `FinOrdMatrix`. If you've been using the abstract type constructors, like in the unit tests, nothing will break.

Resolves #184.

### DIFF
--- a/src/categorical_algebra/FinRelations.jl
+++ b/src/categorical_algebra/FinRelations.jl
@@ -1,7 +1,8 @@
 """ Computing in the category of finite sets and relations, and its skeleton.
 """
 module FinRelations
-export BoolRig, FinOrdRelOb, FinOrdRelation, force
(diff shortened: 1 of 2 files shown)


epatters:
I had the same train of thoughts about that :)

PR #198, opened 2020-07-14
Experimental Petset Implementation

olynch:
I have copied some of the features of Petri.jl into a format that uses CSets. I am now thinking that we could move this into AlgebraicPetri. I would like some feedback first, however, and I would like to be more clear on the goals of this whole endeavor.

### DIFF
--- a/experiments/PetSet/PetSet.jl
+++ b/experiments/PetSet/PetSet.jl
@@ -12,47 +12,24 @@ using Catlab.CategoricalAlgebra.CSets
 
 @present SITOS(FreeCategory) begin
   species::Ob
-  input::Ob
+  inputs::Ob
   transitions::Ob
-  output::Ob
+  outputs::Ob
 
-  is::Hom(input,species)
-  it::Hom(input,transitions)
-  os::Hom(output,species)
-  ot::Hom(output,transitions)
+  is::Hom(inputs,species)
+  it::Hom(inputs,transitions)
+  os::Hom(outputs,species)
+  ot::Hom(outputs,transitions)
 end
 
 PetSet = CSetType(SITOS, index=[:is,:it,:os,:ot])
 
-# macro petset(head, body)
-#     name = @match head begin
-#         name::Symbol => name
-#         _ => throw(ParseError("Ill-formed petset header $head"))
-#     end
-#     let_expr = Expr(:let, Expr(:block), translate_petset(body))
-#     esc(Expr(:(=), name, let_expr))
-# end
+nspecies(ps::PetSet) = nparts(ps,:species)
+ntransitions(ps::PetSet) = nparts(ps,:transitions)
+ninputs(ps::PetSet) = nparts(ps,:inputs)
+noutputs(ps::PetSet) = nparts(ps,:outputs)
 
-# function translate_petset(body::Expr)::Expr
-#     @assert body.head == :block
-#     code = Expr(:block)
-#     append_expr!(code, :(_petset = PetSet()))
-#     for expr in strip_lines(body).args
-#         append_expr!(code, translate_expr(expr))
-#     end
-#     append_expr!(code, :(_petset))
-#     code
-# end
-
-# function translate_expr(expr::Expr)::Expr
-#     @match expr begin
-#         Expr(:(::), [name::Symbol, type::Symbol]) =>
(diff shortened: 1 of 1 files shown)


jpfairbanks:
Yeah this looks great, I agree that it should be merged into AlgPetri instead of Catlab

jpfairbanks:
We could also extend this definition of a petri net with the initial value problem as a reaction network. The C-Set

```julia
@present ReactionNet(FreeCategory) begin
  species::Ob
  inputs::Ob
  transitions::Ob
  outputs::Ob
  number::Ob

  is::Hom(inputs,species)
  it::Hom(inputs,transitions)
  os::Hom(outputs,species)
  ot::Hom(outputs,transitions)
  rate::Hom(transitions, number)
  u0::Hom(species, number)
end
```

Then models of this theory (C-Sets) are initial value problems for this system. And the open C-Sets construction of glueing along species would allow you to keep track of the rates and initial conditions when you do the glueing.

olynch:
Could we do that with theory extension?

epatters:
You could even do

```julia
@present TheoryReactionNet <: TheoryPetriNet begin
  number::Ob

  rate::Hom(transitions, number)
  u0::Hom(species, number)
end
```

and then write generic code that works for both Petri nets and reaction nets on their common interface.

epatters:
Looks like we had the same thought @olynch :)

olynch:
OK, sweet, I'll do a fresh implementation in the AlgebraicPetri repo, along with the initial value problem.

olynch:
I'm not totally sure how `data` works in a CSet. Do you provide the type of the data when you define the *type*, or do you provide the type of the data when you want to create a *value*?

jpfairbanks:
For documentation I think we need both a blog post style for users and a technical write-up intended for CT people. I also don't know how the data field works, trying to figure it out led me to file #199!

epatters:
> I'm not totally sure how `data` works in a CSet. Do you provide the type of the data when you define the _type_, or do you provide the type of the data when you want to create a _value_?

Currently, you assign the data type when create the `CSet` instance, not when you create the CSet type via `CSetType`. See this [example](https://github.com/AlgebraicJulia/Catlab.jl/blob/d1c561a757e87a0425c5d88fec13a3e203b9dcb1/test/categorical_algebra/CSets.jl#L64) from the unit tests.

Arguably it should be the other way around, although this would require packing yet more information into the `CSet` type parameters. Any thoughts on this are welcome.

> On a related note, I don't know what your plans for doing long-form documentation of CSets, but I would be happy to help, with either a blog-post style introduction, or more technical documentation.

You would be my hero if you wrote some decent docs :) Like James said, ultimately we need both forms of docs but if I had to pick one, I'd start with an introductory, blog-style piece. A lack of such expository content is currently a big barrier for potential Catlab users.

experiments/PetSet/PetSet.jl
```diff
@@ -0,0 +1,104 @@
...
+
+PetSet = CSetType(SITOS, index=[:is,:it,:os,:ot])
+
+nspecies(ps::PetSet) = nparts(ps,:species)
+ntransitions(ps::PetSet) = nparts(ps,:transitions)
+ninputs(ps::PetSet) = nparts(ps,:inputs)
+noutputs(ps::PetSet) = nparts(ps,:outputs)
+
+# Need a better name for this
+function from_std_petri(S,T)
```
mehalter, line 33: We could also probably have another method call along the lines of `from_std_petri(m::Petri.Model) = from_std_petri(m.S, m.Δ)`

experiments/PetSet/PetSet.jl
```diff
@@ -0,0 +1,104 @@
...
+            if s in keys(outs)
+                add_parts!(ps,:outputs,outs[s],(os=j,ot=i))
+            end
+        end
+    end
+    ps
+end
+
+function transition_matrix(ps)
+    T = zeros(Int64,(ntransitions(ps),nspecies(ps),2))
```
mehalter, line 57: Possibly change this to using `SparseArrays` and `spzeros`

epatters:
Closing this in lieu of https://github.com/AlgebraicJulia/AlgebraicPetri.jl/pull/12. This will be great to have.

PR #203, opened 2020-07-16
LinearMaps and LinearOperators are optional dependencies

epatters:
Uses [Requires.jl](https://github.com/JuliaPackaging/Requires.jl), which was already being used for optional graphics features.


epatters:
Not too long ago somebody complained about Catlab having too many dependencies. I want to take away that excuse.

PR #207, opened 2020-07-18
Tensor notation for undirected wiring diagrams

epatters:
It turns out that a tensor network---a general contraction of a set of tensors---is *exactly* the data of an undirected wiring diagram, at least up to renaming of variables. This PR adds:

- Macro `@tensor_network` to parse an undirected wiring diagram from a tensor contraction expression
- Function `compile_tensor` to generate and compile a tensor expression from an undirected wiring diagram, intended for use with packages like [TensorOperations.jl](https://github.com/Jutho/TensorOperations.jl)

The two operations are essentially inverse to each other. I think it makes sense to include this functionality in Catlab itself because the connection is fundamental and no special dependencies are needed.

### DIFF
--- a/docs/src/apis/graphics.md
+++ b/docs/src/apis/graphics.md
@@ -1,4 +1,4 @@
-# Graphics
+# [Graphics](@id graphics)
 
 ```@autodocs
 Modules = [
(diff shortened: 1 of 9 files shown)


epatters:
Regrettably, Git seems to have lost track of the fact that the file `RelationalPrograms.jl` is renamed from `ParseRelations.jl`.

PR #212, opened 2020-07-23
Drawing undirected wiring diagrams with Graphviz

epatters:
The outer ports are represented as dangling wires, which are not fixed to an outer boundary. The drawings are therefore rudimentary, but this feature should serve as a quick-and-dirty way to visualize undirected wiring diagrams until we develop better methods.


mehalter:
Looks great! LGTM

PR #219, opened 2020-07-26
Experiments: StructArray CSet, and Schema Instances

olynch:
The first one is sort of horrifically complicated, but I think it's rather interesting. This reworks CSets using StructArrays, and also adds the ability to give a type to a morphism.

The second one is complicated in its own way, but has three stand-out features that recommend it. The first is that I use (or, am inspired by...) the theory of "Algebraic Databases", as developed in the paper by Schultz, Spivak, Vasilakopoulou and Wisnesky. The upshot of this is that a `SInstance` is based on a presentation of a `Schema`, which is defined as 

``` julia
@theory Category(Ob,Hom) => Schema(Ob,Hom,Concrete,Attr) begin
  Concrete::TYPE
  Attr(dom::Ob,codom::Concrete)::TYPE
end
```

For instance, the theory of graphs with node decorations is given by:

``` julia
@present TheoryDecGraph(FreeSchema) begin
  V::Ob
  E::Ob
  X::Concrete

  src::Hom(E,V)
  tgt::Hom(E,V)
  dec::Attr(V,X)
end
```

The second innovation I make is that all of the data of a schema presentation is condensed in a type *which can be a type parameter*, namely `SchemaDesc`, which cuts down on a lot of writing. For instance,

``` julia
@generated function _subpart(cset::SInstance{Ts,S}, ::Val{name}) where {S,Ts,name}
  if name ∈ S.hom
    :(csets.subparts.$name)
  elseif name ∈ S.attr
    :(csets.attrs.$name)
  else
    throw(KeyError(name))
  end
end
```

The third innovation is that instances can be declared with type parameters, i.e. for decorated graphs:

``` julia
const DecGraph{T} = SIntance{Tuple{T},SInstanceTypeParams(TheoryDecGraph)...}
```

This allows us to write, say `DecGraph{Int}` for `Int`-decorated graphs, or just `DecGraph` for graphs decorated by something unknown.

I have flip-flopped on whether we want to formalize the special treatment of data morphisms, and I think I've finally landed on the decision that it's better to formalize it, because in the interface to, say, limits and colimits, attributes and morphisms *are* going to be treated very differently. Also, I don't think that we want to allow morphisms between concrete types.

I am also now realizing that we could also do

``` julia
@theory Category(Ob,Hom) => Schema(Ob,Hom,Attr) begin
    Attr(dom::Ob)
end
```

and remove the "Concrete" target, so that each attribute just gets its own type. This is simpler, but I don't know if it's better; please comment.


epatters:
Thanks, these both look really interesting. I will dig into it and follow up with more detailed comments.

epatters:
Regarding the first experiment, on StructArrays, it is as you say a bit complicated, but I still like the general idea and the code could probably be simplified somewhat. I was imagining that the morphisms and data attributes with a given domain would be combined into a singlet table. For example, the edge table of an edge-weighted graph would look like:

```
julia> t = StructArray((src=[1,2,3], tgt=[2,3,1], weight=[0.5,1.0,1.5]))
3-element StructArray(::Array{Int64,1}, ::Array{Int64,1}, ::Array{Float64,1}) with eltype NamedTuple{(:src, :tgt, :weight),Tuple{Int64,Int64,Float64}}:
 (src = 1, tgt = 2, weight = 0.5)
 (src = 2, tgt = 3, weight = 1.0)
 (src = 3, tgt = 1, weight = 1.5)

julia> typeof(t)
StructArray{NamedTuple{(:src, :tgt, :weight),Tuple{Int64,Int64,Float64}},1,NamedTuple{(:src, :tgt, :weight),Tuple{Array{Int64,1},Array{Int64,1},Array{Float64,1}}},Int64}

julia> t[1]
(src = 1, tgt = 2, weight = 0.5)

julia> t.src
3-element Array{Int64,1}:
 1
 2
 3
```

From a usability perspective, this seems quite natural to me, because it would allow a C-set to treated like a little relational database, with exactly one table for every object in C.

epatters:
Regarding the other experiment: I read through a good chunk of the Algebraic Databases paper. I can't say I understand everything but I understand its perspective. To summarize, there are at least three approaches to data attributes:

1. Present an ordinary category, treating attributes like any other morphism, and distinguish the data attributes only in the implementation. This is the approach currently taken for C-sets in Catlab and it is also the perspective of Spivak's original paper "Functorial data migration." Despite its naivete, I remain sympathetic to this approach for its appealing mathematical simplicity, which makes the category of schemas equivalent to the category of small categories.

2. Extend the theory of categories with a type for attributes, like you sketched above:

```julia
@theory Category(Ob,Hom) => Schema(Ob,Hom,Attr) begin
    Attr(dom::Ob)::Type
end
```
Under this approach, every data attribute gets its own data type. This is almost exactly what is done in Spivak and Wisnesky's paper "Relational foundations for functorial data migration", Sec 2.5.

3. Extend the theory of categories with *typed* data attributes, like in your PR:

```julia
@theory Category(Ob,Hom) => Schema(Ob,Hom,DataType,Attr) begin
  DataType::TYPE
  Attr(dom::Ob,codom::DataType)::TYPE

  compose(f::Hom(A,B), g::Attr(B,X))::Attr(A,X) ⊣ (A::Ob, B::Ob, X::DataType)

   # Axioms for a left category action...
end
```

This is a simplified version of Schultz, Spivak, Vasilakopoulou, and Wisnesky's paper "Algebraic Database," where the simplification happens by omitting the "other half" of the profunctor which equips the data types with Lawvere theories.

It is interesting to note the steady increase in technicality of these papers, as the mathematics becomes more elaborate to model "real" queries in SQL databases. At the moment, I cannot make up my mind about how we should proceed.

epatters:
The idea of condensing the type parameters is clever. I hadn't thought of that and it would indeed be great to cut down on the typing. One worry is about subtyping. Currently, it is the case that if

```julia
const AbstractGraph = AbstractCSetType(TheoryGraph)
const Graph = CSetType(TheoryGraph, ...)
const PropertyGraph = CSetType(TheoryPropertyGraph, ...)
```
then `Graph <: AbstractGraph` and `PropertyGraph <: AbstractGraph`, which is very convenient. Will that still be the case under your proposal? If not, could we preserve most of its benefit by shifting the definition of `getproperty` to `CSet` itself?

epatters:
Owen, if you think what I said about StructArrays makes sense, then maybe we should move forward a complete implementation of that while we mediate on the proper handling of data attributes.

olynch:
OK, I think the three things that are monotonic improvements on the current CSet implementation would be
- StructArrays (because we have a static guarantee that each column associated with a given entity is the same length)
- Condensed type parameters (assuming that subtyping works as desired)
- Being able to have types (like `DecGraph{T}`) that have the type of their data in their type.

Here are things that are not necessarily monotonic
- Moving to a different handling of data attributes

My feeling about all of this continues to be that if we don't have a theory of the limitations of our data structure, then we are going to run into trouble. This is an argument for 2/3 over 1. A "true" implementation of a functor from C to Set would not force some objects to become finite sets and other objects to become Julia types: any object could be a finite set or a Julia type, and we would also have functions from Julia types to finite sets or Julia types. In making the distinction between data and non-data objects, you are *already* not having your category of schemas be the category of finite sets, the difference between 1 and 2/3 is that 2/3 owns up to this.

Here is my argument for the increased complexity of 3: from an implementation perspective, it doesn't take that much more work to move around the types. However, there are more operations that we can define on it, because it makes sense to talk about commutative diagrams where the last morphism is an attribute morphism. Also, the framework also allows us to have "derived" attributes, like "the age of my boss".

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,98 @@
...
+
+struct SInstance{Ts <: Tuple,S <: SchemaDesc,TablesType <: NamedTuple}
+  tables::TablesType
+  function SInstance{Ts,S}() where {Ts <: Tuple, S <: SchemaDesc}
+    sd = S()
+    tt = tables_type(Tuple(Ts.parameters),sd)
+    empty_tables = [tt.parameters[2].parameters[i](undef,0) for i in 1:length(sd.ob)]
+    new{Ts,S,tt}(tt(empty_tables))
+  end
+end
```
olynch, line 83: This is all of the important code, and I'm kind of astonished at how short it ended up being.

olynch:
I added a new implementation. This implementation probably doesn't have nice subtyping, but is otherwise quite nice (in my opinion). I left my laptop charger 7 miles away, so I'm committing this for now, but I'll flesh it out once I have a reliable source of power :P.

experiments/MoarCSets/SInstances.jl
epatters: Awesome, glad it is going well!

epatters:
@olynch, I've come around to your view on how to handle data types and attributes.

One reason why the compromise (2) is not a good choice is that it means you cannot have equations between data attributes because there is no way of ensuring they get the same types in instances. This is made explicit in "Relational foundations for data migration," where they say: "Importantly, and unlike ['Functorial data migration'], paths may not refer to attributes; they may only refer to nodes and edges." But we definitely what to allow this. For example, in typed wiring diagrams, equations between data attributes are what enforce type compatibility between ports that are wired together. 

So the choice is between (1) and (3), and I agree with you that (3) will ultimately be best. I'll write this up in a separate issue so that it doesn't get lost in this PR.

olynch:
At this point, I think that all of the core functionality for SInstances is implemented, or there is no clear stumbling block to implementing it. What's left is to fill in the details, and to talk about replacing CSets.

olynch:
I'm starting to see that the genius and the difficulty of the original CSet implementation was not in the data structure, but in the indexing... That is, writing boilerplate operations for, say, a custom graph type without indexing is not so hard, but once indexing is in the picture, there's some real code savings to be had by using CSets, and it's pretty cool.

Also, I had an idea for unique indexing, which is that we let some morphisms be partial (which is probably desirable anyways), and then add some code that special-cases a partial morphism that is declared as the one-sided inverse of another morphism. I guess this means that we might also want partial morphisms from data which are not necessarily inverses of anything, but this actually might be useful for something else. @epatters, @jpfairbanks, @mehalter, any comments/feedback welcome, on this proposal and on my code in general.

epatters:
> I'm starting to see that the genius and the difficulty of the original CSet implementation was not in the data structure, but in the indexing... That is, writing boilerplate operations for, say, a custom graph type without indexing is not so hard, but once indexing is in the picture, there's some real code savings to be had by using CSets, and it's pretty cool.

Agreed, I think that a lot of the value comes from generalizing the adjacency list representation of graphs, so that you don't have to reinvent/reimplement it every time you want a data structure that involves "parts of different kinds that fit together in a certain way."

> 
> Also, I had an idea for unique indexing, which is that we let some morphisms be partial (which is probably desirable anyways), and then add some code that special-cases a partial morphism that is declared as the one-sided inverse of another morphism. I guess this means that we might also want partial morphisms from data which are not necessarily inverses of anything, but this actually might be useful for something else. @epatters, @jpfairbanks, @mehalter, any comments/feedback welcome, on this proposal and on my code in general.

I like the general idea of describing unique indexing algebraically, so that there is a clean separation between the semantical and computational aspects of C-sets. Partial morphisms are already unofficially allowed, because morphisms are allow to take the value `0` meaning `undef` and this is handled appropriately in the code. But this hasn't been advertised and it was done mainly for convenience in implementation, so that the data doesn't have to be initialized "all at once."

Anyway, thanks for working on the C-sets. I'll review the code in this PR today. It would be helpful to update the PR description to summarize what you've done so far.

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,406 @@
...
+    ob_syms, hom_syms = nameof.(obs), nameof.(homs)
+    ob_num = ob -> findfirst(ob_syms .== ob)::Int
+    new{Tuple(ob_syms), Tuple(hom_syms),
+        Tuple(@. ob_num(nameof(dom(homs)))), Tuple(@. ob_num(nameof(codom(homs))))}()
+  end
+end
+
+CatDescType(pres::Presentation{Schema}) = typeof(CatDesc(pres))
+
+function Base.getproperty(AD::Type{T},i::Symbol) where
```
epatters, line 46: I don't know if this technically counts as ["type piracy"](https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy-1) because we "own" the type `CatDesc`, nevertheless this smells funny, especially having to call `invoke` to simulate a "super" call in a language where that pattern is more standard. I think we should consider a different approach, even if means having to make a few more keystrokes, say `typeparam(T, :ob)`.

**Update**: Instead of calling `invoke`, we can fall back to `getfield`, see this [discourse thread](https://discourse.julialang.org/t/whats-the-difference-between-fields-and-properties/12495/2). That should be an adequate fix, so we can keep this nice syntactic sugar.

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,406 @@
...
+  CD.ob[ADom[ad_attr_num(AD,attr)]]
+end
+
+function ad_codom(AD::Type{T}, attr::Symbol) where
+  {CD,Data,Attr,ADom,ACodom,T <: AttrDesc{CD,Data,Attr,ADom,ACodom}}
+  Data[ACodom[ad_attr_num(AD,attr)]]
+end
+
+function make_index(CD::Type{<:CatDesc},AD::Type{<:AttrDesc},Ts::Type{<:Tuple},Idxed::Tuple)
+    function make_idx(name)
```
epatters, line 126: Extra indentation here.

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,406 @@
...
+      if name ∈ CD.hom
+        Vector{Vector{Int}}()
+      elseif name ∈ AD.attr
+        Dict{Ts.parameters[ad_codom(AD,name)],Vector{Int}}()
+      else
+        error("invalid index parameter")
```
epatters, line 132: A nicer error message would be something like `"$name is not a morphism or data attribute: cannot create index"`

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,406 @@
...
+    SInstance{CD,AD,Ts,Idxed}()
+  end
+  function SInstance{CD,AD,Ts,Idxed,TT,IT}(tables::TT,index::IT) where
+    {CD <: CatDesc, AD <: AttrDesc{CD}, Ts <: Tuple,Idxed,TT <: NamedTuple,IT <: NamedTuple}
+    new{CD,AD,Ts,Idxed,TT,IT}(tables,index)
+  end
+end
+
+
+function fields(::Type{T}) where {T <: NamedTuple}
```
epatters, line 179: I think this custom function can be replaced with the builtin `fieldnames`:

```julia
julia> fieldnames(typeof((x=1,y=2)))
(:x, :y)
```

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,406 @@
...
+function fieldtype(::Type{T},i::Integer) where {T <: NamedTuple}
+  fieldtypes(T)[i]
+end
+
+function fieldtype(::Type{T},s::Symbol) where {T <: NamedTuple}
+  i = findfirst(fields(T) .== s)
+  fieldtype(T,i)
+end
+
+function fields(::Type{T}) where {T <: StructArray{<:NamedTuple}}
```
epatters, line 192: Can use `fieldnames` here as well:

```julia
julia> fieldnames(eltype(typeof(StructArray((a=[1, 2], b=["x","y"])))))
(:a, :b)
```

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,406 @@
...
+  end
+  for i in 1:length(Attr)
+    push!(cols[ADom[i]], (Attr[i],ts[ACodom[i]]))
+  end
+  [(col_names = first.(cols[i]), col_types = last.(cols[i])) for i in 1:length(Ob)]
+end
+
+abstract type AbstractSInstance{CD,AD,Ts} end
+
+struct SInstance{CD <: CatDesc, AD <: AttrDesc{CD}, Ts <: Tuple, Idxed,
```
epatters, line 153: Not sure what would be the best name, but I'm thinking that `SchemaInst` would be more memorable than `SInstance`. Or we could really go for it and use `DBInstance`, in which case we might also rename `Schema` to `DBSchema`. What do you think?

experiments/MoarCSets/SInstances.jl
```diff
@@ -0,0 +1,406 @@
...
+  deleteat!(a, i)
+end
+
+add_part!(ins::SInstance,type::Symbol,subparts::NamedTuple) =
+  _add_parts!(ins,Val(type),StructArray([subparts]))[1]
+add_parts!(ins::SInstance,type::Symbol,subpartses::StructArray{<:NamedTuple}) =
+  _add_parts!(ins,Val(type),subpartses)
+
+@generated function _add_parts!(ins::SInstance{CD,AD,Ts,Idxed,TT},::Val{ob},
+                                subpartses::T) where
```
epatters, line 260: You don't have to change it, but when I read `subpartses` and I have to read it [like this](https://lotr.fandom.com/wiki/Gollum).

olynch: Yeah, that was intentional :P

olynch: Tasty subpartses.....

experiments/MoarCSets/SInstances.jl
olynch: I don't like `DBInstance`; I don't want people thinking of these things as databases because people think of databases as storing data long term, and normally you only have one database. `SchemaInst` is nice, if a bit long and `Inst` is kind of hard to pronounce. I have to say, `CSet` is a really nice name, but this isn't quite it...

We could call it `DCSet` for "decorated CSet"? Let me keep thinking.

experiments/MoarCSets/SInstances.jl
olynch: Oh yeah, that makes sense.

experiments/MoarCSets/SInstances.jl
olynch: Yeah, sounds good to me, also this reminds me of: https://meta.wikimedia.org/wiki/Cunningham%27s_Law

olynch:
Another thing: it might be interesting to provide an alternate implementation of this same interface that uses JuliaDB under the hood. Putting this down so I don't forget about it and to toss it out for you to think about, but not a priority now.

experiments/MoarCSets/SInstances.jl
epatters: Haha, I didn't know that one but it's so true!

experiments/MoarCSets/SInstances.jl
epatters: Not sure about the general name, but at the very least we should have type aliases

```
const AbstractCSet{CD} = AbstractSchemaInst{CD,(),()}
const CSet{CD} = SchemaInst{CD,(),()}
```

but these statements are not misleading.

epatters:
We should definitely think about it. Especially after this PR switches to `StructArrays`, we should have an internal data structure that is very similar to a collection of interlinked `IndexedTable`s from JuliaDB, so interoperability will hopefully be very natural.

experiments/MoarCSets/SInstances.jl
olynch: We might want to add the index parameters to `CSet`, but yeah, I like that a lot, and that allows us to write all of the limits and colimits for `CSet`s!

olynch:
Can you all review this semi-final version before I replace CSets with it?

olynch:
I can't reproduce the test failure locally....

epatters:
I can reproduce it on master even with this:

```julia
julia> @theory Category(Ob,Hom) => Schema(Ob,Hom,Data,Attr) begin
         Data::TYPE
         Attr(dom::Ob,codom::Data)::TYPE
       end
ERROR: LoadError: MethodError: no method matching basename(::Missing)
Closest candidates are:
  basename(::AbstractString) at path.jl:176
Stacktrace:
 [1] default_group(::Missing) at ./logging.jl:260
 [2] macro expansion at ./logging.jl:320 [inlined]
 [3] parse_theory_binding(::Expr) at /Users/epatters/Local/compositionality/catlab/src/core/GAT.jl:206
 [4] parse_theory_head(::Expr) at /Users/epatters/Local/compositionality/catlab/src/core/GAT.jl:199
 [5] theory_builder(::Expr, ::Expr; signature::Bool) at /Users/epatters/Local/compositionality/catlab/src/core/GAT.jl:136
 [6] theory_builder(::Expr, ::Expr) at /Users/epatters/Local/compositionality/catlab/src/core/GAT.jl:136
 [7] @theory(::LineNumberNode, ::Module, ::Any, ::Any) at /Users/epatters/Local/compositionality/catlab/src/core/GAT.jl:102
in expression starting at REPL[4]:1
```

However, this works:

```julia
julia> @theory Schema{Ob,Hom,Data,Attr} <: Category{Ob,Hom} begin
         Data::TYPE
         Attr(dom::Ob,codom::Data)::TYPE
       end
Schema
```

So, I don't know for sure, but there might be two separate issues here:

1. A bug in #237
2. A CI bug where the experiments are running on Catlab master (however I don't see why that would be the case)

olynch:
OK, I rebased this on top of a fix for a bug in #237, and it now passes the tests. I'm going to start replacing CSets now ("speculative execution" predicated on there not being too much wrong with my ACSets implementation).

olynch:
OK, I'm starting to think that there might be something strange going on with the CI, or with how Julia tests work, because when I run this with:

```julia
>]activate .

>include("test/runtests.jl")
```

it passes all the tests (I've commmented out everything past CategoricalAlgebra). However, when I run it with `]test Catlab`, it fails in the same way as it does in the CI.

epatters:
I can confirm that the tests pass on my local system when run as

```
julia --project=test test/runtests.jl
```

(which is how they're supposed to run on CI) and they also pass when run as

```
julia --project=. test/runtests.jl
```

(since the later tests with extra dependencies are currently commented out). I will try to figure out what is going on with the CI.

epatters:
Here is something from the CI log that doesn't look right to me:

```
Checking out the ref
/usr/bin/git log -1
 commit baa82fd5433b009cad88a14e42bbc36d902b42af
Author: olynch <oclynch888@gmail.com>
Date:   Sat Aug 15 02:47:56 2020 +0000

    Merge d91a30df359253b4b375c1e8219a5e05465220b3 into 1c33bec549542cc89b38be689d5e551b5171fdb9
```

A few weeks ago I had a problem where GitHub CI didn't pull the right commit after I force-pushed to a branch. After a lot of frustration I resolved the problem by making a trivial amendment to the commit and force-pushing again. Can you try this?

olynch:
I tried making a trivial amendment and forcepushing, to no avail.

epatters:
Dang. I'll look into it further.

olynch:
Again, burned by Julia 1.0... OK I'll look at this tomorrow.

epatters:
The Julia v1.0 fix should be as simple as `using Compat: isnothing`.

epatters:
Great, I think this is basically ready! Just a few requests and it should be ready to merge:

- Remove the dependency on `LabelledArrays`, which has been replaced with `StructArrays` in this PR
- Add a `[compat]` entry for `StructArrays`, as required by the Julia package repository
- Use the longer names `AttributedCSet` and `AbstractAttributedCSet` as the official type names, with exported aliases `const ACSet = AttributedCSet` and `const AbstractACSet = AbstractAttributedCSet` for convenience
- Keep the module name as `Catlab.CategoricalAlgebra.CSets` for now, since the name `CSets` will be somewhat more familiar than `ACSets`.

epatters:
Thanks very much for all your work on this! Merging.

Issue #194, opened 2020-07-13
Functor types


jpfairbanks:
Why don't we create an abstract type Functor that has the API:

```julia
dom(F::Functor)::Theory
codom(F::Functor)::Theory
function (F::Functor)(x::Ob{:generator})
    ...
end
function (F::Functor)(f::Hom{:compose})
    foldl(compose, map(x->F(x), f.args))
end
function (F::Functor)(f::Hom{:generator})
    ...
end
```

The idea being that the functor type creator just has to specify what happens on the generators of the theory and the term constructors. Then the `MonoidalFunctor <: AbstractFunctor` would have the following additional method definitions:

```julia
function (F::MonoidalFunctor)(x::Ob{:otimes})
    foldl(otimes, map(F, x.args))
end
function (F::MonoidalFunctor)(f::Hom{:compose})
    foldl(compose, map(F, f.args))
end
function (F::MonoidalFunctor)(f::Hom{:otimes})
    foldl(otimes, map(F, f.args))
end
```

In this setting, in order to make a functor, you would just have to define a type `MyFunctor<:MonoidalFunctor` which supplies the action on the generators. We could supply a `PresentedMonoidalFunctor` which works like the current implementation by storing a dictionary mapping the generators to their images under the functor.

```julia
struct PresentedMonoidalFunctor{K,V} <: MonoidalFunctor
    g::Dict{K,V}
end
function (F::PresentedMonoidalFunctor)(x::Ob{:generator})
    F.g[x]
end
function (F::PresentedMonoidalFunctor)(f::Hom{:generator})
    F.g[x]
end
```

I think we could simplify a lot of this code if we store the GATExprs with the
operation as a `Function` instead of a `Symbol`. The current implementation looks
like:

```julia
julia> X = Ob(FreeSymmetricMonoidalCategory, :X)
julia> dump(X)
Catlab.Theories.FreeSymmetricMonoidalCategory.Ob{:generator}
  args: Array{Symbol}((1,))
    1: Symbol X
  type_args: Array{GATExpr}((0,))
```

Instead of storing `head(ex) == :otimes` we could store the function `otimes`, this would allow a single abstract functor type with the idea that the codomain category will support all the term constructors of the domain category so that you can implement arbitrary GAT functors:

```julia
function (F::Functor)(x)
    foldl(head(x), map(F, x.args))
end

#you still need a base case for generators
Generator = Union{Ob{:generator}, Hom{:generator}}
[...]

PR #220, opened 2020-07-28
Functors: Turns out they are pretty simple

olynch:
I implemented @jpfairbanks idea in #194. It seems like it's pretty simple and straightforward, I'm not sure if there is other stuff to do with it. Eventually, it would be cool if the declaration of `AbstractFunctor` and `AbstractMonoidalFunctor` could be done automatically, based on the theory signature, but I think that would not change the current design which is done manually.


test/Project.toml
```diff
@@ -10,6 +10,7 @@ LightXML = "9c8b4983-aa76-5018-a973-4c85ecc9e179"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LinearMaps = "7a12625a-238d-50fd-b39a-03d52299707e"
 LinearOperators = "5c8ed15e-5a4c-59e4-a42b-c7e8811fb125"
+Match = "7eb4fadd-790c-5f42-8a69-bfa0b872bfbf"
```
epatters, line 13: We're using MLStyle's `@match` now, so we don't want this dependency on Match (see #214).

src/theories/Functor.jl
```diff
@@ -0,0 +1,23 @@
+export AbstractFunctor, AbstractMonoidalFunctor
+
+abstract type AbstractFunctor end
+
+function dom(F::AbstractFunctor)::Category end
+function codom(F::AbstractFunctor)::Category end
+
+function (F::AbstractFunctor)(f::HomExpr{:compose})
+  compose(map(x->F(x), f.args))
```
epatters, line 9: I could be wrong, but I think we can make it even simpler by writing `map(F, f.args)`.

epatters:
Some further context, which I know @jpfairbanks is aware of but @olynch may not be: we have an existing implementation of `functor` in `core/Syntax.jl`. It's not that great, and we should eventually replace it entirely with something along these lines. However, one thing that the current implementation gets right, and that @jpfairbanks noted at the end of #194, is that you don't have to write boiler plate code for every term constructor in the GAT. It doesn't have to happen in this PR, but whatever eventually replaces the existing `functor` should not require any boilerplate either.

jpfairbanks:
Yeah I think that having a generic 

```julia
struct DictFunctor{T} <: AbstractFunctor
   dom::T
   codom::T
   generators::Dict{GATExpr, GATExpr}
end
```

 that stores the map of the generators explicitly as a dictionary would be good. If you wanted to put more of the computation into the julia runtime, then you could have subtype of AbstractFunctor that used dispatch to decide what to do for each generator. Then the method lookup table machinery in julia would be used instead of this dict. I'm not sure which performance option is better, but my understanding of julia suggests that an explicit dict would be faster for large categories and if you needed to make changes, but if you have large hom expressions, but few generators, then the julia machinery might be faster.

olynch:
OK, I just wrote a macro that automatically creates a abstract type for a "theory map", writes the necessary boilerplate for applying elements of this abstract type to term constructors, and also optionally implements a dict-based concrete struct that has a dict for each type of term, and uses that dict for generators.

Example usage

```julia
using Catlab.Theories, Catlab.Present

@present C1(FreeCategory) begin
  A::Ob
  B::Ob
  C::Ob
  f::Hom(A,B)
  g::Hom(A,B)
  h::Hom(B,C)
end

A,B,C = generators(C1,:Ob)
f,g,h = generators(C1,:Hom)

F = DictFunctor(C1, C1, Dict(:A => A, :B => B, :C => C), Dict(:f => f, :g => f, :h => h))
```

```julia
> F(compose(g,h))
compose(f,h)
```

The code generated by

```julia
@theory_map AbstractFunctor(Category(ObExpr, HomExpr)), dict_impl => DictFunctor
```

is

```julia
    abstract type AbstractFunctor end
    function dom(f::AbstractFunctor)::Category
        #= /g/Catlab.jl/src/core/Syntax.jl:430 =#
    end
    function codom(f::AbstractFunctor)::Category
        #= /g/Catlab.jl/src/core/Syntax.jl:431 =#
    end
    begin
        #= /g/Catlab.jl/src/core/Syntax.jl:454 =#
        function (F::AbstractFunctor)(x::HomExpr{:id})
            #= /g/Catlab.jl/src/core/Syntax.jl:455 =#
            id(map(F, x.args))
        end
    end
    begin
        #= /g/Catlab.jl/src/core/Syntax.jl:454 =#
        function (F::AbstractFunctor)(x::HomExpr{:compose})
            #= /g/Catlab.jl/src/core/Syntax.jl:455 =#
            compose(map(F, x.args))
        end
    end
    begin
        struct DictFunctor <: AbstractFunctor
            dom::Any
            codom::Any
            ob_map::Dict{Symbol, ObExpr}
            hom_map::Dict{Symbol, HomExpr}
        end
        begin
            #= /g/Catlab.jl/src/core/Syntax.jl:474 =#
            function (F::DictFunctor)(x::ObExpr{:generator})
                #= /g/Catlab.jl/src/core/Syntax.jl:475 =#
                F.ob_map[x.args[1]]
            end
        end
        begin
            #= /g/Catlab.jl/src/core/Syntax.jl:474 =#
            function (F::DictFunctor)(x::HomExpr{:generator})
                #= /g/Catlab.jl/src/core/Syntax.jl:475 =#
                F.hom_map[x.args[1]]
            end
        end
    end
```

There are definitely some rough spots, but I wanted to share my progress so far.

olynch:
Can I have some feedback on this code/direction for this to go in?

epatters:
Sorry, this fell off my radar. I'll get back to it soon.

jpfairbanks:
I think with a test for FreeSMC in addition to FreeCat, this should be ready. Can't wait to update all our examples to use the DictFunctor!

olynch:
Oh no.... it looks like I have accidentally depended on something that was introduced post-1.0: "cannot add methods to an abstract type". What does this mean?

olynch:
It looks like there isn't really a workaround for 1.0 if we want to do this the way we are doing it. The alternative would be to define a function that "calls" the functor, so instead of `F(x)`, we would have `call_functor(F,x)`. I think I'm going to change to that, and I'll open an issue to move to `F(x)` once we can drop 1.0 compatibility.

olynch:
I'm really stumped by the following error message.

```
Functors: Error During Test at /g/Catlab.jl/test/theories/Functor.jl:36
  Test threw exception
  Expression: fmap(F1, compose(g, h)) == compose(f, h)
  MethodError: no method matching fmap(::Main.TestTheories.TypeF1, ::Catlab.Theories.FreeCategory.Hom{:generator})
  Closest candidates are:
    fmap(!Matched::Catlab.Theories.DictFunctor, ::Catlab.Theories.HomExpr{:generator}) at /g/Catlab.jl/src/core/Syntax.jl:484
    fmap(!Matched::Catlab.Theories.DictSymmetricMonoidalFunctor, ::Catlab.Theories.HomExpr{:generator}) at /g/Catlab.jl/src/core/Syntax.jl:484
  Stacktrace:
   [1] (::getfield(Catlab.Theories, Symbol("##4#6")){Main.TestTheories.TypeF1})(::Catlab.Theories.FreeCategory.Hom{:generator}) at /g/Catlab.jl/src/core/Syntax.jl:464
   [2] iterate at ./generator.jl:47 [inlined]
   [3] _collect at ./array.jl:632 [inlined]
   [4] collect_similar(::Array{Catlab.Theories.FreeCategory.Hom{:generator},1}, ::Base.Generator{Array{Catlab.Theories.FreeCategory.Hom{:generator},1},getfield(Catlab.Theories, Symbol("##4#6")){Main.TestTheories.TypeF1}}) at ./array.jl:561
   [5] map(::Function, ::Array{Catlab.Theories.FreeCategory.Hom{:generator},1}) at ./abstractarray.jl:1987
   [6] fmap(::Main.TestTheories.TypeF1, ::Catlab.Theories.FreeCategory.Hom{:compose}) at /g/Catlab.jl/src/core/Syntax.jl:464
   [7] top-level scope at none:0
   [8] include at ./boot.jl:317 [inlined]
   [9] include_relative(::Module, ::String) at ./loading.jl:1044
   [10] include at ./sysimg.jl:29 [inlined]
   [11] include(::String) at /g/Catlab.jl/test/theories/Theories.jl:1
   [12] macro expansion at /g/Catlab.jl/test/theories/Theories.jl:26 [inlined]
   [13] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
   [14] top-level scope at /g/Catlab.jl/test/theories/Theories.jl:26
```

When I run `methods(fmap)`, I get

```
10 methods for generic function fmap:

    fmap(T::AbstractTheoryMap, x::Expr) in Catlab.Syntax at /g/Catlab.jl/src/core/Syntax.jl:411
    fmap(F::Catlab.Theories.AbstractFunctor, x::Catlab.Theories.HomExpr{:id}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Catlab.Theories.AbstractFunctor, x::Catlab.Theories.HomExpr{:compose}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Catlab.Theories.AbstractSymmetricMonoidalFunctor, x::Catlab.Theories.HomExpr{:id}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Catlab.Theories.AbstractSymmetricMonoidalFunctor, x::Catlab.Theories.HomExpr{:compose}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Catlab.Theories.AbstractSymmetricMonoidalFunctor, x::Catlab.Theories.ObExpr{:otimes}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Catlab.Theories.AbstractSymmetricMonoidalFunctor, x::Catlab.Theories.HomExpr{:otimes}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Catlab.Theories.AbstractSymmetricMonoidalFunctor, x::Catlab.Theories.ObExpr{:munit}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Catlab.Theories.AbstractSymmetricMonoidalFunctor, x::Catlab.Theories.HomExpr{:braid}) in Catlab.Theories at /g/Catlab.jl/src/core/Syntax.jl:464
    fmap(F::Main.TestTheories.TypeF1, morph::Catlab.Theories.HomExpr{:generator}) in Main.TestTheories at /g/Catlab.jl/test/theories/Functor.jl:27
```

which includes a method that should be selected... Is there something that I'm missing?

src/core/Syntax.jl
```diff
@@ -406,6 +406,90 @@ function functor(types::Tuple, expr::GATExpr;
   invoke_term(theory_type, types, name, term_args...)
 end
 
+abstract type AbstractTheoryMap end
+
+function fmap(T::AbstractTheoryMap, x::Expr) end
```
epatters, line 411: Should be `x::GATExpr` instead of `x::Expr`. That's probably why you're getting the error.

olynch: I changed it, and it didn't seem to fix the error.

epatters:
OK, I don't know what the problem is, but it probably has to do with the generated code, since the stack trace points to the definition of `fmap` in `map_term_constructor`.

src/core/Syntax.jl
```diff
@@ -406,6 +406,90 @@ function functor(types::Tuple, expr::GATExpr;
   invoke_term(theory_type, types, name, term_args...)
 end
 
+abstract type AbstractTheoryMap end
+
+function fmap(T::AbstractTheoryMap, x::GATExpr) end
```
jpfairbanks, line 411: Is fmap the right name for this? If we think of functors as containers in the "polynomial functors are containers" sense in Haskell the applying a functor is a generalization of the function `map(f)= list -> [f(x) for x in list]` but we aren't bound by the naming conventions of Haskell. Maybe `fcall`? In this setting the GATExpr is the container so this should be like `gatmap(f::AbstractTheoryMap, x::GATExpr)`

olynch: I mean, ideally we wouldn't have to name this at all because we would just be overloading the function call syntax. What we call it until we can drop 1.0 support, I don't care, so I'll just follow what you suggest and call it gatmap.

Also, maybe we should rename Catlab to Gatlab :P

jpfairbanks: I think so too!

epatters: Haha, James made the same joke a while ago!

epatters:
Sorry I've been slow to get back to this one, and thanks a lot for working on it.

In addition to several unaddressed comments dating back to the original PR, I'll add this comment. Overall I like the approach of having abstract types for functors with different concrete implementations, e.g. based on dictionaries. I am concerned about two related things:

1. We already have a `functor` function in the `Syntax` submodule. I don't want to end up with two totally disjoint implementations of functors that do basically the same thing. It doesn't necessarily have to happen in this PR, but we need to converge to a single system.

2. The essential property of functors (or more generally homomorphisms of models of GATs) out of GAT expressions is that they are completely determined by their action on generators. Thus as a matter of principle, it should not be necessary to manually write code for every other term constructor (`compose`, `otimes`, etc). Practically speaking, it's not too big a deal for small theories like `Category` or `SymmetricMonoidalCategory`, but for big theories like we have in GLA, you're going to have a write a ton of boilerplate code. Although it has other issues, this is something that the current `functor` gets right: it works for any GAT, with no boilerplate required.

I'm confident there's a way to do this that will make everyone happy and get the benefits of both approaches. I'll think about it and follow up with more thoughts.

jpfairbanks:
Yeah I agree that we need to have one system for functors and that the point of this is to minimize the amount of boilerplate. Maybe we should address the fact that the GATExprs don't store julia functions in the head of each term constructor and instead have to do the `invoke_term` business https://github.com/AlgebraicJulia/Catlab.jl/blob/da743c821cfa4923cb2f283b16d64a11ced2548c/src/core/GAT.jl#L676
before we implement the generic functor refactor. If we stored julia functions in the GATExpr, would you need to implement boilerplate for this method?

epatters:
You mentioned that possibility before and I think it's a really good idea. If we switch from symbols to functions, we can avoid the ugly and slow reflection performed by `invoke_term`. I like the idea of addressing this first, then revamping the `functor` interface along the lines of this PR, but now without the boilerplate problem.

olynch:
>     2. The essential property of functors (or more generally homomorphisms of models of GATs) out of GAT expressions is that they are completely determined by their action on generators. Thus, as a matter of principle, you shouldn't have to manually write code for every other term constructor (`compose`, `otimes`, etc). Practically speaking, it's not too big a deal for small theories like `Category` or `SymmetricMonoidalCategory`, but for big theories like we have in GLA, you're going to have a write a ton of boilerplate code. Although it has other issues, this is something that the current `functor` gets right: it works for any GAT, with no boilerplate required.

I'm pretty sure that `@theory_map` writes that boilerplate automatically... Unless I'm misunderstanding your point.

epatters:
Oh, maybe I'm misunderstanding the code, but I was referring to lines like

```julia
function (F::AbstractMonoidalFunctor)(f::ObExpr{:otimes})
  otimes(map(x->F(x), f.args))
end
```

which looked like boilerplate. Let's discuss further today to make sure we're on the same page.

olynch:
Let's resurrect this PR. In response to your question:

```julia
function (F::AbstractMonoidalFunctor)(f::ObExpr{:otimes})
  otimes(map(x->F(x), f.args))
end
```

is in fact automatically generated by @theory_map.

epatters:
I would also like to resurrect this PR! However, I'd suggest that we postpone that until after our current push on ACSets is finished, hopefully early next year. At that point, I'd like to make a big effort to overhaul how functors are handled in Catlab and the ideas in this PR should certainly be part of that.

jpfairbanks:
@epatters, is this PR superceded by the recent FinFunctors PRs?

epatters:
Only partly. With regard to the category of categories, what is now in Catlab supersedes this PR and goes a fair bit beyond it, implementing the full 2-category structure of Cat. OTOH, this PR made a start on doing things more generically, with types for monoidal functors and symmetric monoidal functors.

I suppose there are tradeoffs between coding everything by hand, like I've done for FinCat, and taking a more automated approach based on the GATs, as in this PR. Given the fundamental role played by Cat, I think it's reasonable to have a dedicated implementation for FinCat, just like we do for FinSet, even if there is a certain amount of boilerplate. However, when we turn to the full menagerie of 2-categories of categories with extra structure ("doctrines"), we're going to need a more automated approach.

I will close this PR for now, but we should revisit these ideas, especially in conjunction with the (eventual) rewrite of the GAT system.

Issue #223, opened 2020-07-30
Category of Restricted Types


olynch:
I propose that we have a category whose objects are `RestrictedTypes`

``` julia
struct RestrictedType
  T::DataType
  filter::Function # : T -> Bool
end
```

and where a morphism between `a` and `b` is a function `f` from `a.T` to `b.T` such that if `x : a.T` then `a.filter(x)` implies `b.filter(f(x))`.

This actually has products and coproducts (but not equalizers or coequalizers).

I can implement this in probably an hour if you all think it's a good idea. The inspiration for this was `PetriDecorator` in AlgebraicPetri.

``` julia
function (pd::PetriDecorator)(n::FinOrd)
    return p -> typeof(p) <: Petri && ns(p) == n.n
end
```

epatters:
Resolved by #224. Thanks Owen!

PR #224, opened 2020-08-01
ENH: trial implementation of DecidableSets

olynch:
Nobody commented on #223, so I went ahead and implemented it.


src/categorical_algebra/DecidableSets.jl
```diff
@@ -0,0 +1,58 @@
+module DecidableSets
+export DecidableSet,DecidableHom
+
+import FunctionWrappers
+import FunctionWrappers: FunctionWrapper
+using Catlab.GAT
```
epatters, line 6: These should be relative imports, see the other submodules in this directory.

epatters, line 6: These should be relative imports, see the other submodules in this directory.

epatters:
Thanks, merging!

PR #235, opened 2020-08-07
Refactor interface for free diagrams and (co)limits

### LINKED ISSUE #171: Interface for diagram shapes and limits
James and I had a quick chat and discussed a potential interface for the CategoricalAlgebra.ShapeDiagrams where we have the base shapes of Span, Cospan, Product, Coproduct, Equalizer, and Coequalizer defined in a way where the interface allows you to build arbitrary shape diagrams out of those building blocks. Then to solve things like pushouts and pullbacks, you just have the functions like limit, colimit, and dual on arbitrary shapes. For example if you do the `colimit(s::Span)`, you would get the cospan that solves the pushout defined by `s`. You could also do something like `limit(cs::Cospan)` give you the span that solves the pullback defined by `cs`. Futhermore you would be able to use the interface like `dual(colimit(s::Span)) == limit(cs::Span)` given that `s` and `cs` form a valid pushout.

### LINKED ISSUE #185: Pushout/pullback design
I think we should change the pushout interface to return a type that supports accessing all the parts of the pushout construction. This will prevent the kind of abstraction leak we see in AbstractPetri.

```julia
struct Pushout{S}
    span::S
    coprod::Cospan
end

Pushout(s::Span) = Pushout(s, coproduct(s))
```

with methods

```julia
u(p::Pushout) = begin
    f,g = left(p.span), right(p.span)
    ι1, ι2 = left(p.coprod), right(p.coprod)
    return coequalizer(f⋅ι1, g⋅ι2)
end

Cospan(p::Pushout) = begin
    coeq = u(.coproduct)
    return Cospan(f⋅ι1⋅coeq, g⋅ι2⋅coeq)
end
```

epatters:
Refactors the interface for free diagrams and defines the interface for limits and colimits:

- Submodule `ShapeDiagrams` renamed to `FreeDiagrams` to conform to [terminology of the nLab](https://ncatlab.org/nlab/show/free+diagram)
- Free diagrams of specific shapes: [multispans](https://ncatlab.org/nlab/show/multispan), multicospans, and [parallel morphisms](https://ncatlab.org/nlab/show/parallel+morphisms) and, as special cases, spans, cospans, and parallel pairs
- Free diagrams of arbitrary shape, based on C-set graphs
- Conversion of the specialized diagram types to general diagrams
- Data structures for limits and colimits, both general and specific, which encapsulate the diagrams that they are (co)limits over (#185)
- Consistency with GATs for categories with (co)limits (#177)
- Implements iteration interface for both free diagrams and (co)limits (#185)

Resolves #171 and resolves #185.


epatters:
@jpfairbanks @mehalter @olynch, I'd appreciate any feedback since you've all had thoughts on/worked on this part of the codebase.

olynch:
It seems to me like this just consolidates and cleans up the shapes, but fundamentally doesn't change how they are used. That is, this solves little inconsistencies, but doesn't really rethink how we are doing limits and colimits.

One crazy idea that would be a rethinking of how we do limits and colimits would be to have a type like a CSet, but for generic categories; i.e. a functor from a small, presented category to a generic category. This would be implemented by a vector of objects and a vector of homs in the generic category. Then we could have natural transformations between such types, which would again just be vectors of homs. A (co)limit would result in a natural transformation from the constant functor (which would be another implementation of the abstract interface of a functor from a small category to a generic category) to the original diagram.

But this is probably not a good use of your time to implement as it fundamentally wouldn't change much, and though it might be a good use of my time, I think I should finish other things before starting a new project.

epatters:
That's right, there is nothing radical here: just trying to clean up and improve the approach we already have.

If/when we have more general machinery describing functors into an arbitrary category, we could consider re-expressing (co)limits in the style that directly mirrors in the standard mathematical presentation, but I agree that is a project best left for another day :)

PR #237, opened 2020-08-10
More conventional Julia syntax for GAT extension

olynch:
I have implemented the change in the GAT parsing, and also converted a couple theories to the new syntax (right now, both syntaxes are supported). I can go convert the other theories, but I thought I'd get some feedback first.

### DIFF
--- a/src/core/GAT.jl
+++ b/src/core/GAT.jl
@@ -203,6 +203,8 @@ function parse_theory_head(expr::Expr)::TheoryHead
 end
 
 function parse_theory_binding(expr::Expr)::TheoryBinding
+  @warn "Using Haskell-style theory declaration with parentheses is deprecated," *
+    " use Julia-style with curly braces." _file=missing _line=missing
   @match expr begin
     Expr(:call, name::Symbol, params...) => TheoryBinding(name, params)
     _ => throw(ParseError("Ill-formed theory binding $expr"))
--- a/src/core/GAT.jl
+++ b/src/core/GAT.jl
@@ -7,6 +7,7 @@ using Base.Meta: ParseError
 using Compat
 using AutoHashEquals
 using DataStructures: OrderedDict
+using Logging
 using MLStyle: @match
 
 using ..Meta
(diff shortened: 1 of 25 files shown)


epatters:
Great, thanks for starting this! So as to avoid deprecating/removing more stuff later, let's only allow **two** syntaxes: the old one with `=>` and parentheses, and a new one with `<:` and curly braces. I guess this will require some mild restructuring of the parsing code.

And yes, please go ahead and convert the other theories! We'll also need to update the docs in `docs/src/apis/core.md`.

olynch:
I made a vim macro for converting the old syntax to the new syntax! Very satisfying. (don't worry, evil in emacs, not real vim)

epatters:
> I made a vim macro for converting the old syntax to the new syntax! Very satisfying. (don't worry, evil in emacs, not real vim)

Phew, you had me worried for a second.

src/core/GAT.jl
```diff
@@ -188,11 +189,16 @@ end
...
+  parse_either = parse_theory_binding_either
   @match expr begin
     (Expr(:call, :(=>), Expr(:tuple, bases), main)
       => TheoryHead(parse(main), map(parse, bases)))
+    (Expr(:(<:), main, Expr(:tuple,bases))
+      => TheoryHead(parse_jl(main), map(parse_jl, bases)))
     Expr(:call, :(=>), base, main) => TheoryHead(parse(main), [parse(base)])
-    _ => TheoryHead(parse(expr))
+    Expr(:(<:), main, base) => TheoryHead(parse_jl(main), [parse_jl(base)])
+    _ => TheoryHead(parse_either(expr))
```
epatters, line 201: Maybe you can avoid having to define `parse_either` by replacing this last case with two cases:

```julia
Expr(:call, _...) => TheoryHead(parse(expr))
_ => TheoryHead(parse_jl(expr))
```

olynch: Good call.

epatters: Oh, you mean in `@instance`, not `@syntax`.

olynch:
I don't know why it passes on 1.0 and 1.4 but segfaults on 1.5....

epatters:
Lately I've been seeing occasional, irreducible segfaults on GitHub CI with Julia 1.5, which is troubling. I'm hoping that the problem just resolves itself so I don't have to try to debug it. Let me try restarting the tests.

epatters:
Yep, that "fixed" it. Let's keep an eye on this and please let me know if you encounter any segfaults on your local machine.

epatters:
Anyway, looks like this is ready! Thanks @olynch.

Issue #225, opened 2020-08-02
Use FunctionWrappers.jl for FinSet and FinRel


epatters:
Specifically, in the concrete types `FinFunctionCallable` and `FinRelationCallable`.

epatters:
@olynch, this makes sense, right? If so, would you be willing to take care of it?

olynch:
Yes, I can do this.

epatters:
Awesome, thanks!

olynch:
I think that in order to do this, FinRel/FinSet has to be parametric in not only the type of storage but also the type of element. A `FunctionWrapper{Int, Tuple{Int}}` doesn't always work, because if `S` is, say, `Set{Symbol}`, it won't work. Should I do this refactor?

epatters:
I see your point. Yes, please do. I guess we should have `FinSet{S,T}` where `S` is the underlying set type and `T` is the element type. This should be backwards compatible (I'd feel bad about breaking this interface again) and still allow convenient dispatch through `FinSet{Int}` and `FinSet{<:AbstractSet}`. What do you think?

olynch:
Sounds good to me!

olynch:
It seems to me that this will end up with the category `FinSet` being the disjoint union of categories `FinSet{S,T}`. This is not necessarily a bad thing, just am pointing it out in case it is objectionable.

epatters:
Hmm, so couldn't have, say, a morphism from `FinSet{Set{Symbol}}` to `FinSet{Set{Int}}` using `FunctionWrapper{Int,Tuple{Symbol}}`?

olynch:
Well, I would have to make

```
@auto_hash_equals struct FinFunctionCallable{S1,T1,S2,T2} <: FinFunction{S1,T1,S2,T2}
  func::FunctionWrapper{T1,Tuple{T2}} # Usually `Function` but can be any Julia callable.
  dom::FinSet{S1,T1}
  codom::FinSet{S2,T2}
end
```

epatters:
OK, for now let's just assume that `FinFunction`s are always between the same `FinSet{S,T}`. We can revisit this if it becomes a problem in the future. (At the moment, I don't think anybody is using anything besides `FinSet{Int}`!)

olynch:
Sounds good to me!

epatters:
Resolved by #239.

PR #239, opened 2020-08-14
Use FunctionWrappers for FinSets and FinRelations

olynch:
This addresses #225

### DIFF
--- a/src/categorical_algebra/FinRelations.jl
+++ b/src/categorical_algebra/FinRelations.jl
@@ -6,6 +6,7 @@ export BoolRig, FinRel, FinRelation, FinRelationCallable, FinRelationMatrix,
 
 import Base: +, *
 using AutoHashEquals
+import FunctionWrappers: FunctionWrapper
 
 using ...GAT
 using ...Theories: DistributiveBicategoryRelations
@@ -44,8 +45,11 @@ Base.show(io::IO, x::BoolRig) = print(io, x.value)
 
 See also: `FinSet`.
 """
-@auto_hash_equals struct FinRel{S} <: AbstractSetOb{S}
+@auto_hash_equals struct FinRel{S,T} <: AbstractSetOb{S,T}
   set::S
+  function FinRel(i::Int)
+    new{Int,Int}(i)
+  end
 end
 
 iterable(s::FinRel{Int}) = 1:s.set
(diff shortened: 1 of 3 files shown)


src/categorical_algebra/FinRelations.jl
```diff
@@ -44,8 +45,11 @@ Base.show(io::IO, x::BoolRig) = print(io, x.value)
 
 See also: `FinSet`.
 """
-@auto_hash_equals struct FinRel{S} <: AbstractSetOb{S}
+@auto_hash_equals struct FinRel{S,T} <: AbstractSetOb{S,T}
   set::S
+  function FinRel(i::Int)
```
epatters, line 50: See comment below for `FinSet` constructor.

src/categorical_algebra/FinSets.jl
```diff
@@ -36,8 +37,11 @@ functions, through types `FinSet{S} where S <: AbstractSet`, as well as the
 skeleton of this category, through the type `FinSet{Int}`. In the latter case,
 the object `FinSet(n)` represents the set {1,...,n}.
 """
-@auto_hash_equals struct FinSet{S} <: AbstractSetOb{S}
+@auto_hash_equals struct FinSet{S,T} <: AbstractSetOb{S,T}
   set::S
+  function FinSet(i::Int)
```
epatters, line 42: This constructor should probably be outside the `struct` so we can construct other kinds of `FinSet`, something like:

```julia
FinSet(n::Int) = FinSet{Int,Int}(n)
FinSet(set::S) where {T,S<:AbstractSet{T}} = FinSet{S,T}(set)
```

src/categorical_algebra/FinSets.jl
```diff
@@ -115,48 +121,48 @@ Base.collect(f::FinFunction) = force(f).func
...
 
-compose_impl(f::FinFunction, g::FinFunction) = g ∘ f
+compose_impl(f::FinFunction{S,T}, g::FinFunction{S,T}) where {S,T} = g ∘ f
 compose_impl(f::FinFunctionVector, g::FinFunctionVector) = g.func[f.func]
 
 # Limits
 ########
 
-function product(Xs::StaticVector{0,FinSet{Int}})
+function product(Xs::StaticVector{0,FinSet{Int,Int}})
```
epatters, line 130: Since the second type parameter is basically an implementation detail, I suggest changing all these signatures to have form:

```julia
product(Xs::StaticVector{0,<:FinSet{Int}})
```

test/categorical_algebra/FinSets.jl
```diff
@@ -29,7 +29,7 @@ h = FinFunction([3,1,2], 3)
 ########
 
 # Terminal object.
-@test ob(terminal(FinSet{Int})) == FinSet(1)
+@test ob(terminal(FinSet{Int,Int})) == FinSet(1)
```
epatters, line 32: With the above changes to `terminal`, `product`, etc, we shouldn't have to change any of the unit tests.

src/categorical_algebra/FinRelations.jl
```diff
@@ -44,10 +45,12 @@ Base.show(io::IO, x::BoolRig) = print(io, x.value)
 
 See also: `FinSet`.
 """
-@auto_hash_equals struct FinRel{S} <: AbstractSetOb{S}
+@auto_hash_equals struct FinRel{S,T} <: AbstractSetOb{S,T}
   set::S
 end
 
+FinRel(i::Int) = new{Int,Int}(i)
```
epatters, line 52: Looks like a typo here with `new` instead of `FinRel`.

epatters:
Looks good, thanks!

PR #240, opened 2020-08-14
BUG: _file=missing causes error in @warn

olynch:
The deprecation warning on the old syntax errors when run; this fixes the error.


epatters:
Too bad that didn't work. Thanks for fixing it.

PR #245, opened 2020-08-31
Another approach to empty StructArrays in C-sets

epatters:
This PR implements a simpler workaround to https://github.com/JuliaArrays/StructArrays.jl/issues/148 that avoids the use of a wrapper type.

CC @olynch


olynch:
Sure, looks good to me! I forgot that Union types were a thing... Maybe a similar approach could work for changing StructArrays itself?

epatters:
The maintainers have more control so they could probably do a better solution. My best idea at the moment would be to add specialized `StructArray` constructors for the empty struct case that would assign the [`fieldarrays`](https://github.com/JuliaArrays/StructArrays.jl/blob/a84dd4efe6069675fffbb7468118eb796159e8f9/src/structarray.jl#L7) field to be a `Vector{Tuple{}}` (array of empty structs). Unless someone was depending on the old inconsistent behavior I would expect this change to be nonbreaking.

Anyway, I'll merge this now.

PR #248, opened 2020-09-01
Added more restrictive compose dependency

mehalter:
Pre v0.7 Compose is not compatible with Julia 1.0+ and its giving me issues with my package version resolving.


epatters:
Thanks Micah, hopefully this helps with the package delivery issues.

PR #252, opened 2020-09-03
Version Bump on Master

mehalter:
Version bump the master branch. Currently if you depend on Catlab v0.7.4+ and `]dev Catlab` you get a version issue.


mehalter:
Wait, this would be better by doing a merge from the branch v0.7.4 -> master and then making sure we don't delete the source branch

PR #253, opened 2020-09-03
V0.7.4 Version Bump on Master

mehalter:
Version bump the master branch. Currently if you depend on Catlab v0.7.4+ and `]dev Catlab` you get a version issue.


mehalter:
We should make sure that we don't delete the source branch when we merge this one since we have a release tagged on that branch (not sure what the effects of that would be)

epatters:
I'm confused because the master branch already has the Tikz standalone fix but it's showing up in the diff.

PR #258, opened 2020-09-04
Improved type checking of add_parts

mehalter:
This fixes the bug where type checking is too strict when adding parts to a CSet


epatters:
Thanks Micah, can you add a test case, something like

```julia
d = Dendrogram{Number}()
add_part!(d, :X, parent=0, height=10.0)
@test subpart(d, :height) == [10.0]
```

epatters (commented):
Thanks Micah, can you please add a test case, something like

```julia
d = Dendrogram{Number}()
add_part!(d, :X, parent=0, height=10.0)
@test subpart(d, :height) == [10.0]
```

mehalter:
Added a test for this case :D

epatters:
Perfect, thanks!

PR #266, opened 2020-09-09
Tidier Types for ACSets

olynch:
Right now this just cleans up the type printout for `CatDesc` and `AttrDesc`. The problem with tidying up the type for `ACSet` is that normally `ACSet` is wrapped in several layers of `UnionAll`, and so we can't actually access the type parameters some of the time. It may still be possible/desirable to change the printing for `ACSet`; comments are welcome.

### DIFF
--- a/src/theories/Schema.jl
+++ b/src/theories/Schema.jl
@@ -113,6 +113,18 @@ function codom(CD::Type{T}, hom::Symbol) where {Ob,Hom,Dom,Codom, T <: CatDesc{O
   Ob[codom_num(CD,hom)]
 end
 
+function Base.show(io::IO, CD::Type{<:CatDesc})
+  n = length(CD.hom)
+  sep = ","
+  ob_descs = map([CD.ob...]) do ob
+    hom_descs = map((1:n)[dom.(CD,1:n) .== ob]) do i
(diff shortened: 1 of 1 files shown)


olynch:
It looks like the failed test was a fluke, can someone rerun the CI?

epatters:
Just ran it again. Good ol' GitHub CI sometimes randomly fails to download stuff.

epatters:
Thanks for doing this! There is an error coming from the docs that looks genuine.

Could you please add a basic unit test demonstrating that the show methods work? Also, with this implemented, maybe we should display the `ACSet` type in the `show` method for `ACSet`, so it doesn't just say `ACSet` with no type information.

olynch:
I'm going to close this for now, because I can't figure out a good way to deal with UnionAll types. If someone has a good idea, we can come back to this. This also may be worth talking to Julia about.

epatters:
Ok, bummer. If overriding `Base.show` for types is a no-go, we could still improve the display of the constructor in `show` for instances: instead of writing `ACSet`, write `CSet{CD}` or `ACSet{CD,AD}` as appropriate with `CD` and `AD` formatted nicely using helper functions.

PR #267, opened 2020-09-11
Missing type parameters in abstract type for ACSets

epatters:
I think this commit should restore the intended behavior, which tries to strike a balance between allowing subtyping where useful and having an informative abstract type.

@mehalter, can you confirm that this fixes the problem you reported today?


mehalter:
I just did a long list of local tests with the Petri CSet types and this appears to be working correctly. Thanks for getting to this so quickly!!

epatters:
No problem. Thanks for checking it!

PR #268, opened 2020-09-12
Convert morphism expressions to UWDs

### LINKED ISSUE #218: Convert morphism expressions to undirected wiring diagrams
Convert symbolic expressions for morphisms in a bicategory of relations, or more generally in a hypergraph category (#216), into undirected wiring diagrams.

To do this, it may be helpful to create a shim around UWDs that temporarily separates the outer ports into inputs and outputs. Then the conversion can be done functorially just like for directed WDs.

epatters:
Resolves #218. Also includes miscellaneous cleanup related to UWDs.

I began this awhile ago but was blocked by the poor handling of data types in the original C-sets implementation. Thankfully that was resolved by the C-sets refactor!

### DIFF
--- a/src/programs/RelationalPrograms.jl
+++ b/src/programs/RelationalPrograms.jl
@@ -10,34 +10,33 @@ using MLStyle: @match
 
 using ...CategoricalAlgebra.CSets, ...Present
 using ...WiringDiagrams.UndirectedWiringDiagrams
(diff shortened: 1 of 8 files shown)


epatters:
@bosonbaas, this should belatedly help with the integration in https://github.com/AlgebraicJulia/AlgebraicRelations.jl/issues/6.

PR #271, opened 2020-09-15
Limits of C-sets

epatters:
Implements limits of C-sets without attributes, building on both #269 and #270. The algorithm for colimits is dual, so implementing it will be a copy-paste. However, writing the tests will take longer, so I'll leave colimits for next time.

This PR also refactors the interface for (co)limits and universal properties to be consistent across all diagram shapes. As a result, any limit of C-sets can be reduced to a limit in FinSet in a simple and generic way.

### DIFF
--- a/src/categorical_algebra/CSetMorphisms.jl
+++ b/src/categorical_algebra/CSetMorphisms.jl
@@ -101,9 +101,7 @@ force(α::ACSetTransformation) =
   dom(α::ACSetTransformation) = α.dom
   codom(α::ACSetTransformation) = α.codom
 
-  function id(X::ACSet)
-    ACSetTransformation(map(t -> id(FinSet(length(t))), X.tables), X, X)
-  end
+  id(X::ACSet) = ACSetTransformation(map(id, finsets(X)), X, X)
 
   function compose(α::ACSetTransformation, β::ACSetTransformation)
     # Question: Should we incur cost of checking that codom(β) == dom(α)?
@@ -111,6 +109,8 @@ force(α::ACSetTransformation) =
                         dom(α), codom(β))
   end
 end
+
+finsets(X::ACSet) = map(table -> FinSet(length(table)), X.tables)
 
 # Limits and colimits
 #####################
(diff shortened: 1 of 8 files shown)


test/categorical_algebra/CSetMorphisms.jl
```diff
@@ -37,6 +37,74 @@ add_edges!(h, [1,2], [2,1])
...
+ϕ = CSetTransformation((V=[1,2], E=[1,2]), g, h)
+ψ = CSetTransformation((V=[1,2], E=[1,3]), g, h)
+@test is_natural(ϕ) && is_natural(ψ)
+eq = equalizer(ϕ, ψ)
+@test ob(eq) == I
+@test force(incl(eq)[:V]) == FinFunction([1,2])
+@test force(incl(eq)[:E]) == FinFunction([1], 2)
+
+# Pullback in Graph from (Reyes et al 2004, p. 53).
+# This test exposes an error in the text: using their notation, there should be
```
epatters, line 92: Products and pullbacks of C-sets can be unintuitive compared to coproducts and pushouts. If the authors of this textbook had a program for computing limits, they would have caught this error!

src/categorical_algebra/CSetMorphisms.jl
```diff
@@ -110,4 +110,69 @@ force(α::ACSetTransformation) =
...
+    Yf = universal(limits[d], Multispan(ob(limits[c]), Yfs))
+    set_subpart!(Y, f, collect(Yf))
+  end
+  πs = pack_components(map(legs, limits), map(X -> Y, Xs), Xs)
+  Limit(diagram, Multispan(Y, πs))
+end
+
+""" Diagram in C-Set → named tuple of diagrams in FinSet
+"""
+unpack_diagram(diagram::DiscreteDiagram{<:AbstractCSet}) =
```
olynch, line 141: I like this design; very slick.

olynch:
I would totally read a "Evan Patterson's Thirteen Top Productivity Tips" listicle on buzzfeed.

src/categorical_algebra/CSetMorphisms.jl
epatters: Thanks!

epatters:
Tip 1: Live in a place where you cannot leave your home due to both COVID-19 *and* toxic smokey air, so that you have nothing better to do than code all day.

PR #278, opened 2020-09-19
Double Category of Squares

jpfairbanks:
The goal of this PR is to implement the double category of commutative squares over any category. Starting off with

- [x] Implement Square as (left, right, top bottom) as a subtype of FreeDiagram.
- [x] Define \circ_1, \circ_2 as horizontal and vertical composition
- [x] Make a @theory DoubleCategory with
    - Ob, Hom1, Hom2, Cell for the 0,1,2 cells respectively
    - Term constructors compose1, compose2 for the 2 types of 2-cell compositions
    - Include 2 copies of the definition of category for the two types of 1-cells
- [x] Make an instance @instance DoubleCategory(Category.Ob, Category.Hom, Category.Hom, Square)
I should harmonize naming with 2 categories signature of the theory of 2-categories https://github.com/AlgebraicJulia/Catlab.jl/blob/f218d45f6f85e113eede4247bf3d63fe63751b81/src/theories/Category.jl#L82

### DIFF
--- a/src/categorical_algebra/FreeDiagrams.jl
+++ b/src/categorical_algebra/FreeDiagrams.jl
@@ -11,13 +11,14 @@ export AbstractFreeDiagram, FreeDiagram, FixedShapeFreeDiagram, DiscreteDiagram,
   SMultispan, SMulticospan, ParallelPair, ParallelMorphisms,
   ob, hom, dom, codom, apex, legs, feet, left, right,
   nv, ne, src, tgt, vertices, edges, has_vertex, has_edge,
-  add_vertex!, add_vertices!, add_edge!, add_edges!
+  add_vertex!, add_vertices!, add_edge!, add_edges!,
+  SquareDiagram, hcompose, vcompose
 
 using AutoHashEquals
 using StaticArrays: StaticVector, SVector, @SVector
 
 using ...Present, ...Theories, ...CSetDataStructures, ...Graphs
-import ...Theories: ob, hom, dom, codom
+import ...Theories: ob, hom, dom, codom, ⋅, top, bottom, left, right
 using ...Graphs.BasicGraphs: TheoryGraph
 
 # Diagrams of fixed shape
@@ -177,6 +178,93 @@ Base.lastindex(para::ParallelMorphisms) = lastindex(para.homs)
 
 allequal(xs::AbstractVector) = all(isequal(x, xs[1]) for x in xs[2:end])
 
+# SquareDiagrams form a Double category
+#######################################
+
+"""    SquareDiagram(top, bottom, left, right)
+
+creates a square diagram in a category, which forms the 2-cells of the double category Sq(C).
+The four 1-cells are given in top, bottom, left, right order, to match the GAT of a double category.
+"""
+@auto_hash_equals struct SquareDiagram{Ob,Homs<:AbstractVector} <:
+    FixedShapeFreeDiagram{Ob}
+  corners::Vector{Ob}
+  sides::Homs
+end
+
+function SquareDiagram(homs::AbstractVector)
+  length(homs) == 4 || error("Square diagrams accept exactly 4 homs, in order top, bottom, left, right")
+  obs = [dom(homs[1]), dom(homs[2]), dom(homs[4]), codom(homs[4])]
+
+  # checking well-formedness
+  # top-left share domains
+  @assert dom(homs[1]) == dom(homs[3])
+  # bottom-right share codomains
+  @assert codom(homs[2]) == codom(homs[4])
+  # left-bottom intersection
+  @assert codom(homs[3]) == dom(homs[2])
+  # top-right intersection
+  @assert codom(homs[1]) == dom(homs[4])
+
+  SquareDiagram(obs, homs)
+end
+
+SquareDiagram(top, bottom, left, right) = SquareDiagram([top, bottom, left, right])
+
+ob(sq::SquareDiagram) = sq.corners
+hom(sq::SquareDiagram) = sq.sides
+
+top(sq::SquareDiagram) = sq.sides[1]
+bottom(sq::SquareDiagram) = sq.sides[2]
+left(sq::SquareDiagram) = sq.sides[3]
+right(sq::SquareDiagram) = sq.sides[4]
+
+
+"""    hcompose(s₁, s₂)
+
+compose two squares horizontally as shown below:
+    1   -f->   3  -g->   5
+    |          |         |
+    |          |         |
+    v          v         v
+    2  -f'->   4  -g'->  6
+"""
+function hcompose(s₁::SquareDiagram, s₂::SquareDiagram)
+    @assert ob(s₁)[3] == ob(s₂)[1]
+    @assert ob(s₁)[4] == ob(s₂)[2]
+    @assert right(s₁) == left(s₂)
+
+    f = top(s₁)
+    f′= bottom(s₁)
+    g = top(s₂)
+    g′= bottom(s₂)
+    return SquareDiagram(f⋅g, f′⋅g′, left(s₁), right(s₂))
+end
+"""    vcompose(s₁, s₂)
+
+compose two squares vertically as shown below:
+    1   -->  3
+    |        |
+    |        |
+    v        v
+    2  -->   4
+    |        |
+    |        |
+    v        v
+    5  -->   6
+"""
+function vcompose(s₁::SquareDiagram, s₂::SquareDiagram)
+    @assert ob(s₁)[2] == ob(s₂)[1]
+    @assert ob(s₁)[4] == ob(s₂)[3]
+    @assert bottom(s₁) == top(s₂)
+
+    f = left(s₁)
+    f′= right(s₁)
+    g = left(s₂)
+    g′= right(s₂)
+    return SquareDiagram(top(s₁), bottom(s₂), f⋅g, f′⋅g′)
+end
+
 # General diagrams
 ##################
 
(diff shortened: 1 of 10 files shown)


jpfairbanks:
@epatters, I think I found a bug in the `@syntax` macro. Commit 11e364 causes Catlab to think that I am overloading the term constructor for `Hom`, which doesn't make sense.

jpfairbanks:
@epatters, I implemented `Sq(FinSet)` as an instance of double category, but it would be awesome if this construction could be generalized to arbitrary `Sq(C)`. That seems like it would need to be a function from instances of `Category` to instances of `DoubleCategory` that is based on `Sq(C) = @instance DoubleCategory{Ob(C), Hom(C), Hom(C), SquareDiagram} begin ... end`. This is so many levels of types and metaprogramming it is actually scary. How would that even work?

olynch:
Do we have $\Cat$ as a category yet?

jpfairbanks:
I don't think so. What julia types would you use to represent Ob(Cat), Hom(Cat)? I think this is where we run into "implementing CT vs using CT to model the implementation."

src/categorical_algebra/FreeDiagrams.jl
```diff
@@ -175,6 +176,93 @@ Base.lastindex(para::ParallelMorphisms) = lastindex(para.homs)
...
+# SquareDiagrams form a Double category
+#######################################
+"""    SquareDiagram(top, bottom, left, right)
+
+creates a square diagram in a category, which forms the 2-cells of the double category Sq(C).
+The four 1-cells are given in top, bottom, left, right order, to match the GAT of a double category.
+"""
+@auto_hash_equals struct SquareDiagram{Ob,Homs<:AbstractVector} <:
+    FixedShapeFreeDiagram{Ob}
+  corners::Vector{Ob}
```
olynch, line 188: Why not a `StaticVector` of length 4?

src/theories/Category.jl
```diff
@@ -121,3 +122,106 @@ end
...
+  # identity two cell on 1 object
+  id2(X::Ob)::Hom2(X→X, X→X, X↓X, X↓X)          ⊣ (X::Ob)
+  id2(X) == id2(idH(X), idH(X), idV(X), idV(X)) ⊣ (X::Ob)
+  # identity two cell from a HomH
+  id2(f::(X→Y))::Hom2(X→Y, X→Y, X↓X, Y↓Y) ⊣ (X::Ob, Y::Ob)
+  id2(f) == Hom2(f, f, idV(X), idV(Y))    ⊣ (X::Ob, Y::Ob, f::(X→Y))
+  # identity two cell from a HomV
+  id2(f::(X↓Y))::Hom2(X→X, Y→Y, X↓Y, X↓Y) ⊣ (X::Ob, Y::Ob)
+  id2(f) == Hom2(idH(X), idH(Y), f, f)    ⊣ (X::Ob, Y::Ob, f::(X→Y))
+
```
olynch, line 179: No associativity/ unitality for 2-cell composition? This GAT is already pretty brutal, so I can understand why, but maybe add a comment? Though, we have to think a bit about weakness, what is the Catlab policy on weakness?

jpfairbanks: Yeah I didn't know what the right answer was for that so I left it out. What are the important issues about weakness that you can see?

mehalter:
This is looking pretty good! The GAT is pretty gnarly. I'm not quite sure how we should handle the typing issue here for the generalized case. Also for the `id2` function where you can't tell the difference between horizontal and vertical if they are the same type, I think we should have 2 different names for those because I can't think of another way to fix the issue where the types are the same.

jpfairbanks:
Should we use `id2H` and `id2V` or `idH2` and `idV2`?

mehalter:
ooooof that's a tricky one. I think I prefer `id2H` and `id2V` because the `H` and `V` describe an `id2` not the 2 describing `idH` and `idV`. Also as I type that out I might have flawed logic

mehalter:
Also before we merge this we should definitely get some tests for the instance code in there:

https://codecov.io/gh/AlgebraicJulia/Catlab.jl/pull/278/diff#D4-20

olynch:
> I don't think so. What julia types would you use to represent Ob(Cat), Hom(Cat)? I think this is where we run into "programming with CT vs using CT to model programming.

Ob(Cat) would be pairs of (ob,hom), and Hom(Cat) would be types that implement Functor? A bit of a hack, but it might be useful.

jpfairbanks:
Yeah we could make abstract types, `AbstractCategory` and `AbstractFunctor` and then wrapper structs like: `CatOb{Ob, Hom}` and `CatHom{C<:AbstractCategory}` if you define an instance of Category, you can put your obs and homs into a CatOb wrapper in order to make it an object in the Category of Categories? We have to engineer around the fact that julia doesn't have traits yet. So we can't say the `Union{T where T is an instance of Category}` We could also just live that dynamic dispatch lifestyle and say "if you want to call `Sq(C)` you had better make sure that `Ob(C), Hom(C)` are types and form an instance of `Category`." else, you will get mysterious runtime errors.

src/theories/Category.jl
epatters: So far weakness is not an issue that we've had to address in Catlab. All the theories for monoidal categories are strict. IMO, we should not worry about this until it actually becomes a problem. Typically we strictify everything, e.g. working with the skeleton of FinSet rather than Set.

src/categorical_algebra/FreeDiagrams.jl
epatters: Or named tuples with descriptive field names, or even just fields? IIRC, @jpfairbanks said in Zulip that he didn't want to always work with a conventional order of corners/sides. I'd agree with that.

src/theories/Category.jl
```diff
@@ -121,3 +122,106 @@ end
...
+    (⋅) := composeH
+    (⋆) := composeV
+  end
+
+  idH(A::Ob)::(A → A) ⊣ (A::Ob)
+  idV(A::Ob)::(A ↓ A) ⊣ (A::Ob)
+  composeH(f::(A → B), g::(B → C))::(A → C) ⊣ (A::Ob, B::Ob, C::Ob)
+  composeV(f::(A ↓ B), g::(B ↓ C))::(A ↓ C) ⊣ (A::Ob, B::Ob, C::Ob)
+
+  # Category axioms for Horizontal morphisms
```
epatters, line 158: I don't see `compose` or `id` defined for `HomH` or `HomV`.

epatters:
> ooooof that's a tricky one. I think I prefer `id2H` and `id2V` because the `H` and `V` describe an `id2` not the 2 describing `idH` and `idV`. Also as I type that out I might have flawed logic

I'd also prefer `id2H` and `id2V` to `idH2` and `idV2`, but if we assume that 0-cells never have the same type as 1-cells (and they shouldn't) we could even just use `idV` and `idH` again.

src/theories/Category.jl
```diff
@@ -121,3 +122,106 @@ end
...
+
+  # Category axioms for Vertical morphisms
+  ((f ⋅ g) ⋅ h == f ⋅ (g ⋅ h)
+    ⊣ (A::Ob, B::Ob, C::Ob, D::Ob, f::(A ↓ B), g::(B ↓ C), h::(C ↓ D)))
+  f ⋅ idV(B) == f ⊣ (A::Ob, B::Ob, f::(A ↓ B))
+  idV(A) ⋅ f == f ⊣ (A::Ob, B::Ob, f::(A ↓ B))
+
+  # identity two cell on 1 object
+  id2(X::Ob)::Hom2(X→X, X→X, X↓X, X↓X)          ⊣ (X::Ob)
+  id2(X) == id2(idH(X), idH(X), idV(X), idV(X)) ⊣ (X::Ob)
```
epatters, line 172: These two lines should be replaced with a single line:

```julia
id2(X::Ob)::Hom2(idH(X), idH(X), idV(X), idV(X)) ⊣ (X::Ob)
```

Likewise for the next two identity constructors.

epatters:
> @epatters, I implemented `Sq(FinSet)` as an instance of double category, but it would be awesome if this construction could be generalized to arbitrary `Sq(C)`. That seems like it would need to be a function from instances of `Category` to instances of `DoubleCategory` that is based on `Sq(C) = @instance DoubleCategory{Ob(C), Hom(C), Hom(C), SquareDiagram} begin ... end`. This is so many levels of types and metaprogramming it is actually scary. How would that even work?

The truth is that the `@instance` macro is really quite limited, both because Julia doesn't take interfaces seriously and because the constructions that we want to do, like this one, go beyond the notion of interface present in mainstream programming languages anyway. In this case, the simplest thing to do is just implement the double category interface, as defined by the GAT, for `SquareDiagram` and not worry about it declaring it as an `@instance`.

olynch:
Probably in about two years, we will have a very interesting wishlist of features, and maybe we can hand it off to some PLT person and they will make an interesting language out of it, with native support for GATs and attributed C-sets and whatnot.

jpfairbanks:
That would be a great PhD thesis topic. Some guy named Jeff implemented [a cool language](https://github.com/JeffBezanson/phdthesis) based on using types to take math seriously in PL.

jpfairbanks:
@epatters, I couldn't figure out why julia can't find [these methods](https://github.com/AlgebraicJulia/Catlab.jl/pull/278/commits/f1494ecea7509c0890f17d2fa56759e1cfbf192b#diff-84c5c63c7462fc9dea61a7970e4f124eR51).

epatters:
I'll take over this PR, fix the merge conflicts, and get it merged.

epatters:
I just noticed that this PR uses and exports both `hcompose`/`vcompose` and `composeH`/`composeV` in different places. IMO, we should only define one and use it consistently. Since most of the PR uses the latter, maybe we should just go with that.

PR #283, opened 2020-09-29
Top-level module for graphs

### LINKED ISSUE #272: Reflexive graphs
Extend the `Graphs` module with [reflexive graphs](https://ncatlab.org/nlab/show/reflexive+graph) (`ReflexiveGraph`) and symmetric reflexive graphs (`SymmetricReflexiveGraph`), plus appropriate Graphviz visualization in the `GraphvizGraphs` module.

epatters:
This PR creates a top-level module `Graphs` for the growing list of graph-like data structures supported by Catlab. Previously `Graphs` was a submodule of `CategoricalAlgebra`, which doesn't really make sense. However, `Graphs` depends on C-sets and `FreeDiagrams` depends on `Graphs`, so in order to perform this separation, the core data structure for C-sets has to be bootstrapped in its own private module before the `CategoricalAlgebra` module is defined. Note that all C-sets functionality is still available as re-exports through `CategoricalAlgebra.CSets`, as you would expect.

In addition, this PR adds support for reflexive graphs and symmetric reflexive graphs (closes #272).


epatters:
The overall diff shown by GitHub is rather useless. As you can see from the individual commits, most of the diff is due to renaming files.

PR #290, opened 2020-10-14
Copy parts between C-sets with different schemas

### LINKED ISSUE #285: Copy tables between C-Sets of different types
From our discussion this afternoon, would we be able to create a similar function to the generated function [copy_parts!](https://github.com/AlgebraicJulia/Catlab.jl/blob/master/src/categorical_algebra/CSetDataStructures.jl#L457) which can copy tables between CSets of different types?

The use-case I'm thinking of is in copying tables into ACSet instances that are derived from "parent" presentations. If I were to define a presentation as below that extends the `TypedRelationDiagram` presentation
```julia
@present TheoryQuery <: TheoryTypedRelationDiagram begin
  field::Attr(Port, Name)
end
```
I'd like to be able to have an instance `wd` based on `TheoryTypedRelationDiagram` and an instance `q` based on `TheoryQuery` and copy information from `wd` to `q` as follows

```julia
copy_parts!(q, wd, (Junction=:, Box=:, OuterPort=:, Port:))
```
Thanks!

epatters:
The function `copy_parts!` is generalized to allow copying from a C-set to a C'-set where C !=C'. By default, it copies all common parts and subparts. A helper function `copy_parts_only!, which copies only data attributes (not other subparts), is also exposed in the public API.

Resolves #285.

### DIFF
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -1,11 +1,11 @@
-""" Data structure for C-sets (copreshaves) and attributed C-sets.
+""" Data structure for C-sets (copresheaves) and attributed C-sets.
 """
 module CSetDataStructures
 export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
   AbstractACSetType, ACSetType, AbstractCSetType, CSetType,
   tables, nparts, has_part, subpart, has_subpart, incident,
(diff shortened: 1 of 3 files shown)


bosonbaas (approved):
This looks exactly like what I was thinking.
This will be very useful, thanks!

epatters:
Great, I will merge it then!

PR #297, opened 2020-10-18
Use our multigraph type for underlying graph of directed wiring diagram

epatters:
With #295 finally implemented, we can switch the underlying graph of a directed wiring diagram from a `MetaGraphs.MetaDiGraph` to an attributed C-set graph, a subtype of `Catlab.Graphs.AbstractGraph`. Besides eating our own dogfood, using a multigraph instead of a simple graph simplifies the low-level implementation of DWDs, since edges now correspond one-to-one with wires instead of edges having *lists* of associated wires.

This PR is an internal refactor that should not break backwards compatibility, with one small exception: the accessor `graph(diagram::WiringDiagram)` now returns a Catlab graph rather than a LightGraphs graph. If you want the latter, you have to call `LightGraphs.DiGraph(graph(diagram))`.


jpfairbanks (approved):
So many red lines, C-Sets are so elegant!

PR #298, opened 2020-10-19
Composite morphisms and attributes in C-set accessors

### LINKED ISSUE #294: Using composition in the indexing category for querying C-Set
These syntaxes for querying compound paths in the C-Set should work

```julia
f,g = generator.(C, [:f, :g]) 
subpart(D, X, compose(f, g)) := subpart(D, subpart(D, X, :f), :g)

subpart(D, x, [:f, :g]) == subpart(D, x, compose(f,g))

subpart(D, y, inv(f)) := incident(d, y, :f)

incident(d, y, compose(f,g))  == subpart(d, z, compose(inv(g),inv(f)))
```

epatters:
Resolves #294. See the unit tests for examples.

### DIFF
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -12,9 +12,8 @@ using Compat: isnothing, only
 using PrettyTables: pretty_table
 using StructArrays
 
-using ...Theories: Schema, FreeSchema, dom, codom,
-  CatDesc, CatDescType, AttrDesc, AttrDescType, SchemaType,
-  ob_num, hom_num, data_num, attr_num, dom_num, codom_num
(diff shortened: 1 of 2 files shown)


jpfairbanks (approved):
Looks great! For the `incident(acs, part, [:f, :g])` implementation, we should add a note about the order of composition being `incident(acs, incident(acs, part, :f), :g)` which is like `inv(f) \cdot inv(g)` which people might expect to be backwards. I wasn't sure about the order because chaining incident is less intuitive than chaining subpart.

epatters:
I went back and forth on the order for `incident` but ultimately decided that the least confusing thing was to make the order `[:f,:g]` consistent with `compose(f,g)`. I'll add a note to the docs about this.

PR #309, opened 2020-10-28
Scheduled UWDs: data structures and evaluation

epatters:
This PR adds preliminary support for scheduling and evaluating composites given by undirected wiring diagrams. The focus is on the basic data structures and the composition interface. The only scheduling algorithm provided is a trivial one producing a sequential schedule in an arbitrary order.

The unit tests illustrate this machinery in the case of tensor contraction.


src/programs/RelationalPrograms.jl
```diff
@@ -246,9 +246,6 @@ See also: [`@tensor_network`](@ref), the "inverse" to this macro.
 macro eval_tensor_network(diagram, tensor_macro)
   # XXX: We cannot use `GeneralizedGenerated.mk_function` here because packages
   # like Tullio generate code with type parameters, which GG does not allow.
-  # Thus we are stuck with `eval`. Of course, the real solution is for the Julia
-  # tensor packages to expose a tensor contraction interface that is not based
-  # on macros but on an appropriate data structure---such as wiring diagrams!
```
jpfairbanks, line 251: I didn't notice this 🔥 comment before.

src/wiring_diagrams/ScheduleUndirected.jl
```diff
@@ -0,0 +1,219 @@
+""" Scheduling and evaluation of undirected wiring diagrams.
+
+In category-theoretic terms, this module is about evaluating arbitrary
+composites of morphisms in hypergraph categories.
+"""
+module ScheduleUndirectedWiringDiagrams
```
jpfairbanks, line 6: This is an insanely long module name. Could we call it Schedules since it already lives in UndirectedWiringDiagrams and the import path would be `Catlab.WiringDiagrams.Schedules`?

src/wiring_diagrams/ScheduleUndirected.jl
```diff
@@ -0,0 +1,219 @@
...
+  subpart(x, args..., :composite_junction)
+composite_ports_with_junction(x::AbstractACSet, args...) =
+  incident(x, args..., :composite_junction)
+
+# Evaluation
+############
+
+""" Evaluate a scheduled UWD given a set of generators for the boxes.
+
+The first argument `f` should be a callabe supporting the signature
```
jpfairbanks, line 87: `callable object`?

src/wiring_diagrams/ScheduleUndirected.jl
```diff
@@ -0,0 +1,219 @@
...
+
+where `values` is a list of computed values for boxes or composites in the
+diagram; `juncs` is an equal-length list of `FinSet` functions, mapping the
+ports of each value to junctions; and `outer_junc` is a `FinSet` function
+mapping the outer ports to junctions. This interface is similar to the `ncon`
+function in the [NCON package](https://arxiv.org/abs/1402.0939) for MATLAB or
+the `ncon` function in
+[TensorOperations.jl](https://github.com/Jutho/TensorOperations.jl), except that
+the outer junctions are represented explictly by a third argument rather than
+implictly by using negative numbers in the second argument.
```
jpfairbanks, line 103: I don't think that most users of UWD will be familiar with TensorOperations.jl so this should probably be moved down in the docstring?

src/wiring_diagrams/ScheduleUndirected.jl
```diff
@@ -0,0 +1,219 @@
...
+
+```
+f(values::AbstractVector,
+  juncs::AbstractVector{<:FinFunction{Int}},
+  outer_junc::FinFunction{Int})
+```
+
+where `values` is a list of computed values for boxes or composites in the
+diagram; `juncs` is an equal-length list of `FinSet` functions, mapping the
+ports of each value to junctions; and `outer_junc` is a `FinSet` function
```
jpfairbanks, line 97: The codomain of all these finset funcs have to be the same `m` which is the number of junctions. I think this is really important part of this interface and should be explicit in the docs. Maybe we add to the docstring:

> The elements of `juncs` need to share a codomain `FinSet(m)` which is the number of junctions. This codomain is also the codomain of `outer_junc` thus encoding a cospan between the ports of the values and the outer ports.

src/wiring_diagrams/ScheduleUndirected.jl
```diff
@@ -0,0 +1,219 @@
...
+  # Evaluate diagram starting at the root, assumed to be unique. The root
+  # composite must be handled specially due to outer ports.
+  root = only(filter(c -> parent(d, c) == c, composites(d)))
+  values, juncs = eval_children(root)
+  do_eval(values, juncs, junction(d, outer=true))
+end
+
+""" Convert a scheduled UWD to a nested UWD.
+"""
+function to_nested_diagram(s::AbstractScheduledUWD)
```
jpfairbanks, line 157: I know we use `to_graphviz` and `to_hom_expr` so this is consistent, but it has bothered me for a while, we should just make these constructors for their return type. That can be a cleanup PR at some point before Catlab v1.0

src/wiring_diagrams/ScheduleUndirected.jl
```diff
@@ -0,0 +1,219 @@
...
+# Scheduling
+############
+
+""" Schedule UWD as a sequential chain of binary composites.
+
+This is the simplest possible scheduling algorithm, the equivalent of `foldl`
+for undirected wiring diagrams. Unless otherwise specified, the boxes are folded
+according to the order of their IDs, which is arbitrary.
+"""
+sequential_schedule(d::AbstractUWD) = sequential_schedule(d, boxes(d))
```
jpfairbanks, line 204: This seems like an ideal place to use multiple dispatch to select the scheduling algorithm. 

```julia
abstract type AbstractSchedule end 
struct SequentialSchedule <: AbstractSchedule end
struct BisectionSchedule <: AbstractSchedule end 

function schedule(d::AbstractUWD, alg::AbstractSchedule=SequentialSchedule())
    schedule(d, alg)
end

function schedule(d::AbstractUWD, alg::SequentialSchedule)
    ...
end

function schedule(d::AbstractUWD, alg::BisectionSchedule)
    # schedule by recursive bisection compose(1:floor(Int, n/2), floor(Int, n/2)+1:end)
end
```

src/wiring_diagrams/Undirected.jl
```diff
@@ -68,38 +68,45 @@ UndirectedWiringDiagram(port_types::AbstractVector{T}) where T =
...
     junction(d, nport, outer=true) : junction(d, ports(d, box)[nport])
 end
 
-nboxes(d::AbstractUWD) = nparts(d, :Box)
-njunctions(d::AbstractUWD) = nparts(d, :Junction)
-boxes(d::AbstractUWD) = parts(d, :Box)
-junctions(d::AbstractUWD) = parts(d, :Junction)
+nboxes(d::AbstractACSet) = nparts(d, :Box)
+njunctions(d::AbstractACSet) = nparts(d, :Junction)
+boxes(d::AbstractACSet) = parts(d, :Box)
```
jpfairbanks, line 87: Not necessary for this PR, but do we want to generate methods for every Object and Hom for the ACSets just have a convention that if you have a hom called f, then you get a julia function called `f(acs::ACSettype) = subpart(acs, :f)` and the same for objects nXs? Maybe that will lead to too many methods?

src/wiring_diagrams/WiringDiagrams.jl
```diff
@@ -8,16 +8,19 @@ include("MonoidalDirected.jl")
 include("MonoidalUndirected.jl")
 include("Algorithms.jl")
 include("Expressions.jl")
-include("Serialization.jl")
-include("GraphML.jl")
-include("JSON.jl")
+include("ScheduleUndirected.jl")
```
jpfairbanks, line 11: File name and module name don't match see earlier comment about shortening the module name.

jpfairbanks (commented):
Looks good, I had some docs and interface points to make. The biggest one is using multiple dispatch to encode the choice of scheduling algorithm so that users can extend the set of schedulers from external packages. This is patterned on how DiffEqs does `alg=Vern7()` to select a solver.

src/programs/RelationalPrograms.jl
epatters: Haha, besides being 🔥 it's also a bit unfair since functions like `ncon` take as input something that could be construed as a UWD, albeit packaged up in a strange way.

src/wiring_diagrams/ScheduleUndirected.jl
epatters: In some cases I agree, in others not. Anyway, yeah, we can defer that discussion to later.

src/wiring_diagrams/Undirected.jl
epatters: Good question. I go back and forth on it. On one hand, writing these accessors leads to boilerplate; on the other hand, we'd have to do more code gen, which is a downside in its own right, and it's also often useful to tweak the API in ways that couldn't be automatically code gen-ed.

src/wiring_diagrams/ScheduleUndirected.jl
epatters: What would be gained by using multiple dispatch here instead of just having different functions? Seems to me that this just moves the names of algorithms from functions to types. But maybe there is some advantage to that?

epatters:
Regarding submodule names, this PR follows the convention currently used in `Catlab.WiringDiagrams`. I'm not opposed to changes the names, although I figure most people just do `using Catlab.WiringDiagrams` anyway. If we do change it, we should do so in a separate PR that consistently addresses the whole module.

Also, one of my biggest Julia gripes is the code organization system based on `include`-ing files into other files—have we learned nothing from C/C++?—but let me resist the urge to get further into that.

src/wiring_diagrams/ScheduleUndirected.jl
epatters: Oh, I see your comment above about other packages defining their own scheduling algorithms. But for that they can just write functions that take a UWD and output a `ScheduledUWD`, right? Then again, I don't see why DiffEqs does this either, so maybe I am just missing something.

epatters: Another advantage is that you can specify a default scheduling algorithm. That's good enough for me, I suppose. I'll make this change.

PR #316, opened 2020-11-01
The UWD algebra of structured multicospans

epatters:
Sequel to #315 that realizes structured multicospans as an algebra of the operad of undirected wiring diagrams. This is the fully unbiased alternative to structured cospans as a hypergraph category.

A bit more testing is needed, but with the right machinery in place this is surprisingly easily to implement: just set up the appropriate diagram and compute its colimit!

@mehalter, this should replace the whole pipeline of `@program` -> `to_hom_expr` -> `functor`.

### DIFF
--- a/src/categorical_algebra/StructuredCospans.jl
+++ b/src/categorical_algebra/StructuredCospans.jl
@@ -130,7 +130,7 @@ begin
   end
 
   munit(::Type{StructuredCospanOb{L}}) where L =
-    StructuredCospanOb{L}(ob(initial(dom(L))))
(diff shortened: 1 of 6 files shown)


epatters:
OK, the syntax

```julia
seq = @relation (a,c) where (a,b,c) begin
  g(a,b)
  h(b,c)
end
k = seq([g, h])
```

is not going to work on Julia 1.0. What should call this operation then? `oapply` or `omap`?

jpfairbanks:
I like `oapply` or `oeval` because it is like evaluating a formula instead of mapping over a data structure (of course formulas are stored as structures).

PR #317, opened 2020-11-02
Pullback functorial data migration for C-sets

slibkind:
I implemented functorial data migration between CSets which the functor represented by an object dictionary and a hom dictionary.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -268,9 +268,10 @@ Migrates an instance of one ACSet
 given dictionaries representing a functor between theories
 """
 
-function migrate!(Y::ACSet{CD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD}
+function migrate!(Y::ACSet{CD, AD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD, AD}
   @assert CD.ob ⊆ keys(FOb)
   @assert CD.hom ⊆ keys(FHom)
+  @assert AD.attr ⊆ keys(FHom)
 
   for (obY, obX) in FOb  
     add_parts!(Y, obY, nparts(X, obX))
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -273,12 +273,12 @@ function migrate!(Y::ACSet{CD, AD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD, 
   CD.hom ⊆ keys(FHom)   || error("Every hom in $CD must be a key in $FHom")
   AD.attr ⊆ keys(FHom)  || error("Every attribute in $AD must be a key in $FHom")
   
-  for (obY, obX) in FOb  
-    add_parts!(Y, obY, nparts(X, obX))
+  for obY in CD.ob
+    add_parts!(Y, obY, nparts(X, FOb[obY]))
   end
 
-  for (homY, homX) in FHom
-    set_subpart!(Y, homY, subpart(X, homX))
+  for homY in CD.hom ∪ AD.attr
+    set_subpart!(Y, homY, subpart(X, FHom[homY]))
   end
   return Y
 end
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -260,4 +260,31 @@ cocone_objects(diagram) = ob(diagram)
 cocone_objects(span::Multispan) = feet(span)
 cocone_objects(para::ParallelMorphisms) = SVector(codom(para))
 
+# Functorial Data migration
+###########################
+
+""" 
+Migrates an instance of one ACSet 
+given dictionaries representing a functor between theories
+"""
+
+function migrate!(Y::ACSet{CD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD}
+  @assert CD.ob ⊆ keys(FOb)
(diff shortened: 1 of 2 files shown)


epatters:
Thanks @slibkind, this will be a great feature to have.

Looks good to me. The only issue I see is that this code won't handle data attributes. Let me know whether you'd like to do that extension or whether I should.

slibkind:
Thanks, Evan! I'll add that extension this evening

slibkind:
To some extent we can treat data attributes like homs. For example, this code has the correct output:

```julia
@present TheoryLabeledGraph <: TheoryGraph begin
  ELabel::Data
  VLabel::Data

  vlabel::Attr(V,VLabel)
  elabel::Attr(E,ELabel)
end
const LabeledGraph = ACSetType(TheoryLabeledGraph, index=[:src, :tgt, :vlabel, :elabel])

@present TheoryLabeledDDS <: TheoryDDS begin
  Label::Data
  label::Attr(X, Label)
end
const LabeledDDS = ACSetType(TheoryLabeledDDS, index=[:Φ, :label])

d = LabeledDDS{Int}()
add_parts!(d, :X, 4, Φ=[2,3,4,1], label=[1, 2, 3, 4])

LabeledGraph{Number, Int}(d, Dict(:V => :X, :E => :X),
                Dict(:src => id(X), :tgt => :Φ, :vlabel => :label, :elabel => :label))
```

A nice extension would be to infer the data attribute types from a dictionary of attributes, but I'm stumped about how to proceed.

src/categorical_algebra/CSets.jl
```diff
@@ -260,4 +260,31 @@ cocone_objects(diagram) = ob(diagram)
...
 
+# Functorial Data migration
+###########################
+
+""" 
+Migrates an instance of one ACSet 
+given dictionaries representing a functor between theories
+"""
+
+function migrate!(Y::ACSet{CD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD}
```
jpfairbanks, line 271: Instead of using a pair of dictionaries, we should introduce a struct

```julia
struct FreeFunctor{C,D,T<:Dict,U<:Dict}<: AbstractFunctor
FOb::Dict
FHom::Dict
end

dom(f::FreeFunctor{C,D, T,U}) = C
codom(f::FreeFunctor{C,D, T,U}) = D
```
with appropriate constructors. Then we can have `FreeFunctor` as an `@instance` of Category so that you could build them up with pushouts and such.

epatters: I'd prefer to keep this PR simple and defer this to another day. #220 already did a lot along these lines and I'd rather try to deal with all that in a more coordinated way.

src/categorical_algebra/CSets.jl
```diff
@@ -268,9 +268,10 @@ Migrates an instance of one ACSet
 given dictionaries representing a functor between theories
 """
 
-function migrate!(Y::ACSet{CD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD}
+function migrate!(Y::ACSet{CD, AD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD, AD}
   @assert CD.ob ⊆ keys(FOb)
   @assert CD.hom ⊆ keys(FHom)
+  @assert AD.attr ⊆ keys(FHom)
```
jpfairbanks, line 274: A julia idiom for these kinds of input validation checks is to use `condition || error("Error message about $var")`. The condition is always tested and since `||` is short circuiting, the error function is only called if the condition is false. I tend to use this `condition || error` style for validating inputs like this and `@assert` for program invariants which should only be false if there is a bug in the code.

slibkind: Thanks! I'll update that.

src/categorical_algebra/CSets.jl
```diff
@@ -260,4 +260,32 @@ cocone_objects(diagram) = ob(diagram)
...
 
+# Functorial Data migration
+###########################
+
+""" 
+Migrates an instance of one ACSet 
+given dictionaries representing a functor between theories
+"""
+
+function migrate!(Y::ACSet{CD, AD}, X::ACSet, FOb::Dict, FHom::Dict) where {CD, AD}
```
olynch, line 271: I would like to see this be more aggressively typed, but I definitely think we should merge this now, and then when we later have a better Functor type, we can use that.

test/categorical_algebra/CSets.jl
```diff
@@ -243,4 +243,44 @@ colim = pushout(α, β)
 @test !is_natural(α′) # Vertex labels don't match.
 @test_throws ErrorException pushout(α′, β)
 
+# Functorial Data Migration
+@present TheoryDDS(FreeSchema) begin
```
olynch, line 247: Do we have this theory in theories? Seems like something that we should have, maybe.

olynch (approved):

src/categorical_algebra/CSets.jl
epatters: Agreed, there was some discussion of this above.

test/categorical_algebra/CSets.jl
epatters: You mean exported from Catlab somewhere? Right now it's only in the unit tests in various places but DDS is a classic example, so probably makes sense to make it official.

jpfairbanks (approved):

PR #321, opened 2020-11-05
Fix braidings in symmetric monoidal double categories

epatters:
The treatment of the braidings in symmetric monoidal double categories is *not* symmetric with respect to the vertical and horizontal directions (see reference cited in docstring). This PR tries to fix it. I think the term constructors are now right. We're still missing some axioms across the double category GATS, but that's less important.


mehalter:
This looks much better! LGTM

mehalter (approved):

Issue #324, opened 2020-11-10
Slow subpart Evaluation


bosonbaas:
It seems that the `Val` function contributes an appreciable amount to the overhead of calling functions like `subpart` and `incident`. Below is example code that showcases the speed difference between calling `subpart` and `_subpart` (running `_subpart` skips the call to `Val` from `subpart`).

```julia
using Catlab.Programs.RelationalPrograms
using Catlab.CategoricalAlgebra
using BenchmarkTools

rel = @relation (a, b) begin
    f(a, b)
end

subpart(rel, :junction)
b1 = @benchmark subpart(rel, :junction)

val_junction = Val(:junction)
CategoricalAlgebra.CSetDataStructures._subpart(rel, val_junction)
b2 = @benchmark CategoricalAlgebra.CSetDataStructures._subpart(rel, val_junction)
```
When ran on my machine, the median time of `subpart` is `4.299 μs`, whereas the median time of `_subpart` is `18.082 ns`

epatters:
Thanks for this report Andrew. That slowdown is quite heavy and also seems totally unnecessary, so hopefully we can figure out how to work around it.

epatters:
By pure chance, I was browsing the Julia Slack today and saw a [thread](https://julialang.slack.com/archives/C67TK21LJ/p1604864140293000) about precisely this issue: the performance implications of dispatching using `Val`. This thread will eventually disappear but you can read it for now.

One important suggestion is to use `Val{x}` and dispatch on `::Type{Val{x}}` instead of using `Val(x)` and dispatching on `::Val{x}`. (I don't think this is documented in the Julia manual.) A quick experiment suggests that making this small change already gives a big improvement.

I'll follow up with a PR tomorrow.

epatters:
@bosonbaas and @jpfairbanks, since C-sets are important to our workflow and we want them to be performant, should we set up some permanent benchmarks to track performance over time? Is there standard tooling for this? I figure there should be, since the Julia community is generally so focused on performance.

epatters:
@bosonbaas, with #327, where do we stand on this? Is it fast enough now?

bosonbaas:
Yep, it's no longer showing up on the profiler!

[...]

PR #327, opened 2020-11-11
Speed up dispatch on value types

epatters:
As described in #324, dispatching on `::Val{x}` is slower than dispatching on `::Type{Val{x}}` because, I assume, `Val(x)` causes an allocation while `Val{x}` doesn't. On my machine, this change gives a 20x speedup on the benchmark in #324.

@bosonbaas, can you compare on the original benchmark that prompted this issue?


bosonbaas (approved):
This has significantly improved performance, thanks!

PR #333, opened 2020-11-18
Convenient Macro for Creating ACSets

olynch:
Right now this is very basic; I plan to add data validation and possibly new syntaxes later on. Currently,
``` julia
g = @acset Graph begin
  V = 2
  E = 2
  src = [1,2]
  tgt = [2,1]
end
```
works, but this will not successfully reject most ill-formed ACSets.

### DIFF
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -5,10 +5,12 @@ export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
   AbstractACSetType, ACSetType, ACSetTableType, AbstractCSetType, CSetType,
   tables, parts, nparts, has_part, subpart, has_subpart, incident,
   add_part!, add_parts!, set_subpart!, set_subparts!, rem_part!, rem_parts!,
-  copy_parts!, copy_parts_only!, disjoint_union
+  copy_parts!, copy_parts_only!, disjoint_union, @acset, init_acset
 
 using Compat: isnothing, only
 
+using MLStyle: @match
+using ...Meta
 using PrettyTables: pretty_table
 using StructArrays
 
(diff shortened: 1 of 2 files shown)


epatters:
Thanks, it will be very convenient to have a constructor like this, especially for interactive work and unit tests. I get tired of writing `add_parts!` over and over again.

jpfairbanks:
@epatters, does this need anything else before we can merge? Let's not hold up this initial feature for the additional features like robust input validation and the CSV style syntax

epatters:
I agree, but the docstrings need to be cleaned up and most importantly, we need to have unit tests.

olynch:
OK, I had some grand plans also involving some @generated code, but I guess I'll just polish this a little and ship it, and do that later.

jpfairbanks:
Yeah, I'm doing stuff with port graphs now and this would be super handy for interactive work. I'm currently doing a lot on paper and then typing it in.

olynch:
OK, I'd feel comfortable merging this now. Sorry for the unnecessary delay.

jpfairbanks (approved):
LGTM! I like that because this is a macro, it should create compile time constants if you use it like:

```julia
triangle() = @acset Graph begin
  V = 3
  E = 3
  src = [1,2,3]
  tgt = [2,3,1]
end

hexagon() = [... some construction of pushouts of triangles ...]
```

that constant should propagate through the code base.

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -5,10 +5,12 @@ export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
   AbstractACSetType, ACSetType, ACSetTableType, AbstractCSetType, CSetType,
   tables, parts, nparts, has_part, subpart, has_subpart, incident,
   add_part!, add_parts!, set_subpart!, set_subparts!, rem_part!, rem_parts!,
-  copy_parts!, copy_parts_only!, disjoint_union
+  copy_parts!, copy_parts_only!, disjoint_union, @acset, init_acset
 
 using Compat: isnothing, only
 
+using MLStyle: @match
```
epatters, line 12: This import looks to be unused.

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -800,4 +802,47 @@ function deletesorted!(a::AbstractVector, x)
...
+  Expr(:call, esc(:eval), expr)
+end
+
+"""
+TODO: Well-formedness check
+TODO: Could also rely on a @generated function that took in a "flat" named tuple
+TODO: Alternative syntax for @acset input based on CSV
+TODO: Actual CSV input
+"""
+function init_acset(T::Type{<:ACSet{CD,AD,Ts}},body) where {CD <: CatDesc, AD <: AttrDesc{CD}, Ts <: Tuple}
```
epatters, line 827: This function is publicly exported, but should it be? If it is exported, it should have a real docstring.

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -800,4 +802,47 @@ function deletesorted!(a::AbstractVector, x)
...
+  body = strip_lines(body)
+  @assert body.head == :block
+  code = quote
+    acs = $(T)()
+  end
+  for elem in body.args
+    @assert elem.head == :(=)
```
epatters, line 834: I do like to use `@match` for stuff like this:

```julia
lhs, rhs = @match elem begin
  Expr(:(=), lhs, rhs) => (lhs, rhs)
  _ => error("Every line of `@acset` must be an assignment")
end
```

In this case it's not any shorter but IMO it's easier to read.

epatters (commented):
Thanks Owen! Looks good, just a few small comments.

olynch:
Yup, those changes all make sense!

epatters:
Thanks, LGTM! Merging now.

PR #337, opened 2020-11-22
Operad of Circular PortGraphs

jpfairbanks:
Circular Port Graphs are halfway between DWDs and UWDs. They support operadic composition with `ocompose`. Half-edge graphs and PDE stencils should be algebras for this operad. It would be great to support bundling of the connections like in UWDs for structured multicospans, but I don't see an obvious implementation of that yet. 

Follow-on issues: 

1. Bundling wires
1. Graphics, these should be very easy to draw with graphviz port graphs
1. Algebras, half edge graphs, stencils/meshes, AlgebraicPetri spatial models


src/wiring_diagrams/CPortGraphs.jl
```diff
@@ -0,0 +1,194 @@
+module CPortGraphs
+using ...Theories
+using ...CategoricalAlgebra
+import ...CategoricalAlgebra.CSets: migrate!
+using ...Present
+using ...Graphs
+using ...Graphs.BasicGraphs
```
epatters, line 7: Encompassed by `using ...Graphs`.

src/wiring_diagrams/CPortGraphs.jl
```diff
@@ -0,0 +1,194 @@
...
+    compose(inv,tgt) == src
+end
+
+@present ThOpenSymCPortGraph <: ThSymCPortGraph begin
+    OP::Ob
+
+    con::Hom(OP, P)
+end
+
+const CPortGraph = ACSetType(ThCPortGraph)
```
epatters, line 44: The pattern I usually follow for C-set types is:

```julia
const AbstractX = AbstractACSetType(TheoryX)
const X = ACSetType(TheoryX, index=[...])
```

Importantly, the abstract type will match any ACSet type for that theory, irrespective of indexing, whereas the concrete type only matches the type with *exactly* the given indices. Thus, the methods bellow should be for the abstract types, not the concrete ones.

For graphs, I usually follow LightGraphs in putting `index=[:src, :tgt]` for graphs and `index=[:src]` for symmetric graphs. This seems like a good default. Advanced users who want to disable indexing for specialized performance reasons can do so by making their own types.

src/wiring_diagrams/CPortGraphs.jl
```diff
@@ -0,0 +1,194 @@
...
+const OSCPGraph = OpenSymCPortGraph
+
+function OpenCPortGraph(g::CPortGraph)
+    x = OpenCPortGraph()
+    copy_parts!(x, g, W=parts(g, :W), P=parts(g, :P), B=parts(g, :B))
+    return x
+end
+
+function migrate!(g::Graph, cpg::CPortGraph) 
+    migrate!(g, cpg,
```
epatters, line 57: I should have raised this in response to #317, but looking at the code for `migrate!` I see that it assumes that `g` is empty. IMO, that's confusing but it's easy enough to fix.

No need to worry about this here. I will fix this in a followup PR.

src/wiring_diagrams/CPortGraphs.jl
```diff
@@ -0,0 +1,194 @@
...
+        t = g[e,:tgt]
+        sbox = g[s, :box]
+        tbox = g[t, :box]
+
+        localport_src = findfirst(s .== incident(g, sbox, :box))
+        localport_tgt = findfirst(t .== incident(g, tbox, :box))
+        @assert length(findall(s .== incident(g, sbox, :box))) == 1
+        @assert length(findall(t .== incident(g, tbox, :box))) == 1
+
+        # println("$s:$sbox→$t:$tbox")
```
epatters, line 167: We can remove these commented `print` statements.

epatters (commented):
Thanks, this looks good! I left a few comments.

This PR needs some documentation, e.g. docstrings for the types, explaining what it's about. For example, you could say that port graphs are intermediate between graphs and wiring diagrams.

I'm not sure how best to handle it, but as we accumulate more methods for `oapply` we'll want to document what they do. I guess you can just keep adding docstrings for the same generic function and they'll all show up. Is there a better way? The important point is that it's not obvious what the operad algebras here do, so it would be good to somehow record that.

src/wiring_diagrams/CPortGraphs.jl
epatters: Extension implemented in #343.

epatters:
Thanks for this feature! Will merge now. In the future, please remember that all code in Catlab has indent width 2 (I just noticed the 4 space width and fixed it).

epatters:
Sorry, I just had another thought. What do you think about replacing `B`, `P`, and `W` with `Box`, `Port`, and `Wire`? This will be more consistent the names used for wiring diagrams.

jpfairbanks:
Yeah I'm totally fine with it. I was figuring out the schemas and Operad compose rule on paper where single letter names work fine but yeah the code base uses full words for this stuff so we should be consistent.

Issue #338, opened 2020-11-25
Attributed Sets


olynch:
For a couple reasons, I think it would be a good idea to have a dedicated type for Attributed Sets that are "backed by" an Attributed C-set. In practice, this would look like

```julia
struct AttributedSet{A <: ACSet, P}
  backing :: A
  parts :: Vector{Int}
end
```

So for instance, if we have

```julia
g = @acset Graph begin
   E = 4
   V = 4
   src = [1,2,3,4]
   tgt = [2,3,4,1]
end
```

Then `AttributedSet{Graph,:E}(g,[2,4])` would represent the data frame

| src | tgt |
|---|---|
| 2 | 3 |
| 4 | 1 |

This is useful for several reasons.

1. It's more straightforward to implement limits in this setting, and then general limits can be reduced to this case.
2. Lots of data science operations only make sense on attributed sets, not attributed c-sets. (see: #335)
3. This is a good thing for a query to return, when we eventually figure out querying.

More generally, it might be interesting to have "views" of ACSets in any shape, just like we can have views of arrays without copying. But I think we should start with AttributedSets.

I'm going to get started on this, but, as always, the earlier I get input on this the easier it is to change things!

epatters:
Good idea. I would definitely support something like this.

Ideally, we can write generic code that supports any `AbstractArray{NamedTuple{...},1}` conforming to the Tables.jl interface, such as the tables in TypedTables:

```julia
struct Table{T <: NamedTuple, N, Data <: NamedTuple{<:Any, <:Tuple{Vararg{AbstractArray{<:Any,N}}}}} <: AbstractArray{T, N}
    data::Data
end
```

Views of the ACSets fit this pattern, as well as things liked TypedTables and StructArrays.

One place where "attributed sets" are already used is in the feet of structured cospans when the boundary object has data attributes. See [`ACSetTableType`](https://github.com/AlgebraicJulia/Catlab.jl/blob/7c02e283a36569d89585215021322025e4e36dc8/src/categorical_algebra/CSetDataStructures.jl#L113) and the `StructuredCospans` module. So whatever framework we come up with for attributed sets should accommodate this use case as well.

olynch:
[...]

PR #340, opened 2020-11-27
ACSetView data structure

olynch:
This addresses #338; see the discussion there for details.

### DIFF
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -5,7 +5,8 @@ export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
   AbstractACSetType, ACSetType, ACSetTableType, AbstractCSetType, CSetType,
   tables, parts, nparts, has_part, subpart, has_subpart, incident,
   add_part!, add_parts!, set_subpart!, set_subparts!, rem_part!, rem_parts!,
-  copy_parts!, copy_parts_only!, disjoint_union, @acset
+  copy_parts!, copy_parts_only!, disjoint_union,
+  @acset, ACSetView
 
 using Compat: isnothing, only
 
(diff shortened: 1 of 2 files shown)


src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -857,4 +858,37 @@ function init_acset(T::Type{<:ACSet{CD,AD,Ts}},body) where {CD <: CatDesc, AD <:
...
+""" Dataframes backed by ACSets
+
+Useful as a "view" onto an ACSet, and also for computing limits
+"""
+
+""" P is a symbol, the object in the schema that this ASet focuses onto
+"""
+struct ACSetView{A <: ACSet, P, Attrs <: NamedTuple} <: AbstractArray{Attrs,1}
+  backing :: A
+  parts :: Vector{Int}
```
epatters, line 870: I figure that we should allow `AbstractVector{Int}` so that we can use ranges (`1:n`) without allocating the corresponding array.

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -857,4 +858,37 @@ function init_acset(T::Type{<:ACSet{CD,AD,Ts}},body) where {CD <: CatDesc, AD <:
...
+
+@generated function Base.getindex(av::ACSetView{<:ACSet,P,Attrs}, i::Int) where
+    {P, Attrs <: NamedTuple}
+  tuple_args = map(Attrs.parameters[1]) do attr
+    :($attr = subpart(getfield(av,:backing), getfield(av,:parts)[i], $(Expr(:quote, attr))))
+  end
+  Expr(:tuple,tuple_args...)
+end
+
+Base.getindex(av::ACSetView,i::Int,name::Symbol) = subpart(getfield(av,:backing), getfield(av,:parts)[i], name)
```
epatters, line 890: Should we restrict to integer indexing? Note that ACSets themselves allow vectorized indexing, which returns a view via the `view` function.

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -857,4 +858,37 @@ function init_acset(T::Type{<:ACSet{CD,AD,Ts}},body) where {CD <: CatDesc, AD <:
   code
 end
 
+""" Dataframes backed by ACSets
+
+Useful as a "view" onto an ACSet, and also for computing limits
+"""
+
+""" P is a symbol, the object in the schema that this ASet focuses onto
```
epatters, line 866: I see two docstrings here, I assume it's supposed to be one docstring?

epatters (commented):
Thanks Owen, I left a few comments.

Also codecov left two warnings, it would be nice to address those with simple tests.

src/categorical_algebra/CSetDataStructures.jl
olynch: Ah, this is definitely a good idea. I was updating this to be a union of `Vector{Int}` and `Colon`, but it's much better to use AbstractVector; that simplifies things a lot.

src/categorical_algebra/CSetDataStructures.jl
olynch: Definitely we should also allow that here.

olynch: Definitely we should also allow that here.

src/categorical_algebra/CSetDataStructures.jl
olynch: I was kind of thinking of the first docstring as the "section header", and the second docstring as the comment for the struct itself.

Also, do you have thoughts on the name `ACSetView`, and also whether this should be in a new file?

olynch: I was thinking of this as a section header, and then a struct docstring.

On a related note, do you think that this should be a in a new file, and not in CSetDataStructures?

epatters: Ah, I see. I am the kind of guy who thinks modules should have sections.

For just the views, I'd say keep it here since it's not too much code, but now that you're adding querying functionality in #345 maybe a new file would make sense?

PR #343, opened 2020-11-30
Extend `migrate!` to attributed C-sets that are not empty

epatters:
As follow-up to #317, the `migrate!` function can now be used like `copy_parts!`, although the latter is `@generated` and the former is not, among other differences.


slibkind (approved):
Looks good to me!

PR #344, opened 2020-11-30
Port names in `@relation` macro

epatters:
This PR extends the `@relation` macro with optional support for named ports. See the docstring and the unit tests for more info.

@bosonbaas, I hope this will be helpful for querying SQL databases. My own intended use case is conjunctive queries of ACSets (not yet implemented!).


bosonbaas:
Thanks, I'll take a look at this PR!

PR #345, opened 2020-11-30
Querying ACSets

olynch:
I accidentally deleted the branch for the previous PR, #340, so this continues that PR. Also, this includes queries. Given that the ACSetView structure still needs review, I figure that it makes sense to have both the ACSetView structure and the queries on the same PR, because changes to one will probably necessitate changes in the other.

### DIFF
--- a/src/categorical_algebra/ACSetViews.jl
+++ b/src/categorical_algebra/ACSetViews.jl
@@ -1,0 +1,187 @@
+""" Data structure for views of attributed C-sets
+"""
+module ACSetViews
+export ACSetView, backing, @compute_prop, @select_where,
+  ACSetViewMorphism
+
+using ...CSetDataStructures
+using ..CSets
+using ..FinSets
+using ...Theories: CatDesc, AttrDesc, dom, codom_num
+
+using MLStyle: @match
+using PrettyTables: pretty_table
+using Tables
+
+# Data types
+############
+
+""" One useful way of thinking of ACSetViews is that they are
+dataframes backed by ACSets.
+
+They are also useful for computing limits
+
+P is a symbol, the object in the schema that the view focuses onto
+"""
+struct ACSetView{A <: ACSet, P, Attrs <: NamedTuple} <: AbstractArray{Attrs,1}
+  backing :: A
+  parts :: AbstractVector{Int}
+  function ACSetView(backing::A, P::Symbol, parts::AbstractVector{Int}) where
+      {CD <: CatDesc,AD <: AttrDesc{CD},Ts <: Tuple, A <: ACSet{CD,AD,Ts}}
+    attr_names = filter(a -> dom(AD,a) == P, AD.attr)
+    attr_types = map(a -> Ts.parameters[codom_num(AD,a)], attr_names)
+    row_type = NamedTuple{attr_names, Tuple{attr_types...}}
+    new{A,P,row_type}(backing,parts)
+  end
+  function ACSetView(backing::A, P::Symbol) where
+      {CD <: CatDesc,AD <: AttrDesc{CD},Ts <: Tuple, A <: ACSet{CD,AD,Ts}}
+    ACSetView(backing,P,1:nparts(backing,P))
+  end
+end
+
+# We're going to overload getproperty, so we use these instead
+
+backing(av::ACSetView) = getfield(av,:backing)
+
+CSetDataStructures.parts(av::ACSetView) = getfield(av,:parts)
+
+
+# Overloaded accessors
+######################
+
+Base.size(av::ACSetView) = size(parts(av))
+
+@generated function Base.getindex(av::ACSetView{<:ACSet,P,Attrs}, i::Int) where
+    {P, Attrs <: NamedTuple}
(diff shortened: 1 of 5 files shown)


epatters:
Writing this down here before I forget: `ACSetView` should implement the [Tables.jl](https://github.com/JuliaData/Tables.jl) interface. This won't add any additional dependency since Catlab already indirectly depends on Tables.jl through StructArrays.jl.

olynch:
OK! Will do.

epatters:
Great, thanks! And as of #350, Catlab will depend explicitly on Tables.jl.

olynch:
This is ready for review @epatters, though I already know I need to add a test for ACSetViewMorphism.

src/categorical_algebra/ACSetViews.jl
```diff
@@ -0,0 +1,185 @@
...
+  print(io, "ACSetView of object $P")
+  println(io)
+  table = tables(backing(av))[P][parts(av)]
+  if !(eltype(table) <: NamedTuple{(),Tuple{}} || isempty(table))
+    pretty_table(io, table, nosubheader=true,
+                 show_row_number=true, row_number_column_title=string(P))
+  end
+end
+
+# Tables.jl interface
```
epatters, line 90: It looks like most of the Tables.jl interface is not covered by tests. Would be good to add a simple test for each function.

src/categorical_algebra/ACSetViews.jl
```diff
@@ -0,0 +1,185 @@
...
+# Macros for Querying/filtering
+###############################
+
+""" This macro takes an expression using attributes of the object selected by the ACSetView,
+and returns an array of the result of that expression applied to the attribute of each part
+in the ACSetView.
+
+Example: @compute_prop(edges_table, 2 * src.weight)
+
+Eventually, this should be superceded by Query.jl or AlgebraicRelations.jl functionality, so I have
```
epatters, line 117: Lol at "purposely made the name a bit obtuse and awkward"!

The ability to use compound attributes is nice. IMO, if you think this is useful (it seems useful to me!), you should give it a worthy name.

src/categorical_algebra/ACSetViews.jl
```diff
@@ -0,0 +1,185 @@
...
+    Expr(:., x, y) => [flatten_dot_expr(x)..., y.value]
+    _ => error("unsupported syntax in @query")
+  end
+end
+
+function flatten_dot_expr(ex::Symbol)
+  [ex]
+end
+
+""" This macro takes in a similar expression to @compute prop, except that the expression
```
epatters, line 150: `@compute prop` -> ```[`@compute_prop`](@ref)```

epatters (commented):
Thanks Owen, this looks really nice! Minor comments above. The main thing is to finish the tests, then this should be ready to merge.

src/categorical_algebra/ACSetViews.jl
olynch: I think this function is useful, but I think that it should be replaced by something more useful. And actually, I think `@compute_prop` is not so bad.

epatters:
Thanks, I'll merge this!

Not sure what is going on with Codecov. It is claiming that lots of things you clearly have tests for are uncovered. Hopefully this problem does not persist.

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -871,4 +873,39 @@ function init_acset(T::Type{<:ACSet{CD,AD,Ts}},body) where {CD <: CatDesc, AD <:
   code
 end
 
+""" Map over a data type, in the style of Haskell functors
+"""
+
+Base.map(B::Type, f::Function, D::Symbol, acs::ACSet) = _map(B, f,Val{D},acs)
```
epatters, line 879: Did this sneak in from #352? It's not supposed to be part of this PR, right?

epatters (commented):
I just realized why the code coverage is so low: the tests aren't running on CI because they aren't included in `test/categorical_algebra/CategoricalAlgebra.jl`! Also, another question below.

olynch:
Oh whoops! I'll fix those two issues, I guess I must have mixed them up when merging or something, and I was running the tests through an ad-hoc test set so I completely forgot to add them!

epatters:
Thanks, once the tests are passing, this should be ready to go.

olynch:
Looks like this is one of the few times where the correct fix is to remove a test!

olynch:
It's failing because it can't import Tables.jl... The tests run on my machine, so I think it might be a CI bug, especially because Tables is in the Project.toml.

epatters:
Ah, it's failing because the tests have their own [Project.toml](https://github.com/AlgebraicJulia/Catlab.jl/blob/master/test/Project.toml), which is needed because the tests use various optional and extra dependencies that Catlab doesn't depend on.

epatters:
Good news: Now that #358 is merged, Tables.jl is included in `tests/Project.toml`, so if you rebase, the tests will hopefully pass on CI.

Bad news: There's now a small merge conflict, which you can also fix by rebasing.

epatters (approved):
LGTM. Thanks Owen!

PR #347, opened 2020-12-04
Functions out of finite sets and other extensions

epatters:
This PR extends and refactors the `FinSets` module to be more flexible. Some of the underlying concrete types have changed, but if you just interact with this module through `FinSet` and `FinFunction` (like you're supposed to) you shouldn't notice any differences.

In summary:

- Created a `Sets` module for the category of sets (not necessarily finite), whose simplest example is `TypeSet{T}`, which just interprets the Julia type `T` as a set
- Moved @olynch's `PredicatedSet{T}` into this module, which serves as another concrete example
- Refactored `FinSets` to use `Sets` whenever possible, e.g. for functions represented by Julia callables
- Added `FinDomFunction` for functions whose *domains* are finite sets

The last point was my original motivation for going down this rabbit hole. Most limits of `FinDomFunction`s can be computed just as easily as limits of `FinFunction`s, yet for database joins we really need the former. This will be the topic of future PRs.

The philosophy here is that there are lots of potentially useful Set-like categories out there and we want to support them all, while ensuring that important subcategories are preserved by composition. E.g., if you compose two `FinFunction`s you get another `FinFunction`, and if the original functions have vector representations, then so does the composite.


olynch:
This looks really nice to me!

epatters:
Thanks! I will merge then.

PR #350, opened 2020-12-06
Support any Tables.jl-compliant backend for tables in C-set

### LINKED ISSUE #330: Consider replacing StructArrays with TypedTables
We should consider whether it's worth replacing [StructArrays.jl](https://github.com/JuliaArrays/StructArrays.jl) with [TypedTables.jl](https://github.com/JuliaData/TypedTables.jl) in the C-sets implementation.

The packages have similar interfaces, compatible with [Tables.jl](https://github.com/JuliaData/Tables.jl), and similar implementations, based on a named tuple of arrays. Thus it should be very straightforward to replace one with the other. On the other hand, my understanding, based on conversations with @jpfairbanks and Lyndon White, is that TypedTable specifically aims to be a fully typed alternative to [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl), while StructArrays has a different philosophy. So, since we think of C-sets as a collection of interconnected data frames, it may make more sense to use TypedTables.

epatters:
This PR:

- allows any column-based table type compliant with [Tables.jl](https://github.com/JuliaData/Tables.jl) to be used for the tables in a C-set
- switches the default table type to [TypedTables.jl](https://github.com/JuliaData/TypedTables.jl), which closes #330

The following concrete table types are known to work:

- StructArrays (the old default)
- TypedTables (the new default)
- named tuples of arrays (validated in the unit tests)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -26,6 +26,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
+Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
 AutoHashEquals = "0.2"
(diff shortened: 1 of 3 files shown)


olynch (approved):
It's pretty shocking how little code change this required! Looks good to me!

epatters:
Thanks! Yeah, the Tables.jl interface seems well thought out. It's nice.

PR #351, opened 2020-12-07
Bipartite graphs

### LINKED ISSUE #300: Bipartite Graphs
We should add Bipartitioned graphs to `Catlab.Graphs`. We can then use this type for both AlgebraicPetri.jl and also be able to represent presentations that would give us pushout support for combining presentations as well as graphically displaying them. I think we something like the following types:

```julia
@present BipartitionedGraph(FreeSchema) begin
  (V1, V2)::Ob
  E::Ob
  src::Hom(E,V1)
  tgt::Hom(E,V2)
end

@present BipartitionedDiGraph(FreeSchema) begin
  (V1, V2)::Ob
  (E1, E2)::Ob
  src1::Hom(E1,V1)
  tgt1::Hom(E1,V2)
  src2::Hom(E2,V1)
  tgt2::Hom(E2,V2)
end
```

epatters:
A bare-bones implementation that provides the schemas, types, and basic accessors and mutators. Resolves #300.

### DIFF
--- a/src/graphs/BasicGraphs.jl
+++ b/src/graphs/BasicGraphs.jl
@@ -100,12 +100,12 @@ add_vertices!(g::AbstractACSet, n::Int; kw...) = add_parts!(g, :V, n; kw...)
 
 """ Add an edge to a graph.
 """
-add_edge!(g::AbstractGraph, src::Int, tgt::Int; kw...) =
+add_edge!(g::AbstractACSet, src::Int, tgt::Int; kw...) =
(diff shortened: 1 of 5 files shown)


epatters:
Bike-shedding questions: how do we feel about the Unicode characters in the schema and Julia API? Or about the naming conventions here for bipartite graphs?

epatters:
Micah approved this one via Zulip, so I'll go ahead and merge.

Issue #335, opened 2020-11-19
Higher Order Functions on ACSets


olynch:
We should implement the higher-order functions that are easy to do so on ACSets.

```julia
function map(f::FunctionWrapper{S,T}, data::Symbol, acs::ACSet{CD,AD,Ts})
```

This would return an ACSet where the `Data` type corresponding to `data` in Ts (which should be `S`) is replaced by `T`. I wish there was a way to express this in types, but I'm not sure if it's possible... Best approximation would be

```julia
function map1(f:FunctionWrapper{S,T}, acs::ACSet{CD,AD,Tuple{T}})::ACSet{CD,AD,Tuple{S}}
```

We should also have `filter`, which takes in a function of a "part", and all the parts that return true are copied to a new array, along with some other parts? The categorical semantics of this need to be worked out a bit.

And then `filterMap`, of course

I don't know if any type of `fold` makes sense, but if so, we should have it. At some point, I'm just going to go through a checklist of all the functions in dplyr, and figure out which ones make sense for ACSets.

And then also maybe some sorting functions?

PR #352, opened 2020-12-07
Haskell-style map function for ACSets

olynch:
Pretty simple, basically:

``` julia
map(Int,s -> Int(s[1]),:Weight,g::WeightedGraph{String})::WeightedGraph{Int}
```

All attributes with a given codomain get converted using the supplied function. Partly addresses #335.

### DIFF
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -5,11 +5,10 @@ export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
   AbstractACSetType, ACSetType, ACSetTableType, AbstractCSetType, CSetType,
   tables, parts, nparts, has_part, subpart, has_subpart, incident,
   add_part!, add_parts!, set_subpart!, set_subparts!, rem_part!, rem_parts!,
-  copy_parts!, copy_parts_only!, disjoint_union, @acset, wrangle_attributes
+  copy_parts!, copy_parts_only!, disjoint_union, @acset
 
 using Compat: isnothing, only
 
-using FunctionWrappers: FunctionWrapper
 using MLStyle: @match
 using ...Meta
 using PrettyTables: pretty_table
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -5,7 +5,7 @@ export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
   AbstractACSetType, ACSetType, ACSetTableType, AbstractCSetType, CSetType,
   tables, parts, nparts, has_part, subpart, has_subpart, incident,
   add_part!, add_parts!, set_subpart!, set_subparts!, rem_part!, rem_parts!,
-  copy_parts!, copy_parts_only!, disjoint_union, @acset
+  copy_parts!, copy_parts_only!, disjoint_union, @acset, wrangle_attributes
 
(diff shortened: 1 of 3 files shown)


olynch:
@epatters btw this is ready to be reviewed.

epatters:
Cool, thanks. I'll review it soon.

epatters:
Thanks Owen. Two questions/comments:

1. Do you think there will be many use cases for this as is? My worry is that if you want to transform multiple attributes, you'd have to call `map` multiple times and end up reallocating the whole C-set each time. Also, maybe sometimes you want to map all data attributes with given codomain, as here, or just a particular data attribute. I think you could generalize this to something like [`mutate`](https://dplyr.tidyverse.org/reference/mutate.html) in the dplyr without too much change to the structure of your code but a big boost in usefulness.

2. In idiomatic Julia you shouldn't usually have to specify types that can be inferred. So instead of supplying `Int` as a first argument, you compute `map(s -> Int(s[1]), ...)` in base Julia, which returns a `Vector{Int}` whose element type you can extract. I guess this would complicate the code a bit.

Putting these suggestions together, the API would imagine is something like

```julia
map(g::WeightedGraph{Symbol}; weight = s -> Int(s[1]))
```

or

```julia
map(g::RelationDiagram{Symbol}; name = titlecase∘string, variable = uppercase∘string)
```

What do you think?

olynch:
OK, so I guess then there is an error if you don't correctly map all of the attributes to a new type?

Yeah, can do!

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -5,17 +5,18 @@ export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
   AbstractACSetType, ACSetType, ACSetTableType, AbstractCSetType, CSetType,
   tables, parts, nparts, has_part, subpart, has_subpart, incident,
   add_part!, add_parts!, set_subpart!, set_subparts!, rem_part!, rem_parts!,
-  copy_parts!, copy_parts_only!, disjoint_union, @acset
+  copy_parts!, copy_parts_only!, disjoint_union, @acset, wrangle_attributes
```
epatters, line 8: I guess this is a spurious export since I don't see `wrangle_attributes` below.

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -5,17 +5,18 @@ export AbstractACSet, ACSet, AbstractCSet, CSet, Schema, FreeSchema,
...
 
 using Compat: isnothing, only
 
+using FunctionWrappers: FunctionWrapper
```
epatters, line 12: Unused import?

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -857,4 +858,75 @@ function init_acset(T::Type{<:ACSet{CD,AD,Ts}},body) where {CD <: CatDesc, AD <:
...
 
+""" Map over a data type, in the style of Haskell functors
+"""
+
+function Base.map(acs::ACSet; kwargs...)
+  fns = (;kwargs...)
+  _map(acs, fns)
+end
+
+function sortunique!(x)
```
epatters, line 869: It'd be nice if this were built into Julia. I call `unique!(sort!(x))` all over the codebase.

epatters (commented):
Thanks Owen, I think this method will be very useful.

epatters:
Awesome, merging now.

PR #365, opened 2020-12-26
Use StaticArrays 1.0

MasonProtter:
It'd be nice to not have Catlab downgrade my StaticArrays


epatters:
Thanks!

I think the error coming from BenchmarksCI is temporary and will be resolved by merging, so I'm going to merge and hope for the best.

PR #367, opened 2021-01-01
Create CompatHelper.yml

ChrisRackauckas:


epatters:
Thanks for the push to do this, I should have set this up a long time ago.

Since I haven't used CompatHelper before, I'm going to copy the official config file, which is slightly different than this one.

PR #370, opened 2021-01-06
ENH: KeyedTables Tables.jl-compatible table

olynch:
As part of the work for #354, this is a Tables.jl-compliant wrapper for any Tables.jl-compliant data type that adds indexing by an arbitrary type. This is useful for when we want to make working with an acset more human-friendly without changing the schema to add decorating morphisms.


epatters:
Thanks Owen and sorry for the delayed review.

Code looks nice. I was wondering where you've used this and whether you intend to use it in Catlab itself. Since this doesn't depend on anything in Catlab, it might be worth releasing it as its own package. It'd be a small package, but there's nothing wrong with that.

olynch:
So, with the new acset implementation, I don't know what to do with this, because I'm now thinking that the way forward is to use the different "FinSet" implementations, which include sets of symbols, for the domains of functions. My plan is to do a big acset hackathon on the plane back from the Netherlands; I'll see what my thoughts are after that.

jpfairbanks:
@olynch is this superseded by struct acsets?

epatters:
@olynch, this is no longer relevant, right? I'm going to suggest that we close this because it doesn't make sense to maintain a general Table type in Catlab unless we have a clear need for it.

PR #371, opened 2021-01-10
Implement DWDs using ACSets, v2

epatters:
Reimplements directed wiring diagrams to be C-sets for suitable schema C. The core imperative interface is unchanged but it's implementation is different.


epatters:
All checks are now passing but I plan to do some more cleanup.

I also need to add more tests: I want the DWD core to have basically 100% coverage, but new branches were introduced by the four cases in ACSet (passing, incoming, outgoing, internal) and not all of them are covered in all methods. Thus, I am not confident that this code works.

Issue #375, opened 2021-01-16
A Radically Different Acset Implementation


olynch:
There are some problems with the current Acset implementation.

1. The generated types are, to put it lightly, a Lovecraftian horror. Looking at error messages for them is *awful*.
2. The way that we put all of the type parameters into the type is a massive hack, and not extensible to, say, allowing for equations in the type, which we might want later down the road.
3. The support for "subtyping" is extremely janky.

Here is my proposal for fixing this. Instead of fixing the parameters to one generic acset type, `ACSetType(TheoryGraph)` should declare a new struct type that subtypes `AbstractACSet` and implements `schema(::Type{Graph})`, which returns the `FreeSchema` associated to `Graph`. Obviously, it then needs to be a macro.

Then all of our `@generated` methods can apply to any struct that subtypes `AbstractACSet` and implements `schema(::Type{Graph})`, as long as we lay out the data in each generated struct in some logical way that corresponds to the schema.

In fact, the current acset implementation could be subsumed by this, by writing a `schema(::Type{ACSet{CD,AD}})` method.

This also decouples the `@generated` methods from the data storage; in theory we could write a *bespoke* implementation for a certain acset that took advantage of some domain-specific features, but then still have all of the `@generated` methods as long as we were careful to lay out the data in a certain way.

This solves 1 because stack traces now use the name of the generated struct, so `Graph` would show up as `Graph`.

This solves 2 because the schema is no longer stored in the type, it is returned by a method on the type.

This solves 3 because we can now have `DecoratedGraph` subtype `AbstractGraph`, if `@ACSetType` looks at the ancestor of the schema, or if we just manually add a parameter to `@ACSetType` telling it which type to subtype when it generates the struct and/or associated abstract type.

This may even have performance benefits, if Julia can specialize code more effectively for custom struct types.

jpfairbanks:
This sounds like it is shifting stuff from generated function compile time to macro time, which allows for more flexibility especially around manipulating names of stuff for the error messages. We definitely need to benchmark compile times before and after this change. @mehalter and @bosonbaas are reporting long precompiled times from AlPetri so we definitely don't want to make things worse. I think the benefits to error reporting and subtyping would be big wins that offset a small increase in compile times.

olynch:
Is it possible, or will it eventually be possible, to do something with `precompile` to help with this? https://julialang.org/blog/2021/01/precompile_tutorial/

epatters:
[...]

PR #376, opened 2021-01-16
Custom Structs for ACSets

olynch:
This is a draft for generating custom structs for ACSets, related to #375. Currently, if pres is the presentation for `WeightedGraph`, then `struct_acset(:WeightedGraph, pres, [:src,:tgt])` produces

``` julia
quote
    #= /g/Catlab.jl/src/categorical_algebra/StructAcsets.jl:42 =#
    struct WeightedGraph{Weight}
        #= /g/Catlab.jl/src/categorical_algebra/StructAcsets.jl:43 =#
        tables::NamedTuple{(:V, :E), Tuple{Table{NamedTuple{(), Tuple{}}}, Table{NamedTuple{(:src, :tgt, :weight), Tuple{Int64, Int64, Weight}}}}}
    end
    #= /g/Catlab.jl/src/categorical_algebra/StructAcsets.jl:46 =#
    schema(::Type{<:WeightedGraph}) = begin
            #= /g/Catlab.jl/src/categorical_algebra/StructAcsets.jl:46 =#
            Presentation{Schema,Symbol}(Catlab.Theories.FreeSchema, (Ob = Catlab.Theories.FreeSchema.Ob{:generator}[V, E], Hom = Catlab.Theories.FreeSchema.Hom{:generator}[src, tgt], Data = Catlab.Theories.FreeSchema.Data{:generator}[Weight], Attr = Catlab.Theories.FreeSchema.Attr{:generator}[weight]), Dict(:src => (:Hom => 1),:weight => (:Attr => 1),:V => (:Ob => 1),:E => (:Ob => 2),:tgt => (:Hom => 2),:Weight => (:Data => 1)), Pair[])
        end
end
```

I still have to add indices and make this into a macro, but as a proof of concept this is totally doable. I'm now having vague recollections of discussing this before; did we decide that something made this undesirable? I feel like it really might solve a lot of our problems.


olynch:
The good news is that I think most of the accessing code doesn't even need to change that much! We even get to clean it up, because we can work with the presentation object directly instead of the clumsy `CatDesc` and `AttrDesc`.

jpfairbanks:
Yes, and functions that are constant given type of the input argument like your schema function, are subject to constant propagation at compile time. So no runtime performance penalty is introduced there.

olynch:
Well, schema is only called a `@generation` time, so it doesn't really matter either way :D

jpfairbanks:
I can see an application where you would want to avoid generated functions and use a more generic implementation to avoid compile costs. Like if you used ACSets to interface with a DB and every query result came back as an ACSet with a different schema. You would want a generic implementation for all schemas rather than specializing on each schema so that the compiler wasn’t invoked for every query. But that application is far away and this refactor will help with that by making subtyping less janky.

olynch:
> I can see an application where you would want to avoid generated functions and use a more generic implementation to avoid compile costs. Like if you used ACSets to interface with a DB and every query result came back as an ACSet with a different schema. You would want a generic implementation for all schemas rather than specializing on each schema so that the compiler wasn’t invoked for every query. But that application is far away and this refactor will help with that by making subtyping less janky.

Yes, this makes a lot of sense, and this is exactly what I mean when I say "custom implementations". I.e.; we can have different implementations of the basic operations, including generic ones.

olynch:
OK, while we're at it, we might as well clean up the interface for declaring an acset. Really, it should be as simple as:

```julia
@acset_type WeightedGraph{T} begin
  V::Ob
  E::Ob with {src::V,weight::T}
  tgt::Hom(E,V)
end
```

I.e., we support both declaring homs in-line and separately, and data objects are automatically declared through the generic type. This also obviates the annoying requirement to have both "WeightedGraph" and "TheoryWeightedGraph". We will just have "WeightedGraph" and "schema(WeightedGraph)".

src/categorical_algebra/StructAcsets.jl
```diff
@@ -0,0 +1,356 @@
...
+  name::Symbol
+  dom::Symbol
+  codom::Symbol
+  idxed::Bool
+end
+
+""" Same information as Presentation{Schema}, but without equations,
+and in a much more concrete/easily accessible format.
+"""
+struct ParsedSchema
```
jpfairbanks, line 68: I guess we can't use an ACSet for this... too bad the CT primordial soup can't be implemented as a soup.

src/theories/ProductSchema.jl
```diff
@@ -0,0 +1,19 @@
...
+  """
+  compose(f::Hom(A,B), g::Attr(B,X))::Attr(A,X) ⊣ (A::Ob, B::Ob, X::Data)
+
+  (compose(f, compose(g, a)) == compose(compose(f, g), a)
+    ⊣ (A::Ob, B::Ob, C::Ob, X::Data, f::Hom(A,B), g::Hom(B,C), a::Attr(C, X)))
+  compose(id(A), a) == a ⊣ (A::Ob, X::Ob, a::Attr(A,X))
+end
+
+@syntax FreeSchema{ObExpr,HomExpr,DataExpr,AttrExpr} ProductSchema begin
+  # should have a normal representation for precompose of a morphism + a generator attribute
```
jpfairbanks, line 17: this normal form should just be a list of homs and the optional attribute name right? So the syntax should associate things into a list and then when you compose two expressions, you make sure that only the last one has a non-null attribute?

jpfairbanks: @epatters, can Catlab support custom structs for the syntax system? That seems like the way to go for making Operads fit in to the whole approach.

jpfairbanks (commented):
I haven't fully processed the parsing code, but I don't see any red flags, @epatters what do you think?

olynch:
@epatters, we need to figure out naming for `CSets.jl`/`Acsets.jl`. Maybe what we currently call `Acsets.jl` should be `AcsetInterface.jl`?

olynch:
One big todo here is to reimplement the tables.jl functionality, now that this doesn't use a tables.jl thing internally.

olynch:
Also need to do `unique_index`

olynch:
And the `@acset` macro, and `map`. Soo much acset infrastructure, this better all be worth it...

epatters (approved):

PR #380, opened 2021-01-29
ENH: Add function to visualize presentation

olynch:
This is just a simple function to draw the generating morphisms in a presentation in graph form.

Questions:
1. Should this go in another module?
2. How does one test something like this, especially given that the output may change as we decide to make the style different?
3. How should data objects look different? Should they be in red, or should they have a square around them?

I will clean up code and add docstrings, but also I would like answers to those questions. Also, I just realized that I need to specialize this to `Presentation{Schema}`.

The generated graph for Graph looks like this:
image


epatters:
Thanks, this will be great to have.

It should probably go in `Catlab.Graphics.GraphivzGraphs` with names like

```
to_graphviz_graph(pres::Presentation) = [your code]
to_graphviz(pres::Presentation) = to_graphviz(to_graphviz_graph(pres))
```

Then users can just call the generic function `to_graphviz` on presentations to get a quick visualization.

olynch:
This is really handy! These diagrams are so much easier to understand than the presentation!

image

Is there graphviz layout wizardry that I can do to make these have a better layout?

olynch:
This looks better, but is it possible to make the edge labels centered on the edges?
image

olynch:
Ah, maybe the way to do this is by taking the layout from Graphviz, and then outputting the actual figure using Compose or TikZ. I'm going to read the code that you wrote for this for wiring diagrams.

olynch:
OK, I think I want to merge this at the current level of functionality, so I'm requesting a review.

src/core/Present.jl
```diff
@@ -117,6 +117,10 @@ function add_definition!(pres::Presentation, name::Symbol, rhs::GATExpr)
   generator
 end
 
+""" Get the index of a generator
+"""
+index_of(pres::Presentation, x::Symbol) = pres.generator_name_index[x].second
```
epatters, line 122: Let's call this `generator_index` or `index_of_generator` since `index_of` sounds very generic.

epatters (commented):
I agree, this will be very handy! I left one inline comment.

Also, this function should have a test of some kind. The minimal effort approach is to write a unit test that checks some basic invariant of the Graphviz graph. Another nice thing (not required) to do for graphics stuff is to create a Literate.jl vignette in the docs. For example, see [this vignette](https://algebraicjulia.github.io/Catlab.jl/stable/generated/graphics/graphviz_wiring_diagrams/) for the Graphviz wiring diagrams. It doubles as a "visual test" that will also break the build if the code errors. It shouldn't be too much work to just copy the examples you've been playing with.

Re: the possibility of other graphics backends, Catlab actually does support parsing graph layouts from the JSON output of Graphviz; see [`parse_graphviz`](https://github.com/AlgebraicJulia/Catlab.jl/blob/5c6ce5a7024e3e9cb771679cf988438e267dfd96/src/graphics/GraphvizGraphs.jl#L22). You could pursue this in a follow-up PR if you get the itch.

jpfairbanks:
I agree that Evan's inline comment and then LGTM

epatters:
To be clear, I am going to insist on some form of testing, however minimal :)

epatters:
Thanks Owen! Merging now.

PR #383, opened 2021-02-12
No type piracy in ACSet schema description types

### LINKED ISSUE #382: getproperty type piracy
Hi,

I would like to draw your attention to a type piracy that Catlab has in its definition of `getproperty`. This is causing a slowdown in surprising ways in Julia. Ref: https://github.com/JuliaLang/julia/issues/39534

epatters:
Resolves #382.

### DIFF
--- a/src/categorical_algebra/ACSetViews.jl
+++ b/src/categorical_algebra/ACSetViews.jl
@@ -7,7 +7,7 @@ export ACSetView, backing, @compute_prop, @select_where,
 using ...CSetDataStructures
 using ..CSets
 using ..FinSets
-using ...Theories: CatDesc, AttrDesc, dom, codom_num
+using ...Theories: CatDesc, AttrDesc, dom, codom_num, attr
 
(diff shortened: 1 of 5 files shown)


olynch:
So, the one thing is, and this is the reason I made these accessors in the first place, that I use `ob`, `hom`, `data`, and `attr` as variable names all over the place. This isn't necessarily a reason why this shouldn't be merged, but it's something to keep in mind, and maybe we should shift to using variable names `C`, `f`, `X`, and `a`.

epatters:
One indication that this isn't too big a deal is that in the whole codebase, I only had to change the variable names in one or two places! You can't use `ob` as a variable name if you also use it as a function, but you *can* use it as a binding in list comprehensions and local function argument names, which is often good enough.

KristofferC:
An alternative is to create (singleton) instances of the types and then use `getproperty` on those (similar to how `nothing` tends to be used over `Nothing`).

PR #402, opened 2021-04-10
add theories of co/Presheaf

cbw124:
Adding GATs for coPresheaf and Presheaf, as left/right modules of a category.
They are both straightforward. There are possibly some small choices to finalize.
Can also extend both to profunctors, which I'm sure will be useful.

### DIFF
--- a/src/theories/Category.jl
+++ b/src/theories/Category.jl
@@ -46,11 +46,11 @@ end
   Psh(dom::Ob)::TYPE
 
   # functoriality = covariant action
-  actr(x::Psh(A), f::Hom(A,B))::Psh(B) ⊣ (A::Ob, B::Ob)
+  act(x::Psh(A), f::Hom(A,B))::Psh(B) ⊣ (A::Ob, B::Ob)
 
   # action equations
-  actr(actr(x , f) , g) == actr(x , (f ⋅ g))  ⊣ (A::Ob, B::Ob, C::Ob, f::(A → B), g::(B → C), x::Psh(A))
-  actr(x , id(A)) == x  ⊣ (A::Ob, x::Psh(A))
+  act(act(x , f) , g) == act(x , (f ⋅ g))  ⊣ (A::Ob, B::Ob, C::Ob, f::(A → B), g::(B → C), x::Psh(A))
+  act(x , id(A)) == x  ⊣ (A::Ob, x::Psh(A))
 end
 
 @theory Presheaf{Ob,Hom,Psh} <: Category{Ob,Hom} begin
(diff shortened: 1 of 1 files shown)


src/theories/Category.jl
```diff
@@ -39,6 +39,34 @@ We use symbol ⋅ (\\cdot) for diagrammatic composition: f⋅g = compose(f,g).
...
+  # unicode - *decide later*
+
+  # presheaf = object-indexed family
+  Psh(codom::Ob)::TYPE
+
+  # functoriality = contravariant action
+  actr(x::Psh(B), f::Hom(A,B))::Psh(A) ⊣ (A::Ob, B::Ob)
+
+  # action equations
+  actr(actr(x , f) , g) == actr(x , (f ⋅ g))  ⊣ (A::Ob, B::Ob, C::Ob, f::(A → B), g::(B → C), x::Psh(C))
```
epatters, line 66: Should this equation be `actr(actr(x , g) , f) == actr(x , (f ⋅ g))`?

src/theories/Category.jl
```diff
@@ -39,6 +39,34 @@ We use symbol ⋅ (\\cdot) for diagrammatic composition: f⋅g = compose(f,g).
   id(A) ⋅ f == f ⊣ (A::Ob, B::Ob, f::(A → B))
 end
 
+@theory coPresheaf{Ob,Hom,Psh} <: Category{Ob,Hom} begin
```
epatters, line 42: Catlab follows the naming convention that theories are capitalized, so this should be `CoPresheaf` or `Copresheaf`. My preference is for the latter, following the [nLab](https://ncatlab.org/nlab/show/copresheaf).

epatters (commented):
Thanks Christian, this is great! A few small questions/comment inline.

src/theories/Category.jl
cbw124: okay, sounds good.

src/theories/Category.jl
cbw124: yes, good catch. thanks.

jpfairbanks:
We should probably include an Instance of this category with the theory. Are CSets and the pullback data migration (action by precomposition) a good model to offer as the instance? The category would be CAT and the actions are functors into FinSet as implemented by `CSetType`s?

jpfairbanks:
I guess this isn't a category of all presheafs it's like the structure of a presheaf over a fixed indexing category.

jpfairbanks:
How do we feel about `Psh`?  `Presheaf` or `Instance` is more explicit without using that many more characters. `Instance` also has the benefit of using the same name for covariant and contravariant versions.

cbw124:
sure, yeah I would either choose either Presheaf or P. (or Module). isn't Instance a key term, wouldn't it be like overloading?

epatters:
How about `Elem` or `Element`? I think this makes sense because in a presheaf `X` presented by generators and relations, a term

```julia
x::Elem(A)
```

represents an element of the set `X(A)`.

jpfairbanks:
Yeah as you point Instance already means a thing more like Model, I like El, Elm, Elem, or Element. We are pretty terse in our GAT definitions using things like Ob, Hom, and Hom2, so my general preference for longer names can be overridden by the custom of short names in GATs

epatters:
I am happy with `El` or `Elem`. I'd prefer not to have `Elm`.

jpfairbanks:
How about the other obvious option, Elt?

epatters:
OK, if I had to vote for one option, it'd be `El` because (1) it is object-like and has two characters like `Ob` and (2) the [nLab](https://ncatlab.org/nlab/show/category+of+elements) uses the notation el(F) for the category of elements of a functor F: C -> Set.

jpfairbanks:
`El` works for me! I like the 2 letters for objecty GAT sorts and 3 letters for homy GAT sorts.

PR #407, opened 2021-04-18
Presentation from SchemaType

slibkind:
Inverse of the function SchemaType.


slibkind:
I wonder if this function should really be called `Schema` instead of `Presentation`. Or something else?

epatters:
Hi Sophie, is this PR subsumed by #433? If so, I'll close it and focus on the other one.

slibkind:
Yes it is!

PR #419, opened 2021-04-29
Extra axioms for theory of SMCs

0x0f0f0f:


epatters (approved):

epatters:
Thanks for the PR.

As a general point, there is no policy that the axioms for the theories should be minimal, but I think we should strive to make the choice of axioms as canonical as possible, subject to the constraints imposed by the GAT formalisms. (E.g., you cannot just say "this family of transformation is natural", you have to explicitly write down the equations.)

In this case, I included two of the unit equations because they are parallel with the two axioms for products and require several steps to deduce, but omitted the last one because it should be a trivial deduction from the other axioms.

PR #420, opened 2021-04-30
Refactor benchmarks to separate Catlab/LightGraphs

olynch:
This makes it so that the LightGraphs benchmarks can be separated from the Catlab benchmarks when doing analysis of benchmark results.

### DIFF
--- a/benchmark/Graphs.jl
+++ b/benchmark/Graphs.jl
@@ -71,7 +71,6 @@ end
 bench = SUITE["Graph"] = BenchmarkGroup()
 clbench = bench["Catlab"] = BenchmarkGroup()
 clvecbench = bench["Catlab-vectorized"] = BenchmarkGroup()
-clprojbench = bench["Catlab-proj"] = BenchmarkGroup()
 lgbench = bench["LightGraphs"] = BenchmarkGroup()
 
 n = 10000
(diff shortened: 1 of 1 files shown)


benchmark/Graphs.jl
```diff
@@ -121,76 +144,56 @@ end
...
-bench["star-graph-components"] = @benchmarkable connected_components($g)
-bench["star-graph-components-proj"] =
-  @benchmarkable connected_component_projection($g)
-bench["star-graph-components-lightgraphs"] =
-  @benchmarkable LG.connected_components($lg)
-bench["star-graph-components-proj-lightgraphs"] =
-  @benchmarkable lg_connected_components_projection($lg)
+bench = SUITE["SymmetricGraphConnComponent"] = BenchmarkGroup()
+clbench = bench["Catlab"] = BenchmarkGroup()
+lgbench = bench["LightGraphs"] = BenchmarkGroup()
```
epatters, line 158: Are the connected components benchmarks for symmetric graphs gone now?

epatters (commented):
Thanks! This is a better organizational scheme.

benchmark/Graphs.jl
```diff
@@ -77,41 +91,50 @@ end
...
-bench["star-graph-components"] = @benchmarkable connected_components($g)
-bench["star-graph-components-proj"] =
+clbench["star-graph-components"] = @benchmarkable connected_components($g)
+clprojbench["star-graph-components"] =
   @benchmarkable connected_component_projection($g)
-bench["star-graph-components-lightgraphs"] =
+lgbench["star-graph-components"] =
   @benchmarkable LG.weakly_connected_components($lg)
 
+bench = SUITE["GraphConnComponents"] = BenchmarkGroup()
```
epatters, line 122: Should these be defined before the connected components benchmarks above?

benchmark/Graphs.jl
olynch: I think you may be misunderstanding this code structure? The symmetric graph connected component benchmarks for Catlab go in this `clbench`, and the ones for LightGraphs go in this `lgbench`.

epatters: But I don't see any actual benchmarks for symmetric case, i.e., no connected components of symmetric graphs are computed? Compare the sections for graphs and symmetric graphs.

PR #425, opened 2021-05-07
DPO for ACSets

kris-brown:
Double pushout algorithm for ACSets. 

Many rewriting examples in test file.


src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+  k_components = Dict{Symbol, FinFunction}()
+  g_components = Dict{Symbol,Vector{Int}}()
+  K = ACSet{typeof(G).parameters...}()
+  offsets = Dict{Symbol,Vector{Int}}()
+
+
+  for comp in keys(l.components)
+
+    n_comp = nparts(G, comp)  # total # of elements in G
+    l_image = Set(l.components[comp].func)
```
epatters, line 39: You never want to directly access the `func` field of a `FinFunctionVector` because other kinds of `FinFunctions` may have different fields.

So this should be `Set(collect(l))`.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+
+    # m(L/l(G)): image of (complement of image of l) into G
+    orphans = sort(map(x->m[comp](x),
+                  filter(x->!(x in l_image),
+                      1:nparts(L,comp))))
```
epatters, line 44: Can replace `1:nparts(...)` with `parts(...)`.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+  return k => g
+end
+
+"""
+Compute offsets induced by removing particular elements from
+the list 1:n. E.g. if we remove elements 2 and 4 from 1:5,
+our offset is [0,_,1,_,2]
+(the offset at the removed element index is not meaningful)
+"""
+function reindex(n::Int, removed::Vector{Int})::Vector{Int}
```
epatters, line 91: This is just my immediate raction, but I feel like there should be a more elegant approach to handling the IDs that would avoid this `reindex` function. I'll think about it.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+Rewrite with explicit match
+"""
+function rewrite_match(L::ACSetTransformation, R::ACSetTransformation, m::ACSetTransformation)::AbstractACSet
+    @assert L.dom == R.dom
+    @assert L.codom == m.dom
+    @assert valid_dpo(L, m)
+    @assert is_natural(L)
+    @assert is_natural(R)
+    (k, _) = pushout_complement(L, m)
+    l1, _ = pushout(R, k).cocone.legs
```
epatters, line 115: `pushout(R, k).cocone.legs` -> `legs(pushout(R, k))` or even just `pushout(R, k)` because the colimit object is iterable.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+    return l1.codom
+end
+
+"""
+Don't explicitly choose the match
+"""
+function rewrite(L::ACSetTransformation,
```
epatters, line 122: This functions needs a more informative docstring.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+"""
+function valid_dpo(L::ACSetTransformation, m::ACSetTransformation, verbose::Bool=false)::Bool
+  return id_condition(L, m) && dangling_condition(L, m)
+end
+
+"""
+Does not map both a deleted item and a preserved item in L to the same item in G, or two distinct deleted items to the same.
+(Trivially satisfied if match is mono)
+"""
+function id_condition(L::ACSetTransformation, m::ACSetTransformation, verbose::Bool=false)::Bool
```
epatters, line 148: `verbose` should be a keyword-only argument (not positional), so this would read `id(..., m::ACSetTransformation; verbose::Bool=false)` (note the semicolon).

test/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,725 @@
+module DPOTest
+using Test
+using Catlab.CategoricalAlgebra.DPO
```
epatters, line 3: With a few exception, the top-level module re-export their submodules, you can replace most of these imports with:

```julia
using Catlab, Catlab.Theories, Catlab.CategoricalAlgebra
using Catlab.Graphics, Catlab.WiringDiagrams
```

test/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,725 @@
+module DPOTest
```
epatters, line 1: Naming convention for test module is `TestDPO`.

epatters (commented):
Thanks Kris, this a really start!

It will take me longer to fully digest this, but here is a first round of comments/questions. A lot of them are just minor things about Julia and Catlab conventions.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+    add_parts!(K, comp, length(g_components[comp]))
+
+    # Adjust lg function from I→G to refer to K using the offset
+    newFunc = [x - offsets[comp][x] for x in lm.components[comp].func]
+    k_components[comp] = FinFunction(newFunc, n_comp - length(orphans))
+  end
+
+  # Populate attributes for K (unchanged from G)
+  comps, arrows, src, tgt = typeof(G).parameters[1].parameters
+  attrs, srcs = typeof(G).parameters[2].parameters[3:4]
```
epatters, line 63: Unpacking parameters like this unreadable and error-prone. Use the accessors in `theories/Schema.jl` for this.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+  # Populate relations for K (same as G, with offset applied)
+  for (i, col) in enumerate(arrows)
+    src_, tgt_ = comps[src[i]], comps[tgt[i]]
+    new=[val - offsets[tgt_][val] for val in G[col][g_components[src_]]]
+    set_subpart!(K, col, new)
+  end
+
+  # Put together all information into new morphisms
+  k = ACSetTransformation(I, K; k_components...)
+  @assert is_natural(k)
```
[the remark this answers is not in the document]
kris-brown: Correct - these fail when dangling/id conditions aren't met, but now we check for those upfront.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,215 @@
...
+Compute a pushout complement, componentwise in Set. On a formal level:
+For each component, define K = G / m(L/l(I)). There is a natural injection g: K↪G
+For each component, define k as equal to the map l;m (every element in the image in G is also in K).
+
+Returns ACSetTransformations k and g such that (m, g) is the pushout of (l, k)
+
+Implementation-wise, elements of K are ordered in the same order as they appear in G.
+Construct an offset mapping to keep track of the location of elements of G within K.
+"""
+function pushout_complement(l::ACSetTransformation,m::ACSetTransformation)::Pair{ACSetTransformation,ACSetTransformation}
```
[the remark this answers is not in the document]
kris-brown: `l` being monic guarantees uniqueness, but I haven't seen much in the literature for the alternative. E.g. [this](https://core.ac.uk/download/pdf/82404673.pdf) paper merely restricts to monic `l` due to "ease of construction" and because it's "sufficient".

src/categorical_algebra/CospanRewrite.jl
```diff
@@ -0,0 +1,241 @@
...
+    # compute results w/ DPO
+    @assert valid_dpo(b1_b0, b0_d0)
+    @assert valid_dpo(x1_x0, x0_y0)
+    @assert valid_dpo(a1_a0, a0_c0)
+    b1_d1, d1_d0 = pushout_complement(b1_b0, b0_d0);
+    x1_y1, y1_y0 = pushout_complement(x1_x0, x0_y0);
+    a1_c1, c1_c0 = pushout_complement(a1_a0, a0_c0);
+    d1_d2, b2_d2 = pushout(b1_d1, b1_b2);
+    y1_y2, x2_y2 = pushout(x1_y1, x1_x2);
+    c1_c2, a2_c2 = pushout(a1_c1, a1_a2);
```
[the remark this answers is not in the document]
kris-brown: I tried a couple things but don't see a way to do that which clarifies (rather than obfuscates) what's going on. There are 9 pushouts and 3 pushout complements to compute, and it wasn't clear how to factor it into doing the same thing three times). I reorganized the code to highlight some common structure, as a first step.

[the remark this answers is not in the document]
jpfairbanks: Yeah I think we just need to include the figure in the docs and put a reference to the figure in the docstring. There is now other way to unserstand this stuff, but to look at the diagrams.

epatters:
I significantly simplified the implementation of `pushout_complement` by using `copy_parts!`, from which logic was partly duplicated here. I think the further simplifications and efficiency improvements may be possible, I'll save that for another time.

test/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,738 @@
...
+using Catlab.WiringDiagrams
+using Catlab.CategoricalAlgebra.FinSets
+using Catlab, Catlab.Theories, Catlab.CategoricalAlgebra
+using Catlab.Theories: AttrDesc
+
+######################################################################################
+######################################################################################
+# Wiring diagrams
+######################################################################################
+######################################################################################
```
epatters, line 14: Format section headers following the style guide.

test/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,738 @@
...
+
+GWD = WiringDiagram([:A,:A],[:D,:B])
+fv, gv, hv, fv2 = add_box!(GWD, f), add_box!(GWD, g), add_box!(GWD, h), add_box!(GWD, f);
+add_wires!(GWD, Pair[
+  (input_id(GWD),1) => (fv,1),
+  (fv,1) => (gv,1),
+  (gv,1) => (hv,1),
+  (hv,1) => (output_id(GWD),1),
+  (input_id(GWD),2) => (fv2,1),
+  (fv2,1) => (output_id(GWD),2)]);
```
epatters, line 26: Not necessary for this PR, but JSYK, the `@program` macro is convenient for avoiding all this boilerplate when constructing DWDs.

test/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,738 @@
...
+  concentration::Attr(S, Concentration)
+  tname::Attr(T, Name)
+  sname::Attr(S, Name)
+end
+
+const AbstractLabelledReactionNet = AbstractACSetType(TheoryLabelledReactionNet)
+const LabelledReactionNetUntyped = ACSetType(TheoryLabelledReactionNet, index=[:it,:is,:ot,:os])
+const LabelledReactionNet{R,C} = LabelledReactionNetUntyped{R,C,Symbol}
+
+LabelledReactionNet{R,C}(n, ts::Vararg{Union{Pair,Tuple}}) where {R,C} = begin
```
epatters, line 114: I assume this function and everything preceding it is a copy-paste from AlgebraicPetri? I don't like that we have to copy-paste ~100 LoC just to test DPO for Petri nets.

We already have a solid set of tests in this module (good job!), so perhaps we should move these tests to AlgebraicPetri once the functionality is merged into Catlab. Thoughts @mehalter?

test/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,738 @@
...
+R = CSetTransformation(I, line, V₁=[1], V₂=[1,2])
+m1 = CSetTransformation(Lspan, z_, V₁=[1], V₂=[1,2], E=[1, 2])
+m2 = CSetTransformation(Lspan, z_, V₁=[1], V₂=[2,1], E=[2, 1])
+
+@test is_isomorphic(parallel, rewrite_match(L, R, m1))
+@test is_isomorphic(merge, rewrite_match(L, R, m2))
+
+################################################
+################################################
+################################################
```
epatters, line 442: Section header missing text?

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,177 @@
...
+end
+
+"""
+Does not map both a deleted item and a preserved item in L to the same item in G, or two distinct deleted items to the same.
+(Trivially satisfied if match is mono)
+"""
+function id_condition(L::ACSetTransformation{CD, AD},
+                      m::ACSetTransformation{CD, AD};
+                      fail::Bool=false)::Bool where {CD, AD}
+  for comp in keys(L.components)
```
epatters, line 103: `L.components` -> `components(L)` here and throughout PR.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,177 @@
...
+For example, in the CSet of graphs:
+  e1
+1 --> 2   --- if e1 is not matched but either 1 and 2 are deleted, then e1 is dangling
+"""
+function dangling_condition(L::ACSetTransformation{CD, AD},
+                            m::ACSetTransformation{CD, AD};
+                            fail::Bool=false)::Bool where {CD, AD}
+  orphans = Dict()
+  for comp in keys(L.components)
+    image = Set(collect(L.components[comp]))
```
epatters, line 149: `L.components[comp]` -> `L[comp]`

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,177 @@
...
+"""
+    l
+  L ← I
+m ↓   ↓k
+  G ← K
+    g
+
+Compute a pushout complement, componentwise in Set. On a formal level:
+For each component, define K = G / m(L/l(I)). There is a natural injection g: K↪G
+For each component, define k as equal to the map l;m (every element in the image in G is also in K).
```
epatters, line 20: Docstring here and elsewhere should be hard-wrapped to 80 chars. Most editors come with or let you set up a keyboard shortcut for this, which is very convenient.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,177 @@
...
+  return id_condition(L, m;fail=fail) && dangling_condition(L, m;fail=fail)
+end
+
+"""
+Does not map both a deleted item and a preserved item in L to the same item in G, or two distinct deleted items to the same.
+(Trivially satisfied if match is mono)
+"""
+function id_condition(L::ACSetTransformation{CD, AD},
+                      m::ACSetTransformation{CD, AD};
+                      fail::Bool=false)::Bool where {CD, AD}
```
epatters, line 102: More useful than a debug switch that throws an error would be a switch to returns one or more pairs of offending elements when the ID conditions fails and `nothing` otherwise. Thus, you could actually use this information programmatically if you wanted to.

Also, when you *do* want to raise an error like this, use `error(...)` or `throw ...` rather than `@assert false`.

src/categorical_algebra/DPO.jl
```diff
@@ -0,0 +1,177 @@
...
+  for comp in keys(L.components)
+    m_comp = x->m[comp](x)
+    image = Set(collect(L.components[comp]))
+    image_complement = filter(x->!(x in image), 1:nparts(codom(L),comp))
```
epatters, line 106: Not important, but it is slightly shorter to write `parts(...)` than `1:nparts(...)`.

epatters (commented):
Hi Kris, this will be really great. Some feedback focusing on the core DPO part. Will revisit the cospan rewriting later.

test/categorical_algebra/DPO.jl
jpfairbanks: I agree, the rewriting on Petri will get tested in the form of the notebooks @bosonbaas is building for the model space exploration. I think we can trust that the implementation is correct based on other schemas and then put these tests in AlgebraocPetri.

epatters:
I've merged the spin-off PR #446!

Let's create a new PR for the structured cospan rewriting. I have some ideas about how to both simplify and generalize the code, which I will describe in a new issue.

PR #429, opened 2021-05-21
Automorphisms of CSets

kris-brown:
Based on implementation specified here
https://www.math.unl.edu/~aradcliffe1/Papers/Canonical.pdf

Could this eventually replace the Base.hash method for CSets?

It could be straightforwardly extended to ACSets.

### DIFF
--- a/src/categorical_algebra/Automorphisms.jl
+++ b/src/categorical_algebra/Automorphisms.jl
@@ -1,0 +1,241 @@
+module Automorphisms
+export canonical_hash, autos, apply_automorphism, color_refine
+using ..CSets
+using ...Theories
+
+# Color assigned to each elem of each compoennt
+CDict = Dict{Symbol, Vector{Int}}
+
+# The maximum color of an empty color list is 0
+function max0(x::Vector{Int})::Int
+    return isempty(x) ? 0 : maximum(x)
+end
+
+function check_auto(x::CDict)::Nothing
+    for perm in values(x)
+        @assert length(perm) == length(Set(perm))
+    end
+end
+
+"""Apply a coloring to a Cset to get an isomorphic cset"""
+function apply_automorphism(c::CSet,d::CDict)::CSet
+    check_auto(d)
+    new = deepcopy(c)
+    tabs, arrs, srcs, tgts = (typeof(c).parameters[1]).parameters
+    for (arr, srci, tgti) in zip(arrs,srcs,tgts)
+        src, tgt =  [tabs[x] for x in [srci, tgti]]
+        set_subpart!(new, d[src],arr, d[tgt][c[arr]])
+    end
+    return new
+end
+
+"""Lexicographic minimum of all automorphisms"""
+function canonical_iso(g::CSet)::CSet
+    isos = sort([apply_automorphism(g, Dict(a)) for a in autos(g)], by=string)
+    return isempty(isos) ? g : isos[1]
+end
+
+function canonical_hash(g::CSet)::UInt64
+    return hash(string(canonical_iso(g)))
+end
+
+# Store how many of each color (for each in arrow) has a
+# particular point as its target, as well as what color
+# (for each out arrow) the data point is the src of.
+CDataPoint = Tuple{Vector{Vector{Int}}, Vector{Int}}
+# Data req' to color a CSet (each element of each component)
+CData = Dict{Symbol, Vector{CDataPoint}}
+
+"""
+This could be extended to add extra automorphism-invariant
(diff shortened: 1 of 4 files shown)


src/categorical_algebra/Automorphisms.jl
```diff
@@ -0,0 +1,306 @@
...
+export canonical_hash, canonical_iso, autos, apply_automorphism, color_refine, pseudo_cset, pseudo_cset_inv
+using ..CSets
+using ...Theories
+using ...Theories: adom, attr, data, attr, adom, acodom, AttrDesc
+using Catlab.Present
+
+# Color assigned to each elem of each compoennt
+CDict = Dict{Symbol, Vector{Int}}
+
+"""
```
jpfairbanks, line 11: This should have a docstring header like 

```julia
"""    pseudo_cset(g::ACSet{CD,AD})::Tuple{CSet, Vector{Vector{Any}}} where {CD, AD}

To compute automorphisms of Attributed CSets, we create a pseudo CSet which has 
components for each data type.
"""
```

src/categorical_algebra/Automorphisms.jl
```diff
@@ -0,0 +1,306 @@
...
+
+function check_auto(x::CDict)::Nothing
+    for perm in values(x)
+        @assert length(perm) == length(Set(perm))
+    end
+end
+
+"""Apply a coloring to a Cset to get an isomorphic cset"""
+function apply_automorphism(c::CSet,d::CDict)::CSet
+    check_auto(d)
```
jpfairbanks, line 70: if you make `check_auto` into a predicate, than this line should be 

```julia
check_auto || error("d is not auto")
```
but with an informative error message. This will ensure that when the check auto predicate fails, the user sees a meaningful error.

src/categorical_algebra/Automorphisms.jl
```diff
@@ -0,0 +1,306 @@
...
+                     [osr[i] for osr in out_subres])
+                    for i in 1:nparts(g,tgt)]
+    end
+    return res
+end
+
+"""
+Construct permutation σ⁻¹ such that σσ⁻¹=id
+"""
+function invert_perm(x::CDict)::CDict
```
jpfairbanks, line 160: would `Base.invperm` be helpful here? From the docs

```
help?> Base.invperm
  invperm(v)

  Return the inverse permutation of v. If B = A[v], then A == B[invperm(v)].

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> v = [2; 4; 3; 1];

  julia> invperm(v)
  4-element Array{Int64,1}:
   4
   1
   3
   2
```

jpfairbanks (changes requested):
This is looking pretty good. I have some suggestions addressed as inline comments. All the functions that doc strings need to have the docstring start with the function signature indented by 4 spaces. That will ensure that the docstring includes a code formatted version of that function signature.

src/categorical_algebra/Automorphisms.jl
kris-brown: Perfect, thanks!

src/categorical_algebra/Automorphisms.jl
```diff
@@ -0,0 +1,306 @@
...
+    function invert_comp(xs::Vector{Int})::Vector{Int}
+        return [findfirst(==(v), xs) for v in eachindex(xs)]
+    end
+    return Dict([k=>invert_comp(v) for (k, v) in collect(x)])
+end
+
+"""
+Compose permutations
+"""
+function compose_perms(x::CDict, y::CDict)::CDict
```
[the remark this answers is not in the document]
kris-brown: This would be nice! I tried this - it requires some boxing/unboxing of vectors that made the rest of the code look more complicated (and introduced an infinite loop bug that I don't know how to fix), so I'm not sure this is worth it to save the line of duplicated code.

https://github.com/kris-brown/Catlab.jl/commit/c66ea2f1d36fff3462342359bb405df9deab1027

jpfairbanks:
@kris-brown @epatters @olynch, how close are we to getting this merged? For the adjoint school, we need to count matches in DPO and I think we need to be counting isomorphism classes because the automorphisms of the match shouldn't count in the computation of the boltzman distribution.

epatters:
I haven't even started looking at this one yet, and I think Kris has a backlog related to #425 as well. Definitely not happening this week.

If needed immediately for Adjoint School, this could be put temporarily in a different repo.

jpfairbanks:
Copy pasta it is!

PR #431, opened 2021-05-25
ENH: add category of elements to CategoricalAlgebra

jpfairbanks:
This PR implements the conversion of a Database instance into a Knowledge Graph by constructing the category of elements. I'm happy to rename anything especially the module `CatElements` was chosen because I couldn't use `Elements` for both the module and Type name.


jpfairbanks:
Looks like I did not add the relevant files to the test suite.  Because I was running the new test file directly in VS Code.

jpfairbanks:
@slibkind could you check that this category of elements is right? What do you think of my naming convention for naming the Objects and Homs in the Presentation you get from the category of elements? 

https://github.com/AlgebraicJulia/Catlab.jl/pull/431/files#diff-b89e61a4166d18a16b211069a1d0b033628ad496c0071cab2a608a60d7a46d98R57

slibkind:
I'll take a look at this tomorrow afternoon, if that's ok!

slibkind:
ThElements` more or less captures the same data as (1) two `FreeDiagrams` where the first captures the data of (El, Arr, src, tgt) and the second captures the data of (Ob, Hom, dom, cod) and (2) an ACSet transformation (El, Arr, src, tgt) -> (Ob, Hom, dom, cod). This feel more honest to the category theory than encoding it as a single ACSet. What do you think @jpfairbanks?

jpfairbanks:
I agree that `elements` returning a morphism rather than an object is more true to the CT. We need some general machinery for internalizing categorical constructions about C-Sets as C-Sets over a new indexing category. This is another example of things like how C-Set morphsims can be stored as a `Cx2-Set`s. Or the isomorphism between `[C,Set]/X` and `[elements(X), Set]`. So maybe we need to have what is currently `ThElements` be constructed as `ThGraph x ThArrow` and then a general implementation of the internalization of CSet morphisms as objects in `ThGraph x ThArrow`. The question for @epatters is if we want to wait for that machinery to be implemented before merging this part of the implementation. I think we could refactor it into a function that returns an `ACSetTransformation` and then a second function that converts that transformation into an instance of `ThElements`.

slibkind:
So the idea would be implement a method that takes `X::ACSet` to an acset transformation representing the projection from a `FreeDiagram` representing the category of elements to a `FreeDiagram` representing the underlying category `C`. Then a second (much more general) method that takes an acset transformation to the presentation representing the transformation.

This feels like it nicely represents the underlying category theory!

jpfairbanks:
Yeah, so what should we call the Theory of CSet Morphisms? I can refactor this code to do that. But it needs a name. How about `ThCSetHom` for the presentation and then `CSetHom` for the corresponding julia type representing the instances? In a richer type system, we could actually use the existing `CSetTransformation` type as an instance of the `AbstractACSetType(ThCSetHom)` but I think we will have to use a different name.

slibkind:
`ThCSetHom` and `CSetHom` make sense to me!

src/categorical_algebra/CatElements.jl
```diff
@@ -0,0 +1,82 @@
...
+    (El, Arr, Ob, Hom)::Ob
+    Name::Data
+    src::Hom(Arr, El)
+    tgt::Hom(Arr, El)
+    dom::Hom(Hom, Ob)
+    cod::Hom(Hom, Ob)
+    πₑ::Hom(El, Ob)
+    πₐ::Hom(Arr, Hom)
+    nameo::Attr(Ob, Name)
+    nameh::Attr(Hom, Name)
```
olynch, line 23: Can we have the laws here? Just for documentation sake...

src/categorical_algebra/CatElements.jl
```diff
@@ -0,0 +1,82 @@
...
+        c = ob(CD)[ci]
+        d = ob(CD)[di]
+        nc = nparts(X, c)
+        add_parts!(Y, :Arr, nparts(X, c), src = els[ci], tgt = view(els[di], X[f]), πₐ = i)
+    end
+    return Y
+end
+
+
+function Presentation(X::AbstractElements)
```
olynch, line 57: Can this have a docstring too?

src/categorical_algebra/CatElements.jl
```diff
@@ -0,0 +1,82 @@
+module CatElements
+using Catlab
+using ..CSets
+using ..FinSets
+using ...Present, ...Theories
+# using ..CSetDataStructures
```
olynch, line 6: Why is this commented? Should it be deleted?

src/categorical_algebra/CatElements.jl
```diff
@@ -0,0 +1,82 @@
...
+const Elements = ACSetType(ThElements, index=[:src, :tgt, :πₑ, :πₐ])
+
+"""    elements(X::AbstractACSet)
+
+construct the category of elements from an ACSet. This only correctly handles the CSet part.
+This transformation converts an instance of C into a Graph homomorphism. The codomain of the
+homomorphism is a graph shaped like the schema. This is one half of the isomorphism between
+databases and knowledge graphs.
+"""
+function elements(X::ACS) where
```
olynch, line 36: In theory, we should have conversion between schemas and graphs implemented generically in another place, and this should use that, but we can do that in another PR if you don't want to do that here.

olynch, line 36: If this only correctly handles the CSet part, then it should only take in CSets, i.e. `ACS <: AbstractCSet{CD}`.

olynch (changes requested):
Generally looks good, just a couple specific questions/comments.

Could also do with a bit more documentation, maybe a link to an nlab page in a comment, and a description of the design choices made in depicting the category of elements in our framework (because there are obviously other ways to do it; it could have been a GAT).

epatters:
@jpfairbanks, from the discussion above, it sounds like you had plans to refactor this code? I like the idea of having a generic function to convert ACSetTransformations into ACSets.

In addition to Owen's suggestions, the indent width of this code should be converted from 4 spaces to 2.

jpfairbanks:
Yeah, I want to get back to this. I got stuck implementing the inverse construction. Now that Sigma migration is merged that seems super helpful. Is there a simple way to go back from a cat of elements to the CSet using sigma migration?

slibkind:
Yes! If X: C -> Set, then we have a projection from the category of elements back to C, p: El(X) -> C. Suppose that we have an El(X)-Set Y. Then the pushforward of Y via Sigma(p) is the C-Set. 

E.g. if Y is a bipartite graph then the pushforward of Y along p is the underlying graph.

jpfairbanks:
I think I've addressed all the blocking comments on this PR. The refactoring mentioned above can wait until we have a better mechanism for all the "CSets are closed under many constructions" like the fact that we can store CSet Homs as CSets on a bigger schema.

epatters:
Thanks James!

PR #432, opened 2021-05-26
Started CI/CD for Streamlining Review Process

mehalter:
This adds a couple github actions. One will auto comment a review checklist to new PRs as well as auto request reviewers when a PR is marked ready for review or a non-draft PR is opened that doesn't contain the string "DO NOT REVIEW"

The reviewer checklist is WIP 

Signed-off-by: Micah Halter <micah@balena.io>


mehalter:
Closing to test new opening of the PR.

mehalter:
# TODO

- [ ] Review github actions and approve/tweak the suggested workflow
- [ ] Finalize review checklist

epatters:
Thanks for starting this, Micah. Having a PR checklist will be great. I'll follow up later to refine the list.

OTOH, my current feeling is that having a bot automatically request reviews from 4-5 people seems excessive. I think that a manual process involving a combination of triage by me and interested devs signing themselves up for reviews should work well and generate less noise for people.

So I am thinking to just have the PR checklist action, not the review request action. Thoughts on this from anyone?

mehalter:
I think that sounds good. Yeah, I was on the fence about doing the automatic review requests pinging everyone being a little excessive. I think doing just the review checklist comment once a reviewer has been requested as well as the status check to make sure all of the reviewing points have been verified by the reviewer would be a good thing. This automatic reviewer action might be useful later on as you can have it request different reviewers based on the files that have changes. For instance if someone becomes in charge of reviewing documentation changes, that would automatically request their review when `docs/**.md` changes or something. But that would be easy to add later especially as the distributed reviewing process is in place

.github/reviewers.yml
```diff
@@ -0,0 +1,17 @@
+# documentation of options: https://github.com/marketplace/actions/auto-request-review
+
+# eventually move this to using GitHub Teams (we have a developers team)
+# requires the use of a personal access token with access to teams
+# for now, leaving it as an explicit list
+reviewers:
```
jpfairbanks, line 6: Is this the list of people who would get auto-tagged for reviews? I agree that we don't need to spam people about reviews yet. People can explicitly request a relevant review when they open their PR.

jpfairbanks (commented):
Does the list of default

epatters (approved):

epatters:
I've revised the checklist to include only the most important checks, while also linking to the full list of guidelines on the wiki. This should be ready to merge. We can make further changes if needed as we experiment with the workflow.

PR #433, opened 2021-05-29
EHN: Left pushforward data migration

slibkind:
This PR implements the data migration functor `ΣF` for `F: C -> D`. `ΣF` takes a C-Set `X` and produces a D-Set `ΣF(X)`

### DIFF
--- a/src/categorical_algebra/DataMigration.jl
+++ b/src/categorical_algebra/DataMigration.jl
@@ -1,0 +1,159 @@
+""" Data Migration functors
+"""
+module DataMigration
+
+export Σ, Functor
+
+using ...Theories
+using ...Theories: CatDesc, AttrDesc, ob, hom, dom, codom
+using ...CSetDataStructures
+using ..FinSets, ..CSets, ..Limits, ..FreeDiagrams
+using ...Graphs
+using ...Present
+
+import ...CategoricalAlgebra.FreeDiagrams: FreeDiagram
+
+""" Functor
+
+A functor ``F: \\mathcal{C} \\to \\mathcal{D}`` consists of a map of objects and a 
+map of homomorphisms from a domain category ``\\mathcal{C}``  to a codomain
+category ``\\mathcal{D}``.
+"""
+struct Functor{Schema, Ob, Hom}
+  FOb::Dict{Ob, Ob}
+  FHom::Dict{Hom, Hom}
+  dom::Presentation{Schema, Symbol}
+  codom::Presentation{Schema, Symbol}
+end
+
+""" Inv(F::Functor)
+
+For a functor ``F: \\mathcal{C} \to \\mathcal{D}``, 
+returns a function which when given an object (resp. morphism)
+of ``\\mathcal{D}``, returns an array of objects (resp. morphisms) 
+of ``\\mathcal{C}`` in the preimage.
+"""
+function Inv(F::Functor{Schema, Ob, Hom}) where {Schema, Ob, Hom}
+  y::Union{Ob, Hom} -> begin
+    y isa Ob ? D = F.FOb : D = F.FHom
+    [k for (k,v) in D if v == y]
+  end
+end
+
+
+""" add_hom!(d::FreeDiagram{Ob, Hom}, src_ob::Ob, tgt_ob::Ob, hom::Hom)
+
+Adds a hom to `d` from the vertex with object `src_ob` to the vertex with object `tgt_ob`.
+"""
+function add_hom!(d::FreeDiagram, src_ob, tgt_ob, hom) 
+  src_idx = first(incident(d, src_ob, :ob))
+  tgt_idx = first(incident(d, tgt_ob, :ob))
+  return add_edge!(d, src_idx, tgt_idx, hom = hom)
+  end
+
+""" FreeDiagram(pres::Presentation{Schema, Symbol})
+
+Returns a `FreeDiagram` whose objects are the generating objects of `pres` and 
+whose homs are the generating homs of `pres`.
+"""
+function FreeDiagram(pres::Presentation{Schema, Symbol}) where {Schema}
+  obs = Array{FreeSchema.Ob}(generators(pres, :Ob))
+  homs = Array{FreeSchema.Hom}(generators(pres, :Hom))
+  doms = map(h -> generator_index(pres, nameof(dom(h))), homs)
+  codoms = map(h -> generator_index(pres, nameof(codom(h))), homs)
+  return FreeDiagram(obs, collect(zip(homs, doms, codoms)))
+end
+
+
+""" Σ(F::Functor)
+
+which takes a ``\\mathcal{C}``-Set ``X: \\mathcal{C} \\to \\mathsf{Set}`` 
+and produces the pushforward ``\\Sigma_F(X): \\mathcal{D} \\to \\mathsf{Set}``.
+
+``\\mathcal{D}`` must be a free schema with no cycles. 
+"""
+function Σ(F::Functor{Schema, Ob, Hom}) where {Schema, Ob, Hom}
+  diagramC = FreeDiagram(F.dom)
+  diagramD = FreeDiagram(F.codom)
+  FInv = Inv(F)
+
+  comma_cats = FreeDiagram{FreeDiagram, ACSetTransformation}()
+  add_vertices!(comma_cats,
+    nparts(diagramD, :V), 
+    ob = map( _ ->  FreeDiagram{Pair{Ob, Hom}, Hom}(), parts(diagramD, :V))
+  )
+
+  for d in topological_sort(diagramD)
+    F∇d = ob(comma_cats, d)
+    id_d = id(ob(diagramD, d))
+
+    # If FOb[c] = d, add an objects (c, id(d)) to F∇d
+    preimage_d = FInv(ob(diagramD, d)) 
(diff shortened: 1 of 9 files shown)


slibkind:
A few notes: 
- This PR implements a struct `Functor`. Do we want to make this more general and have accompanying methods?
- A potential todo is to refactor the [`migrate!` method](https://github.com/AlgebraicJulia/Catlab.jl/blob/master/src/categorical_algebra/CSets.jl#L568) to match `Σ`

src/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,159 @@
...
+
+""" add_hom!(d::FreeDiagram{Ob, Hom}, src_ob::Ob, tgt_ob::Ob, hom::Hom)
+
+Adds a hom to `d` from the vertex with object `src_ob` to the vertex with object `tgt_ob`.
+"""
+function add_hom!(d::FreeDiagram, src_ob, tgt_ob, hom) 
+  src_idx = first(incident(d, src_ob, :ob))
+  tgt_idx = first(incident(d, tgt_ob, :ob))
+  return add_edge!(d, src_idx, tgt_idx, hom = hom)
+  end
```
olynch, line 52: indentation

src/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,159 @@
...
+returns a function which when given an object (resp. morphism)
+of ``\\mathcal{D}``, returns an array of objects (resp. morphisms) 
+of ``\\mathcal{C}`` in the preimage.
+"""
+function Inv(F::Functor{Schema, Ob, Hom}) where {Schema, Ob, Hom}
+  y::Union{Ob, Hom} -> begin
+    y isa Ob ? D = F.FOb : D = F.FHom
+    [k for (k,v) in D if v == y]
+  end
+end
```
olynch, line 41: maybe better to make a struct InvFunctor and then overload calling

src/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,159 @@
...
+
+
+""" Σ(F::Functor)
+
+which takes a ``\\mathcal{C}``-Set ``X: \\mathcal{C} \\to \\mathsf{Set}`` 
+and produces the pushforward ``\\Sigma_F(X): \\mathcal{D} \\to \\mathsf{Set}``.
+
+``\\mathcal{D}`` must be a free schema with no cycles. 
+"""
+function Σ(F::Functor{Schema, Ob, Hom}) where {Schema, Ob, Hom}
```
olynch, line 75: again, might be best to make a struct SigmaFunctor and then overload calling

olynch: also, would be nice if this function was split up a little

src/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,159 @@
...
+  diagramC = FreeDiagram(F.dom)
+  diagramD = FreeDiagram(F.codom)
+  FInv = Inv(F)
+
+  comma_cats = FreeDiagram{FreeDiagram, ACSetTransformation}()
```
olynch, line 80: the creation of comma_cats could be split out

src/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,159 @@
...
+      for (c, f) in ob(F∇d, vs)
+        for h in FInv(f)
+          dom(h) == c && add_hom!(F∇d, c => f, codom(h) => id_d, h)
+        end
+      end
+    end 
+  end # end of populating comma_cats 
+
+  function pushforward(X::ACSet) 
+    Presentation(X) == F.dom || error("X has the wrong Schema")
```
olynch, line 126: make this an `@assert`, so w can turn it off if we want

src/categorical_algebra/DataMigration.jl
olynch: and if there was documentation of the overall strategy

olynch (changes requested):

src/categorical_algebra/DataMigration.jl
jpfairbanks: Yeah, then you could define that as two methods instead of a using `isa` that feels more natural with the way that a functor is defined. We think of a functor as any thing that has an action on objects and an action on morphisms. The only downside I see to using this interface is that we would need to wrap things that are already callables. https://docs.julialang.org/en/v1/manual/methods/#Function-like-objects If we chose an interface like `fapply(F::AbstractFunctor, X::Ob)` and `fapply(F::AbstractFunctor, f::Hom)` then we could use `fapply(f::Function, x::FinSet)` where `f` is just a regular julia function. 

I think that since our objects and homs live in wrapper types anyway, this isn't that big of an issue and the benefit of `F(X::Ob)` and `F(f::Hom)` as the minimal syntax necessary to apply a functor is worth it.

src/categorical_algebra/DataMigration.jl
jpfairbanks: I like the idea of `Σ(F::AbstractFunctor)::SigmaMigration <: AbstractFunctor` also, we always been ASCII names in case your text editor is broken and you are writing in an old version of `vi` with no unicode support.

src/theories/Schema.jl
```diff
@@ -186,3 +189,18 @@ function attrs_by_codom(AD::Type{T}) where
 end
 
 SchemaType(pres::Presentation{Schema}) = (CatDescType(pres),AttrDescType(pres))
+
+"""
```
jpfairbanks, line 193: In order to get a code block in the autogenerated docs you can add

```julia
    Presentation(CD::Type{T}, AD::Type{S}) where {T <: CatDesc, S <: AttrDesc}
```

to the first line of the docstring with 4 leading spaces (which is a code block in markdown)

jpfairbanks (commented):
I just came back to this and realized I've had a pending review for over a week! 😱  I think after my simple fixes are done, this is ready for @epatters to take a look.

test/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,180 @@
...
+  Dict(V => S, E => S), 
+  Dict(s => id(S), t => ϕ, weight => compose(ϕ, label)),
+  TheoryWeightedGraph, TheoryLabeledDDS
+)
+
+@test wg == WeightedGraph{Int}(ldds, F) 
+
+
+# Left Pushforward data migration
+
```
epatters, line 77: Extra newline.

test/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,180 @@
...
+#################################
+
+Σ = Sigma
+
+@present ThBipartite(FreeSchema) begin
```
epatters, line 82: This theory is already in `Catlab.Graphs`, where it is called `TheoryUndirectedBipartiteGraph`.

test/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,180 @@
...
+)
+
+idF = Functor(
+  Dict(VG => VG, EG => EG),
+  Dict(srcG => srcG, tgtG => tgtG),
+  TheoryGraph, TheoryGraph
+)
+
+ΣF =  Σ(F)
+X = ACSetType(ThBipartite, index=[:src, :tgt])()
```
epatters, line 107: Use `UndirectedBipartiteGraph` from `Catlab.Graphs` here and below.

test/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,180 @@
...
+
+@test Σ(idF)(Y) == Y
+
+@present ThSpan(FreeSchema) begin 
+  (L1, L2, A)::Ob
+  l1::Hom(A, L1)
+  l2::Hom(A, L2)
+end
+
+X = @acset ACSetType(ThSpan, index=[:l1, :l2]) begin
```
epatters, line 134: Especially since future versions of `ACSetType` may involve more metaprogramming, I think we should write

```julia
const Span = ACSetType(ThSpan, index=[:l1, :l2])
```

and avoid putting `ACSetType` calls inline. (Actually, I'm surprised the above works, but that's cool I guess :)

epatters (commented):
Thanks Sophie! I just left small comments. I think once you've addressed the outstanding comments this should be ready to merge.

In the longer term, we might want to improve the algorithm for the left pushforward, but I'll create a separate issue for that since it might be a lot of work.

src/categorical_algebra/DataMigration.jl
```diff
@@ -0,0 +1,312 @@
...
+# interface for abstract functor
+dom(F::Functor) = F.dom
+codom(F::Functor) = F.codom
+Ob(F::Functor) = F.FOb
+Hom(F::Functor) = F.FHom
+
+# To construct a functor where FOb and FHom are keyed by the symbols of 
+# the domain and codomain presentations. Turn them into a dictionaries 
+# whose keys and values are homs.
+function Functor(FOb::Dict{Symbol, Symbol}, FHom::Dict{Symbol, H},
```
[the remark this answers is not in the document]
slibkind: I think the typing here is necessary. In an old iteration we stored FOb and FHom as dictionaries of symbols. However, here `Functor` expects them to be dictionaries of the actual generators. This method is for turning the symbol implementation into the generator implementation.

epatters:
Thanks @slibkind!

PR #434, opened 2021-06-08
BUILD: Fix error in github actions interpreting |

mehalter:
The `yml` before this PR is completely valid. I think there is a bug in the Github Actions interpreter where it is getting confused. Hoping that this will fix things.


epatters (approved):
Thanks for the fixes, Micah! It's too bad that GitHub's YAML parser doesn't work on valid YAML.

PR #441, opened 2021-06-22
BUILD: Remove dependency on comment-on-pr github action

mehalter:
Signed-off-by: Micah Halter <micah@balena.io>

### DIFF
--- a/.github/workflows/reviewchecklist.yml
+++ b/.github/workflows/reviewchecklist.yml
@@ -9,28 +9,31 @@ jobs:
     name: Review Checklist
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


mehalter:
@epatters This removes dependency on the comment-on-pr GitHub actions and does this with just vanilla `github-script`. I tested this with the `pull_request` trigger and it should work with the `pull_request_target` action now. This is a much cleaner approach and is much more future proof. 🤞🏼

jpfairbanks:
Looks good to me!

epatters:
Thanks @mehalter!

epatters (approved):

PR #443, opened 2021-06-23
Homomorphism search, monic or iso on particular components

kris-brown:
For some applications, it is useful to restrict the monic constraint to just certain CSet components. This is implemented here by extra optional keyword arguments `isos` and `monics` which are vectors of symbols.


epatters (approved):
Thanks Kris! I made some small changes myself.

PR #453, opened 2021-07-08
Reference Semagrams in README

olynch:
In the README, it says that Catlab is not meant to be a graphical editor; I edited this to say that Catlab *is* the backend to a graphical editor, however!


epatters (changes requested):
Thanks Owen! Would you mind propagating this change to the [docs homepage]( https://github.com/AlgebraicJulia/Catlab.jl/blob/master/docs/src/index.md), which copies some text from the README.

epatters (approved):

PR #457, opened 2021-07-09
ENH: add function for visualizing CPortGraphs

bosonbaas:
Provides visualization tooling for CPortGraphs. An example of how a star CPortGraph is visualized is provided below.

```julia
to_graphviz(pg, edge_attrs=Dict(:len => "1.0"), port_labels=true, box_labels=true)
```

image

### DIFF
--- a/src/graphics/GraphvizWiringDiagrams.jl
+++ b/src/graphics/GraphvizWiringDiagrams.jl
@@ -13,6 +13,7 @@ using StaticArrays
 import ...Theories: HomExpr
 using ...WiringDiagrams, ...WiringDiagrams.WiringDiagramSerialization
 using ...CategoricalAlgebra.CSets, ...Graphs, ..GraphvizGraphs
+import ...CategoricalAlgebra: migrate!
 import ..Graphviz
 import ..GraphvizGraphs: to_graphviz
 using ..WiringDiagramLayouts: BoxLayout, PortLayout, WirePoint,
@@ -94,7 +95,7 @@ function to_graphviz(f::WiringDiagram;
       port_map[Port(v,kind,i)] = node_id
     end
   end
-  
+
   # Invisible nodes for incoming and outgoing wires.
   if outer_ports
     gv_box = graphviz_outer_box(f;
@@ -103,7 +104,7 @@ function to_graphviz(f::WiringDiagram;
     update_port_map!(input_id(f), OutputPort, gv_box.input_ports)
     update_port_map!(output_id(f), InputPort, gv_box.output_ports)
   end
-  
+
   # Visible nodes for boxes.
   default_attrs = default_directed_graphviz_attrs
   cell_attrs = merge(default_attrs.cell, Graphviz.as_attributes(cell_attrs))
@@ -115,7 +116,7 @@ function to_graphviz(f::WiringDiagram;
     update_port_map!(v, InputPort, gv_box.input_ports)
     update_port_map!(v, OutputPort, gv_box.output_ports)
   end
-  
+
   # Edges.
   for (i, wire) in enumerate(wires(f))
     source, target = wire.source, wire.target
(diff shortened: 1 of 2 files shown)


src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -623,4 +624,64 @@ inverse_rank_dir(rank_dir::String) = @match rank_dir begin
...
+- `box_labels=false`: whether to label boxes with their number
+- `port_labels=false`: whether to label ports with their number
+- `graph_attrs=Dict()`: top-level graph attributes
+- `node_attrs=Dict()`: top-level node attributes
+- `edge_attrs=Dict()`: top-level edge attributes
+
+# TODO:
+The lack of ports might be able to be resolved by introducing an extra node per
+port which is connected to its box with an edge of length 0.
+"""
```
epatters, line 648: You should attach your beautiful docstring to the exported function `to_graphviz` so that people will actually see it!

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -623,4 +624,64 @@ inverse_rank_dir(rank_dir::String) = @match rank_dir begin
...
+  # Generate a map from the global index of a port to the box-specific index
+  port2box = Array{Tuple{Int, Int}}(undef, nparts(cp, :Port))
+  for (b, p) in enumerate(incident(cp, 1:nparts(cp, :Box), :box))
+    port2box[p] .= [(b, i) for i in 1:length(p)]
+  end
+
+  g′ = Graphs.Graph()
+
+  add_parts!(g′, :V, nparts(cp, :Box))
+  add_parts!(g′, :E, nparts(cp, :Wire), src=cp[cp[:src], :box], tgt=cp[cp[:tgt], :box])
```
epatters, line 666: This looks like a pullback data migration with `V => Box`, `E => Wire`, `src => compose(src, box)` and `tgt => compose(tgt, box)`.

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -13,6 +13,7 @@ using StaticArrays
 import ...Theories: HomExpr
 using ...WiringDiagrams, ...WiringDiagrams.WiringDiagramSerialization
 using ...CategoricalAlgebra.CSets, ...Graphs, ..GraphvizGraphs
+import ...CategoricalAlgebra: migrate!
```
epatters, line 16: Import is added but not used?

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -623,4 +624,64 @@ inverse_rank_dir(rank_dir::String) = @match rank_dir begin
...
+- `graph_attrs=Dict()`: top-level graph attributes
+- `node_attrs=Dict()`: top-level node attributes
+- `edge_attrs=Dict()`: top-level edge attributes
+
+# TODO:
+The lack of ports might be able to be resolved by introducing an extra node per
+port which is connected to its box with an edge of length 0.
+"""
+
+function property_graph(cp::CPortGraph;
```
epatters, line 650: For consistency with existing code in this module, this function should be called `to_graphviz_property_graph`.

epatters (changes requested):
Thanks Andrew, especially for the nice docstring.

src/graphics/GraphvizWiringDiagrams.jl
bosonbaas: Thanks for catching that!

src/graphics/GraphvizWiringDiagrams.jl
bosonbaas: Didn't realize you could do composition with the data migration tools, that's neat! I've changed the implementation to using that now.

src/graphics/GraphvizWiringDiagrams.jl
bosonbaas: Now it's used for data migration as mentioned above. I think I put this in originally because I had hopes that data migration would work for this, I'm glad it actually does!

epatters (approved):
Thanks Andrew!

PR #459, opened 2021-07-10
(Co)cartesian category instances from (co)products

### LINKED ISSUE #449: Products and sums of morphisms in FinSet
A natural construction to make is the product of morphisms `f:A->B` and `g:C->D`, obtaining `f x g: A x C -> B x D`. This can be done by using `proj1` and `proj2` on the product of the domains, composing these with `f` and `g` respectively, then finally using the `pair` function.

epatters:
Resolves #449 in a generic way.


epatters:
It just occurred to me that I should use this for the category of acsets too, but I'll do that later.

Issue #455, opened 2021-07-09
`@acset` macro doesn't work with variables that aren't compile time constants


jpfairbanks:
I thought this would work but it doesn't. It is kind of obvious why it doesn't work now that I understand it. 

MWE:
```julia
# This works
n = 10
@acset Graph begin
       V = n
       E = n
       src = n
       tgt = 1:n
end

# but this throws a Variable Not Defined error on n
function star(n)
  @acset Graph begin
     V = n
     E = n
     src = n
     tgt = 1:n
   end
end
```

The variables used in the `@acset` macro are resolved at macro expansion time and not function call time. So you can use variables that already are bound like compile time constants or scripting variables in your acset defintion, but this doesn't extend to local variables that don't have definitions yet. 

Should we add a feature to the acset macro like

```julia
@acset star(n::Int)::Graph begin
     V = n
     E = n
     src = n
     tgt = 1:n
   end
end
```

would make sense to me. Then the argument variables wouldn't get evaluated until the star function was called?

epatters:
Currently we have

```julia
julia> @macroexpand(@acset Graph begin
              V = n
              E = n
              src = n
              tgt = 1:n
       end)
:(eval(Catlab.CSetDataStructures.init_acset(Graph, $(Expr(:copyast, :($(QuoteNode(quote
    #= REPL[3]:2 =#
    V = n
    #= REPL[3]:3 =#
    E = n
    #= REPL[3]:4 =#
    src = n
    #= REPL[3]:5 =#
    tgt = 1:n
end))))))))
```

but I would expect the macro call to expand to something like

```julia
let X = Graph()
  add_parts!(X, :V, n)
  add_parts!(X, :E, n)
  set_subpart!(X, :, :src, n)
  sut_subpart!(X, :, :tgt, 1:n)
  X
end
```

Then there would be no `eval` and hence it wouldn't matter what the scope of `n` is, right?

epatters:
Actually, why not just have macro expand to

```julia
make_acset(Graph; V = n, E = n, src = n, tgt = 1:n)
```

where `make_acset` is some helper function? Why are we generating code that calls `eval`?

jpfairbanks:
That would be much better

PR #463, opened 2021-07-12
@acset macro fix for runtime values

olynch:
Addresses #455


epatters (changes requested):
Thanks Owen, this looks good, but the tests are still failing. Also it looks like an indent width of 4 has crept into a few places.

epatters (approved):
Great, thanks!

PR #465, opened 2021-07-14
Generic interface for pushout complements

epatters:
This PR

- defines a generic interface for pushout complements (`pushout_complement` and `can_pushout_complement`)
- extracts an implementation of pushout complements in FinSet from pushout complements in C-Set
- refactors the latter to use the former


epatters:
@kris-brown, I've done some refactoring around pushout complements. The `DPO` module is now rather small but that's where the structured multicospan rewriting will eventually go :)

kris-brown (approved):

PR #466, opened 2021-07-15
Simplify implementation of @present macro

olynch:
This is refactored out of the struct acsets branch.

### DIFF
--- a/src/core/Present.jl
+++ b/src/core/Present.jl
@@ -12,7 +12,6 @@ export @present, Presentation, generator, generators, generator_index,
   has_generator, equations, add_generator!, add_generators!, add_definition!, add_equation!
 
 using Base.Meta: ParseError
-using Compat
 using MLStyle: @match
 
 using ..Meta, ..Syntax
@@ -205,7 +204,7 @@ end
 macro present(head, body)
   name, pres = @match head begin
     Expr(:call, name::Symbol, syntax_name::Symbol) =>
-      (name, :($(esc(:Presentation))($(esc(syntax_name)))))
+      (name, :($(GlobalRef(Present, :Presentation))($(esc(syntax_name)))))
     Expr(:(<:), name::Symbol, parent::Symbol) => (name,:(copy($(esc(parent)))))
     _ => throw(ParseError("Ill-formed presentation header $head"))
   end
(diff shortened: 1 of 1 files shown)


src/core/Present.jl
```diff
@@ -120,94 +120,98 @@ end
...
-""" Define a presentation using a convenient syntax.
+""" Shorthand for contructing a term in the presentation
 """
-macro present(head, body)
-  name, pres = @match head begin
-    Expr(:call, name::Symbol, syntax_name::Symbol) =>
-      (name, :($(module_ref(:Presentation))($syntax_name)))
-    Expr(:(<:), name::Symbol, parent::Symbol) => (name, :(copy($parent)))
-    _ => throw(ParseError("Ill-formed presentation header $head"))
+function (pres::Presentation)(e::Symbol)
```
epatters, line 125: Let's make this a function with a name rather than making presentation objects callable.

src/core/Present.jl
```diff
@@ -120,94 +120,98 @@ end
...
 
-""" Translate declaration of a generator.
+""" Create and add multiple generators
 """
-function translate_generator(name::Union{Symbol,Nothing}, type_expr)::Expr
-  type_name, args = @match type_expr begin
-    name::Symbol => (name, [])
-    Expr(:call, name::Symbol, args...) => (name, args)
-    _ => throw(ParseError("Ill-formed type expression $type_expr"))
+function add_generators!(pres::Presentation, names::Vector,
```
epatters, line 155: Do we want to overload `add_generator(s)!` with these extra methods? I feel like this is an implementation detail of the `@present` macro and should not be exposed to the user through these generic functions.

src/core/Present.jl
```diff
@@ -120,94 +120,98 @@ end
...
 
-""" Translate definition of generator in terms of other generators.
-"""
-function translate_definition(name::Symbol, def_expr)::Expr
-  Expr(:call, module_ref(:add_definition!), :_presentation,
-       QuoteNode(name), translate_expr(def_expr))
+# Presentation Definition
+#########################
+
+function add_to_presentation(pres, block)
```
epatters, line 165: Function name should end with `!`.

epatters (commented):
Thanks Owen!

src/core/Present.jl
olynch: Actually, it shouldn't mutate! I have no idea why this did mess *everything* up by mutating old presentations...

epatters (approved):

PR #467, opened 2021-07-20
Algebra of sub-C-sets of a C-set

epatters:
This PR defines a generic interface for subobjects in a category and implements the major operations for sub-C-sets: meet, join, top, bottom, implies, negate, subtract, and non.


epatters:
The core functions and tests are written but I may add some special applications to graphs and/or a vignette with graphical examples.

epatters:
Merging so that @cbw124 can start working on this stuff!

PR #470, opened 2021-07-29
Sketched C-sets

epatters:
This PR is a first stab at implementing C-sets with a finite limit sketch on top.


epatters:
I already knew that equalizers didn't yet work, but on further thought I think that a more significant design change is needed. I will follow up with some refactoring.

jpfairbanks:
I'm worried about the scalability of this approach in terms of precomputing and synchronizing the pullbacks. A lazy approach that allows you to iterate over the pullback without materializing it in memory seems more useful. Of course that depends on the usecase, but for example in the example of finite categories and the pullback for composition, the set of compatible morphisms is already potentially `O(n^2)` in the number of edges in the presentation. An efficient algorithm for computing the number of such compatible morphisms is given by the `sum(A^2)` where `A` is the adjacency matrix of the graph. A problematic case would be the `n,m` spider graph with `n+m+1` vertices and `n+m` edges where there are `m*n` compatible edges in the pullback for composition. 

If the explicit realization is easier to implement, then I guess we should it first, but I think that some consideration of what an API for using a lazy/implicit implementation would look like could save a lot of work.

PR #475, opened 2021-07-30
Restore support for non-vector iterators in `@acset` macro

epatters:
A side effect of #463 caused a test failure in AlgDynamics.


epatters:
@olynch, I will go ahead and merge this to fix downstream problems, but feel free to follow up if you have any thoughts.

PR #476, opened 2021-08-01
DOC: add a vignette for preorders

jpfairbanks:
This vignette shows some basic features of catlab through the lens of preorders. I made a folder for Sketches to put more 7 sketches related content in one place.


jpfairbanks:
Oh no, I think I just made a merge commit because I didn't use the rebase helper. @mehalter is there a way to fix it?

epatters (commented):
Thanks, this is nice!

I think we should mention that although Catlab doesn't let you create multiple generators with the same name, it does let you create unnamed generators. This is intended for things like preorders where you might want to specify the existence of inequalities without giving them all names.

docs/literate/sketches/preorders.jl
```diff
@@ -0,0 +1,270 @@
...
+using Catlab.Graphics.Graphviz
+using Catlab.Graphs
+using Catlab.CategoricalAlgebra
+using Catlab.CategoricalAlgebra.DataMigration
+using Test
+
+import Catlab.Theories: compose
+
+#=
+# Definition of a Preorder formalized as a GAT
```
epatters, line 26: We should note that all these theories are defined in the Catlab standard library (`Catlab.Theories`).

jpfairbanks:
> I think we should mention that although Catlab doesn't let you create multiple generators with the same name, it does let you create unnamed generators. This is intended for things like preorders where you might want to specify the existence of inequalities without giving them all names.

Oh that is literally what I wanted to specify!

epatters:
Haha, right. I used to use this for the Data Science Ontology, where I had to write down a lot of subtype relations.

PR #487, opened 2021-08-19
Different internal representations of subobjects

### LINKED ISSUE #472: Different internal representations of subobjects
As a design improvement following #467, we should replace the struct `Subobject` with an abstract type of the same name and then allow different concrete subtypes. This will be useful to avoid premature coercion to the standard representation, which can cause unnecessary allocations.

Specifically, for sub-C-sets of a C-set X, we should support at least three different representations:

- A monomorphism into X (the standard representation)
- A morphism from X into the subobject classifier (#473)
- For each object c in C, a function X(c) -> Bool (the poor man's subobject classifer, which is nevertheless very useful in calculations)

epatters:
Closes #472. This is a cleaner design and will result in fewer allocations when stringing together sub-C-set operations.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -686,10 +686,14 @@ Subobject(X::ACSet, components::NamedTuple) =
   SubACSetComponentwise(X, components)
 Subobject(X::ACSet; components...) = Subobject(X, (; components...))
 
-coerce_subob_component(X::FinSet, subset::SubFinSet) =
-  X == ob(subset) ? subset : error("Domain mismatch: $X != $(ob(subset))")
-coerce_subob_component(X::FinSet, f::FinFunction) =
-  X == codom(f) ? Subobject(f) : error("Domain mismatch: $X != $(codom(f))")
+function coerce_subob_component(X::FinSet, subset::SubFinSet)
+  X == ob(subset) ? subset :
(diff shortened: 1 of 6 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -661,31 +657,50 @@ end
...
+Subobject(X::ACSet, components::NamedTuple) =
+  SubACSetComponentwise(X, components)
 Subobject(X::ACSet; components...) = Subobject(X, (; components...))
 
-coerce_subob_component(f::FinFunction{Int}) = collect(f)
-coerce_subob_component(A::SubFinSet{Int}) = collect(hom(A))
-coerce_subob_component(f::AbstractVector{Int}) = f
-coerce_subob_component(pred::Union{AbstractVector{Bool},BitVector}) =
-  findall(pred)
+coerce_subob_component(X::FinSet, subset::SubFinSet) =
```
jpfairbanks, line 689: Probably worth documenting whether X is supposed to be the domain or codomain in these error messages. By that I mean, are we asserting that `X ⊆ codom(f)` or `dom(f) ⊆ X`? I was confused by it when reading and then referred to the Subobjects for finset for clarity and saw this

```
SubFinSet(X, f) = Subobject(FinFunction(f, X))
```
which made me even more confused.

PR #491, opened 2021-08-23
Fast Left-Kan Extensions Using The Chase implementation

kris-brown:
A direct implementation of the chase algorithm for left kan extensions from this paper https://www.categoricaldata.net/cql/fmc.pdf

This extends our current approach to left kan extensions by allowing us to use acyclic schemas; however, the procedure may fail to converge in some bounded number of iterations.


epatters:
Closing for now, for the reasons that we discussed.

PR #512, opened 2021-09-30
Pushout based chase

kris-brown:
This extends our current approach to left kan extensions by allowing us to use acyclic schemas; however, the procedure may fail to converge in some bounded number of iterations. Some effort would be needed to make this work with attributes.


kris-brown:
This content (with substantial changes) is now part of https://github.com/AlgebraicJulia/Catlab.jl/pull/609

PR #514, opened 2021-10-02
Structured multicospan rewriting

kris-brown:
Structured (multi)cospan rewriting extends DPO rewriting by allowing rules to be composed in parallel, not just in sequence. This should allow us to specify rewrites compositionally as well as build up rewrite rules inductively (given a fixed set of primitive rewrites). The interface is intended to be as similar as possible to DPO for ACSets.


epatters:
This functionality now belongs in [AlgebraicRewriting.jl](https://github.com/AlgebraicJulia/AlgebraicRewriting.jl).

PR #519, opened 2021-10-09
Loose acset morphisms and limits of acsets

epatters:
The motivation for this PR is taking limits of acsets such that the attribute types are included in the pointwise limits, e.g., taking a product of labeled graphs such that the product graph has labels which are tuples of labels from the original graphs. This construction *is* a limit but it's a limit in a category of acsets that differs from our usual one.

We usually work with acset morphisms that fix the attribute types. In this PR, these are called *tight* morphisms. They are appropriate for colimits of acsets. We then introduce *loose* acset morphisms, which are allowed to have arbitrary Julia functions between attribute types. This gives the notion of limit described above. The terminology of "tight" and "loose" is nonstandard but is adopted from the [nLab](https://ncatlab.org/nlab/show/M-category).

Despite these changes, the PR should be nonbreaking.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -239,6 +239,11 @@ function coerce_type_component(type::Symbol, f::SetFunction,
   eltype(codom(f)) <: codom_type || error("Codomain error in component $type")
   return f
 end
+function coerce_type_component(type::Symbol, ::Nothing,
+                               dom_type::Type, codom_type::Type)
+  codom_type == Nothing || error("Codomain error in component $type")
+  ConstantFunction(nothing, TypeSet(dom_type))
+end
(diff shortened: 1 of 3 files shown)


epatters:
I like your general point about carrying around redundant information. In some cases, we do the right thing, e.g., `FinFunctionVector`s only carry the codomain and construct the domain from the length of the vector as needed. However, in the context of acset, there would still be some redundancy, as you point out, because the codomain isn't needed either.. So it would be good to think about systematic approaches to this sort of thing.

I think the right way to approach cross-schema morphisms is through the category of diagrams (possibly upgraded from categories to schemas). That opens a larger question about how to interoperate acsets with a general interface for diagrams, something very relevant to my current work.

Issue #520, opened 2021-10-12
Bundling legs results in type error


bosonbaas:
I ran into an issue when bundling legs of ReactionNets in AlgebraicPetri, and can reproduce this error on an ACSet with multiple attributes. The code below reproduces the error:
```julia
using Catlab, Catlab.Theories, Catlab.Graphs, Catlab.CategoricalAlgebra
using Catlab.Graphs.BasicGraphs: AbstractGraph, TheoryGraph

@present TheoryMultiplyAttributedGraph <: TheoryGraph begin
  Length::AttrType
  Size::AttrType
  Label::AttrType
  vlabel::Attr(V,Label)
  elabel::Attr(E,Label)
  vsize::Attr(V, Size)
  elength::Attr(E, Length)
end

@acset_type MAGraph(TheoryMultiplyAttributedGraph, index=[:src,:tgt]) <: AbstractGraph
const OpenMAGraphOb, OpenMAGraph = OpenACSetTypes(MAGraph, :V)

g0 = @acset MAGraph{Int, Float64, Symbol} begin
  V = 5
  E = 3
  vlabel = [:a, :b, :c, :d, :e]
  elabel = [:a, :b, :c]
  vsize = [0.0, 1.0, 3.0, 3.0, 4.0]
  elength = [1,2,3]
  src = [1, 2, 3]
  tgt = [3, 4, 5]
end
g′ = OpenMAGraph{Int, Float64, Symbol}(g0, FinFunction([1],5), FinFunction([3],5), FinFunction([4],5))
bundle_legs(g′, [1, (2,3)])
```
The error is in the `colimit` tooling, and the first line is:
```
MethodError: Cannot `convert` an object of type Symbol to an object of type Float64
```
I think that I've found a fix to this bug, and will be publishing a related PR.

PR #521, opened 2021-10-12
Fix ordering in (co)domain number accessors for schemas

bosonbaas:
This PR provides a fix for Issue #520. The error was that the [`adom_num`](https://github.com/AlgebraicJulia/Catlab.jl/blob/master/src/theories/Schema.jl#L108) and [`acodom_num`](https://github.com/AlgebraicJulia/Catlab.jl/blob/master/src/theories/Schema.jl#L115) functions assumed a certain ordering on the `codoms` field. This is now explicitly aligned to the order of the `attrs` field.


epatters:
Thank you, Andrew, for reporting the bug and finding a fix! Following the struct acsets refector, I am less familiar than I was previously with this part of the codebase, but I thought that `doms` and `codoms` were named tuples and so I'm surprised that they don't have the correct order.

It also makes me wonder: do the functions `dom_nums` and `codom_nums` also have this subtle bug?

@olynch, it would be great if you could review carefully review this issue.

Issue #530, opened 2021-10-21
Removing box from DWD changes port order


bosonbaas:
The [`rem_box!`](https://github.com/AlgebraicJulia/Catlab.jl/blob/master/src/wiring_diagrams/Directed.jl#L457) function for DWDs can change the order of ports of other boxes.
This is likely because the DWD depends on the relative order of the Port parts to determine the order of box ports, but the [`rem_parts!`](https://github.com/AlgebraicJulia/Catlab.jl/blob/master/src/categorical_algebra/ACSetInterface.jl#L194) often does not preserve this order. A minimal example which demonstrates this is given below:
```julia
using Catlab
using Catlab.WiringDiagrams

A, B, C = [:A], [:B], [:C]
f = Box(:f, A, B)
g = Box(:g, [A,B], C)

d = WiringDiagram([A, B], [C,B])
fv = add_box!(d, f)
gv = add_box!(d, g)

add_wire!(d, (input_id(d),1) => (fv,1))
add_wire!(d, (fv,1) => (output_id(d),2))
add_wire!(d, (input_id(d),1) => (gv,1))
add_wire!(d, (input_id(d),2) => (gv,2))
add_wire!(d, (gv,1) => (output_id(d),1))

@show input_ports(d, 2)
rem_box!(d, 1)
@show input_ports(d, 1)
```
On my computer, this outputs:

input_ports(d, 2) = Any[[:A], [:B]]
input_ports(d, 1) = Any[[:B], [:A]]

bosonbaas:
I could see a few potential solutions for this:
* Option for rem_part! that preserves relative ordering of parts
* Specialized rem_part! implementation specifically for DWDs
* Storage of port order as an attribute on ports

Would love to hear any thoughts on which solution would be preferred.

PR #531, opened 2021-10-22
Shift Algorithm for ACSet Part Removal

bosonbaas:
This PR provides an implementation of the shift algorithm for removing ACSet parts. This resolves issue #530 and adds some options to part removal algorithms.

The shift algorithm simply shifts each part above the part to be removed down by one, and then removes the last part. This preserves the relative ordering of parts, which is important for data structures like Directed Wiring Diagrams which depend on relative order for box port order.


bosonbaas:
Just an update that this is still on my radar, I just ended up including a temporary shim downstream to get things working as deadlines are approaching. Still am planning to work on this (unless someone else gets to it first)

epatters:
Closing since this was never finished and is now quite stale.

PR #553, opened 2021-11-04
Add Literate graphs/graphs.jl

jpfairbanks:
Here are some introductory docs about CSets with the example of graphs.

PR #555, opened 2021-11-05
Vignette for graphs as C-sets

jpfairbanks:
replacement for #553 without being bad at git

### DIFF
--- a/docs/literate/graphs/graphs.jl
+++ b/docs/literate/graphs/graphs.jl
@@ -1,14 +1,10 @@
-using Catlab
+using Catlab, Catlab.Theories
 using Catlab.CategoricalAlgebra
 using Catlab.Graphs
-using Catlab.Graphs.BasicGraphs
 using Catlab.Graphics
 using Catlab.Graphics.Graphviz
-using Catlab.Graphics.GraphvizGraphs
-using Catlab.Graphs.PropertyGraphs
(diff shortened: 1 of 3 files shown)


docs/literate/graphs/graphs.jl
```diff
@@ -0,0 +1,325 @@
...
+using Catlab.Graphics.Graphviz
+#import Catlab.Graphics.Graphviz: to_graphviz, to_graphviz_property_graph
+
+using Colors
+draw(g) = to_graphviz(g, node_labels=true, edge_labels=true)
+
+GraphvizGraphs.to_graphviz(f::ACSetTransformation; kw...) =
+  to_graphviz(GraphvizGraphs.to_graphviz_property_graph(f; kw...))
+
+function GraphvizGraphs.to_graphviz_property_graph(f::ACSetTransformation; kw...)
```
epatters, line 14: This is nice. We should upstream something like it to `Catlab.Graphs`.

epatters (approved):
Thanks for the docs! I fixed the build problems. If you're satisfied, feel free to merge.

docs/literate/graphs/graphs.jl
jpfairbanks: Yeah for sure, I'll post it in an issue.

PR #563, opened 2021-11-10
DOC: adding vignette for WDs as C-Sets

jpfairbanks:
This vignette shows the differences between DWDs UWDs and CPGs with their schemas and associated DSLs.


epatters (approved):
Looks great, thanks!

Issue #558, opened 2021-11-07
Visualize graph homomorphisms using Graphviz


epatters:
The `Graphs` module has built-in support for Graphviz drawing of graphs. It should also support visualizing graph homomorphisms.

A starting point is in #555, which visualizes the codomain graph using colors.

jpfairbanks:
Here is the code from #555 that you can start with, the only changes I think it needs are 

```julia
function GraphvizGraphs.to_graphviz_property_graph(f::ACSetTransformation; kw...)
  pg = GraphvizGraphs.to_graphviz_property_graph(dom(f); kw...)
  vcolors = hex.(range(colorant"#0021A5", stop=colorant"#FA4616", length=nparts(codom(f), :V)))
  ecolors = hex.(range(colorant"#6C9AC3", stop=colorant"#E28F41", length=nparts(codom(f), :E)))
  hex.(colormap("Oranges", nparts(codom(f), :V)))
  for v in vertices(dom(f))
    fv = f[:V](v)
    set_vprops!(pg, v, Dict(:color => "#$(vcolors[fv])"))
  end
  for e in edges(dom(f))
    fe = f[:E](e)
    set_eprops!(pg, e, Dict(:color => "#$(ecolors[fe])"))
  end
  pg
end
```

1. Type parameters should be slightly more constrained so that we don't pirate ourselves. 
2. Color scale should be passed as a keyword argument.

@olynch, do you see a way to generalize this approach for Semagrams? Color coding a Semagram via a homomorphism seems useful.

slwu89:
@jpfairbanks I was playing around with this code, it has a problem when multiple edges/nodes get mapped to a single edge or node in the codomain. Minimal ex:

```julia
G = Graph(2)
add_edges!(G, [1], [2])

H = Graph(1)
add_edges!(H, [1], [1])

f = ACSetTransformation(G, H; V = [1,1], E = [1])

draw(f)
```

jpfairbanks:
Is the problem that multiple vertices are getting sent to 1 in the codomain?  Or is it that there is only one vertex in the codomain? I think the latter because we are doing a colorant range

```julia
  vcolors = hex.(range(colorant"#0021A5", stop=colorant"#FA4616", length=nparts(codom(f), :V)))
  ecolors = hex.(range(colorant"#6C9AC3", stop=colorant"#E28F41", length=nparts(codom(f), :E)))
```

[...]

PR #569, opened 2021-11-12
Arbitrarily color boxes / junctions of a wiring diagram

kris-brown:
This is one way we could allow users to color wiring diagrams. A dictionary mapping box IDs to strings (e.g. `"lightblue"`) can be passed to `to_graphviz`, with the default behavior being unchanged (unfilled boxes, black junctions).

### DIFF
--- a/src/graphics/GraphvizWiringDiagrams.jl
+++ b/src/graphics/GraphvizWiringDiagrams.jl
@@ -427,7 +427,6 @@ function to_graphviz_property_graph(d::UndirectedWiringDiagram;
   # Create nodes for boxes.
   box_vs = add_vertices!(graph, nboxes(d))
   set_vprop!(graph, box_vs, :id, [ "box$b" for b in boxes(d) ])
-  set_vprop!(graph, box_vs, :fillcolor, [ "green" for b in boxes(d) ])
   labels = if box_labels isa Symbol
     node_label.(subpart(d, box_vs, box_labels))
   else
--- a/src/graphics/GraphvizWiringDiagrams.jl
+++ b/src/graphics/GraphvizWiringDiagrams.jl
@@ -85,6 +85,7 @@ function to_graphviz(f::WiringDiagram;
     outer_ports::Bool=true, anchor_outer_ports::Bool=true,
     graph_attrs::AbstractDict=Dict(), node_attrs::AbstractDict=Dict(),
     title::Union{Nothing, Graphviz.Label}=nothing,
+    node_colors::AbstractDict=Dict(),
     edge_attrs::AbstractDict=Dict(), cell_attrs::AbstractDict=Dict())::Graphviz.Graph
   @assert label_attr in (:label, :xlabel, :headlabel, :taillabel)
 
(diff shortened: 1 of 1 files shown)


jpfairbanks:
It looks like #558 could use this too.

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -84,6 +84,7 @@ function to_graphviz(f::WiringDiagram;
     port_size::String="24", junction_size::String="0.05",
     outer_ports::Bool=true, anchor_outer_ports::Bool=true,
     graph_attrs::AbstractDict=Dict(), node_attrs::AbstractDict=Dict(),
+    node_colors::Dict{Int,String}=Dict{Int,String}(),
```
epatters, line 87: Would replace with ``node_colors::AbstractDict=Dict()`` to avoid unnecessarily strong typing.

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -219,7 +226,7 @@ function box_html_label(nin::Int, nout::Int, text_label::String;
     Graphviz.Html("""
       <TABLE BORDER="0" CELLPADDING="0" CELLSPACING="0">
       <TR><TD>$(orientation == TopToBottom ? input_label : output_label)</TD></TR>
-      <TR><TD $(html_attributes(attrs))>$(escape_html(text_label))</TD></TR>
+      <TR><TD $(html_attributes(attrs))>$((text_label))</TD></TR>
```
epatters, line 229: This change shouldn't be in this PR, as it addresses a separate issue and also I suspect it needs to be handled more carefully.

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -481,7 +489,7 @@ end
 
 """ Create a label for the main content of a box.
 """
-node_label(box_value) = box_label(MIME("text/plain"), box_value)
+node_label(box_value) = box_label(MIME("text/html"), box_value)
```
epatters, line 492: Likewise for this change.

epatters (changes requested):

src/graphics/GraphvizWiringDiagrams.jl
kris-brown: sorry - that hack wasn't supposed to leak into the PR!

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -416,6 +423,7 @@ function to_graphviz_property_graph(d::UndirectedWiringDiagram;
   # Create nodes for boxes.
   box_vs = add_vertices!(graph, nboxes(d))
   set_vprop!(graph, box_vs, :id, [ "box$b" for b in boxes(d) ])
+  set_vprop!(graph, box_vs, :fillcolor, [ "green" for b in boxes(d) ])
```
epatters, line 426: Sorry, one more thing. Was this change intended? I feel like the UWD boxes probably shouldn't have a color by default, but if they are colored, that color should at least be configurable.

epatters (commented):
Besides the question/comment below, it looks like you'll need to rebase due to conflict with your other PR, now merged.

src/graphics/GraphvizWiringDiagrams.jl
kris-brown: definitely unintended, thanks!

epatters (approved):

PR #570, opened 2021-11-12
Allow adding titles to wiring diagrams

kris-brown:
For visualizing a dynamic process (each timestep represented by a wiring diagram), I've found it's helpful to have a [title](https://stackoverflow.com/questions/6450765/how-do-you-center-a-title-for-a-diagram-output-to-svg-using-dot) that provides information during that timestep.  E.g.

Screen Shot 2021-11-12 at 12 28 33 PM

To do this, I added a `Title` as a new kind of `Graphviz.Statement` and added a `title::Union{Nothing,String}` keyword argument to `to_graphviz`.

### DIFF
--- a/src/graphics/Graphviz.jl
+++ b/src/graphics/Graphviz.jl
@@ -7,7 +7,7 @@ References:
 """
 module Graphviz
 export Expression, Statement, Attributes, Graph, Digraph, Subgraph,
-  Node, NodeID, Edge, pprint, run_graphviz
+  Node, NodeID, Edge, Title, pprint, run_graphviz
 
 using DataStructures: OrderedDict
 using AutoHashEquals
@@ -107,6 +107,10 @@ Edge(path::Vararg{NodeID}; attrs...) = Edge(collect(path), attrs)
 Edge(path::Vector{String}, attrs::AbstractDict) = Edge(map(NodeID, path), attrs)
 Edge(path::Vector{String}; attrs...) = Edge(map(NodeID, path), attrs)
 Edge(path::Vararg{String}; attrs...) = Edge(map(NodeID, collect(path)), attrs)
+
+@auto_hash_equals struct Title <: Statement
+  title::String
+end
 
 # Bindings
 ##########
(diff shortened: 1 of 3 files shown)


jpfairbanks:
Looks good to me. I like the kwag defaults to nothing so that we can identify explicit empty string vs no string passed.

jpfairbanks (approved):

src/graphics/Graphviz.jl
```diff
@@ -108,6 +108,10 @@ Edge(path::Vector{String}, attrs::AbstractDict) = Edge(map(NodeID, path), attrs)
 Edge(path::Vector{String}; attrs...) = Edge(map(NodeID, path), attrs)
 Edge(path::Vararg{String}; attrs...) = Edge(map(NodeID, collect(path)), attrs)
 
+@auto_hash_equals struct Title <: Statement
```
epatters, line 111: To adhere better to the Graphviz spec, which allows labels in subgraphs too, I would replace the `Title` struct with

```julia
Base.@kwdef struct Label <: Statement
  labelloc::String = ""
  label::String = ""
end
```
We can still use the keyword argument `title` below.

src/graphics/GraphvizWiringDiagrams.jl
```diff
@@ -84,11 +84,16 @@ function to_graphviz(f::WiringDiagram;
     port_size::String="24", junction_size::String="0.05",
     outer_ports::Bool=true, anchor_outer_ports::Bool=true,
     graph_attrs::AbstractDict=Dict(), node_attrs::AbstractDict=Dict(),
+    node_colors::Dict{Int,String}=Dict{Int,String}(),
```
epatters, line 87: This line looks like it snuck in from another PR.

epatters (changes requested):
Thanks Kris. In addition to the comment below, please add a unit test for the pretty-printer to `test/graphics/Graphviz.jl`.

src/graphics/Graphviz.jl
kris-brown: Graphviz [docs](https://graphviz.org/docs/attrs/labelloc/) say only "t" and "b" are allowed for graphs and clusters, so could `labelloc` instead be a bool `top` that defaults to `true`?

epatters: That's fine with me, but one reason to prefer the string is that nothing will break if Graphviz adds other allowed options for `labelloc`.

epatters (approved):

epatters:
Thanks Kris, looks good!

PR #572, opened 2021-11-13
Expose DPO morphism relating input and output of rewriting

kris-brown:
The current DPO interface does not allow users to access the intermediate morphisms involved in computing the rewrite. These can be important for applications that need to know how data in the original CSet relates to data in the rewritten CSet.

### DIFF
--- a/src/categorical_algebra/DPO.jl
+++ b/src/categorical_algebra/DPO.jl
@@ -11,16 +11,31 @@ using ..CSets: dangling_condition
 Apply a rewrite rule (given as a span, L<-I->R) to a ACSet
 using a match morphism `m` which indicates where to apply
 the rewrite.
+              l   r
+           L <- I -> R
+         m ↓    ↓    ↓
+           G <- K -> H
+
+Returns:
+- The morphisms I->K, K->G (produced by pushout complement), followed by
+  R->H, and K->H (produced by pushout)
 """
-function rewrite_match_maps(L::ACSetTransformation, R::ACSetTransformation,
-                         m::ACSetTransformation)::Vector{ACSetTransformation}
+function rewrite_match_maps(
+    L::ACSetTransformation, R::ACSetTransformation, m::ACSetTransformation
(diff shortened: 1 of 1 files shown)


src/categorical_algebra/DPO.jl
```diff
@@ -12,15 +12,19 @@ Apply a rewrite rule (given as a span, L<-I->R) to a ACSet
...
   codom(L) == dom(m) || error("Rewriting where L does not compose with m")
-  (k, _) = pushout_complement(L, m)
-  l1, _ = pushout(R, k)
-  return codom(l1)
+  (ik, kg) = pushout_complement(L, m)
+  rh, kh = pushout(R, ik)
+  return [ik, kg, rh, kh]
 end
 
+rewrite_match(L::ACSetTransformation, R::ACSetTransformation,
```
epatters, line 24: It looks like the main function `rewrite_match` no longer has a docstring and that the docstring now attached to `rewrite_match_maps` applies to the old function. All exported functions should have a docstring.

src/categorical_algebra/DPO.jl
```diff
@@ -12,15 +12,19 @@ Apply a rewrite rule (given as a span, L<-I->R) to a ACSet
...
+function rewrite_match_maps(L::ACSetTransformation, R::ACSetTransformation,
+                         m::ACSetTransformation)::Vector{ACSetTransformation}
   dom(L) == dom(R) || error("Rewriting where L, R do not share domain")
   codom(L) == dom(m) || error("Rewriting where L does not compose with m")
-  (k, _) = pushout_complement(L, m)
-  l1, _ = pushout(R, k)
-  return codom(l1)
+  (ik, kg) = pushout_complement(L, m)
+  rh, kh = pushout(R, ik)
+  return [ik, kg, rh, kh]
```
epatters, line 21: Should return tuple rather than vector. If the functions can be given meaningful names (not sure), it would be nice to return a named tuple.

epatters (changes requested):
Thanks for the PR!

PR #573, opened 2021-11-18
Initiality test for FinFunctors

kris-brown:
A property a FinFunctor may or may not have is being an initial functor (dual of [final functor](https://ncatlab.org/nlab/show/final+functor)). This can be straightforwardly computed for functors between free categories finitely presented by DAGs. For a functor of this sort F: S->T, we check whether, for each object t in Ob(T), that the comma category F/t is nonempty and connected.

This requires the path enumeration of DAGs.

### DIFF
--- a/src/categorical_algebra/FinCats.jl
+++ b/src/categorical_algebra/FinCats.jl
@@ -366,10 +366,13 @@ function is_initial(F::FinFunctor)::Bool
   pathₛ, pathₜ = enumerate_paths.([Gₛ, Gₜ])
 
   function connected_nonempty_slice(t::Int)::Bool
+    edges_into_t = incident(pathₜ, t, :tgt)
     # Generate slice objects
     ob_slice = Pair{Int,Vector{Int}}[] # s ∈Ob(S) and a path ∈ T(F(s), t)
     for s in vertices(Gₛ)
-        append!(ob_slice, [s => p for p in pathₜ[ob_map(F,s) => t]])
+      e = incident(pathₜ, ob_map(F,s), :src) ∩ edges_into_t
+      paths_s_to_t = isempty(e) ? [] : pathₜ[only(e), :eprops]
+      append!(ob_slice, [s => p for p in paths_s_to_t])
     end
 
     # Empty case
@@ -383,7 +386,9 @@ function is_initial(F::FinFunctor)::Bool
     """
     function check_pair(i::Int, j::Int)::Bool
       (m,pₘ), (n,pₙ) = ob_slice[i], ob_slice[j]
-      return any(f -> pₘ == vcat(edges.(hom_map(F,f))..., pₙ), pathₛ[m => n])
+      e = incident(pathₛ, m, :src) ∩ incident(pathₛ, n, :tgt)
+      paths = isempty(e) ? [] : pathₛ[only(e), :eprops]
+      return any(f -> pₘ == vcat(edges.(hom_map(F,f))..., pₙ), paths)
     end
 
     # Use check_pair to determine pairwise connectivity
--- a/src/categorical_algebra/FinCats.jl
+++ b/src/categorical_algebra/FinCats.jl
@@ -13,11 +13,12 @@ module FinCats
 export FinCat, FinCatGraph, Path, ob_generators, hom_generators, equations,
   is_discrete, is_free, graph, edges, src, tgt, presentation,
   FinFunctor, FinDomFunctor, is_functorial, collect_ob, collect_hom, force,
-  FinTransformation, components, is_natural
+  FinTransformation, components, is_natural, is_initial
 
 using AutoHashEquals
 using Reexport
 using StaticArrays: SVector
+using DataStructures: IntDisjointSets, in_same_set
 
 @reexport using ..Categories
 using ...GAT, ...Present, ...Syntax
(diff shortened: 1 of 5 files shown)


src/categorical_algebra/FinCats.jl
```diff
@@ -331,6 +332,47 @@ FinFunctor(ob_map, hom_map, dom::Presentation, codom::Presentation) =
 Categories.show_type_constructor(io::IO, ::Type{<:FinFunctor}) =
   print(io, "FinFunctor")
 
+function is_initial(F::FinFunctor)::Bool
```
epatters, line 335: Function should have a docstring. Would also be helpful to link to the nlab page for "final functor" since this concept is a bit obscure.

src/categorical_algebra/FinCats.jl
```diff
@@ -331,6 +332,47 @@ FinFunctor(ob_map, hom_map, dom::Presentation, codom::Presentation) =
...
+  Gₛ, Gₜ = F.dom.graph, F.codom.graph
```
epatters, line 336: An instance of the abstract type `FinFunctor` is not guaranteed to have `dom` and `codom` fields, e.g. an identity functor has no `codom` field, so you should use the accessor functions, which always work:

```julia
Gₛ, Gₜ = graph(dom(F)), graph(codom(F))
```

src/categorical_algebra/FinCats.jl
```diff
@@ -331,6 +332,47 @@ FinFunctor(ob_map, hom_map, dom::Presentation, codom::Presentation) =
...
+  pathₛ, pathₜ = enumerate_paths.([Gₛ, Gₜ])
+
+  function connected_nonempty_slice(t::Int)::Bool
+    # Generate slice objects
+    ob_slice = Pair{Int,Vector{Int}}[] # s ∈Ob(S) and a path ∈ T(F(s), t)
+    for s in vertices(Gₛ)
+        append!(ob_slice, [s => p for p in pathₜ[F.ob_map[s] => t]])
```
epatters, line 343: `F.ob_map[s]` -> `ob_map(F,s)`

src/categorical_algebra/FinCats.jl
```diff
@@ -331,6 +332,47 @@ FinFunctor(ob_map, hom_map, dom::Presentation, codom::Presentation) =
...
+    # Use check_pair to determine pairwise connectivity
+    connected = IntDisjointSets(length(ob_slice)) # sym/trans/refl closure
+    obs = 1:length(ob_slice)
+    for (i,j) in Base.Iterators.product(obs, obs)
+      if !in_same_set(connected, i, j) && check_pair(i,j)
+        union!(connected, i, j)
+      end
+    end
+
+    return connected.ngroups == 1
```
epatters, line 369: `connected.ngroups` -> `num_groups(connected)`

src/graphs/GraphAlgorithms.jl
```diff
@@ -116,4 +116,30 @@ function longest_paths(g::ACSet;
...
 end
 
+"""Enumerate all paths of an acyclic graph, indexed by src+tgt"""
+function enumerate_paths(G::Graph;
+                         sorted::Union{AbstractVector{Int},Nothing}=nothing
+                        )::DefaultDict{Pair{Int,Int},Set{Vector{Int}}}
+  sorted = isnothing(sorted) ? topological_sort(G) : sorted
+  Path = Vector{Int}
+  paths = [Set{Path}() for _ in 1:nv(G)] # paths that start on a particular V
+  for v in reverse(topological_sort(G))
```
epatters, line 126: Should be `reverse(sorted)`, right? Topological sort is already computed above.

src/graphs/GraphAlgorithms.jl
```diff
@@ -116,4 +116,30 @@ function longest_paths(g::ACSet;
...
+    push!(paths[v], Int[]) # add length 0 paths
+    for e in incident(G, v, :src)
+      push!(paths[v], [e]) # add length 1 paths
+      for p in paths[G[e, :tgt]] # add length >1 paths
+        push!(paths[v], vcat([e], p))
+      end
+    end
+  end
+  # Restructure `paths` into a data structure indexed by start AND end V
+  allpaths = DefaultDict{Pair{Int,Int},Set{Path}}(()->Set{Path}())
```
epatters, line 136: Not committed to this one, but I wonder if we should return an attributed reflexive graph, whose edges are paths in the original graphs with an attribute for the sequence forming the path?

test/categorical_algebra/FinCats.jl
```diff
@@ -175,4 +175,63 @@ G = FinDomFunctor(g)
 @test ob_map(G, :Weight) == TypeSet(Float64)
 @test hom_map(G, :weight) == FinDomFunction([0.5, 1.5])
 
+# Initiality of functors
```
epatters, line 178: Thank you for adding all these tests!

epatters (changes requested):
Thanks @kris-brown, I'm glad to have this feature!

src/graphs/GraphAlgorithms.jl
kris-brown: Yeah, I think that data structure better communicates what's going on! Unsure if creating a new attributed reflexive graph @theory is overkill for this, unless you can imagine us using that data structure in other places, too

PR #584, opened 2021-11-30
Colimits of finite sets with named elements

epatters:
A step towards the sort of thing we'd need for "lacsets" (cc @olynch).


epatters:
This could be improved, but I'm going to rapidly self-merge since we need this yesterday.

PR #588, opened 2021-12-30
Benchmarks for paper

olynch:
This contains updated benchmarks, along with code to generate figures for the paper. It also has a lot of miscellaneous optimizations; before this gets merged, I will look back and describe them.

### DIFF
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -386,6 +386,7 @@ function set_subpart_body(s::SchemaDesc, idxed::AbstractDict{Symbol,Bool},
         if subpart > 0
           insertsorted!(acs.hom_indices.$f[subpart], part)
         end
+        subpart
       end
     elseif unique_idxed[f]
       quote
@@ -397,11 +398,13 @@ function set_subpart_body(s::SchemaDesc, idxed::AbstractDict{Symbol,Bool},
         end
         acs.homs.$f[part] = subpart
         acs.hom_unique_indices.$f[subpart] = part
+        subpart
       end
     else
       quote
         @assert 0 <= subpart <= acs.obs[$(ob_num(s, s.codoms[f]))]
         acs.homs.$f[part] = subpart
+        subpart
       end
     end
   elseif f ∈ s.attrs
@@ -413,6 +416,7 @@ function set_subpart_body(s::SchemaDesc, idxed::AbstractDict{Symbol,Bool},
         end
         acs.attrs.$f[part] = subpart
         set_attr_index!(acs.attr_indices.$f, subpart, part)
+        subpart
       end
     elseif unique_idxed[f]
       quote
@@ -423,6 +427,7 @@ function set_subpart_body(s::SchemaDesc, idxed::AbstractDict{Symbol,Bool},
         end
         acs.attrs.$f[part] = subpart
         acs.attr_unique_indices.$f[subpart] = part
+        subpart
       end
     else
       :(acs.attrs.$f[part] = subpart)
--- a/src/categorical_algebra/CSetDataStructures.jl
+++ b/src/categorical_algebra/CSetDataStructures.jl
@@ -305,10 +305,15 @@ end
 # Mutators
 ##########
 
-@inline ACSetInterface.add_parts!(acs::StructACSet, ob::Symbol, n::Int) = _add_parts!(acs, Val{ob}, n)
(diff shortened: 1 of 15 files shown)


epatters (changes requested):
Looks like there are a few test failures. @olynch, can you fix them so we can get these performance improvements merged?

src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -299,18 +300,23 @@ function incident_body(s::SchemaDesc, idxed::AbstractDict{Symbol,Bool},
...
+                              part, ::Type{Val{f}}, ::Type{Val{copy}}) where
+  {S,Ts,Idxed,UniqueIdxed,f,copy}
+  incident_body(SchemaDesc(S),pairs(Idxed),pairs(UniqueIdxed),f,copy)
 end
 
 # Mutators
 ##########
 
-@inline ACSetInterface.add_parts!(acs::StructACSet, ob::Symbol, n::Int) = _add_parts!(acs, Val{ob}, n)
+@inline add_parts_with_indices!(acs::StructACSet, ob::Symbol, n::Int, index_sizes::NamedTuple) =
```
epatters, line 311: Is this function now part of the public interface? If so, it should have tests and documentation.

src/graphs/BasicGraphs.jl
```diff
@@ -119,17 +133,24 @@ add_vertex!(g::HasVertices; kw...) = add_part!(g, :V; kw...)
 """
 add_vertices!(g::HasVertices, n::Int; kw...) = add_parts!(g, :V, n; kw...)
 
+""" Add vertices with preallocated src/tgt indexes
+"""
+function add_vertices_with_indices!(g::HasVertices, n::Int, k::Int; kw...)
```
epatters, line 138: Are there unit tests for this exported function? Warning from CodeCov below suggests now.

src/graphs/BasicGraphs.jl
```diff
@@ -101,15 +102,28 @@ edges(g::HasGraph) = parts(g, :E)
...
 """
 has_vertex(g::HasVertices, v) = has_part(g, :V, v)
 
 """ Whether the graph has the given edge, or an edge between two vertices.
 """
-has_edge(g::HasGraph, e) = has_part(g, :E, e)
-has_edge(g::HasGraph, src::Int, tgt::Int) =
-  has_vertex(g, src) && tgt ∈ outneighbors(g, src)
+@inline has_edge(g::HasGraph, e) = has_part(g, :E, e)
+@inline function has_edge(g::HasGraph, s::Int, t::Int)
```
epatters, line 120: Are these `@inline`s necessary? Based on my experience, I would guess no, but I'm not completely.

Also, while it may be beyond the scope of this PR, I think we should modify `outneighbors` and co to return iterators, which should avoid the need for longer function defintiion below.

src/graphs/BasicGraphs.jl
```diff
@@ -173,21 +194,23 @@ distinction is moot.
...
 """
-outneighbors(g::AbstractGraph, v::Int) = subpart(g, incident(g, v, :src), :tgt)
+@inline outneighbors(g::AbstractGraph, v::Int) = @inbounds subpart(g, incident(g, v, :src), :tgt)
 
 """ Union of in-neighbors and out-neighbors in a graph.
 """
 all_neighbors(g::AbstractGraph, v::Int) =
   Iterators.flatten((inneighbors(g, v), outneighbors(g, v)))
 
+degree(g,v) = length(incident(g,v,:tgt)) + length(incident(g,v,:src))
```
epatters, line 212: Undocumented and untested.

src/graphs/GraphGenerators.jl
```diff
@@ -1,9 +1,11 @@
 module GraphGenerators
 export path_graph, cycle_graph, complete_graph, star_graph, wheel_graph,
-  parallel_arrows
+  parallel_arrows, erdos_renyi, expected_degree_graph, watts_strogatz
```
epatters, line 3: Are all these new graph generators untested?

src/graphs/Searching.jl
```diff
@@ -0,0 +1,125 @@
+module Searching
```
epatters, line 1: Same question: does this module have tests?

epatters (changes requested):
Thanks Owen for this painstaking work on benchmarks!

Two important things should be resolved before merging:

1. It looks like lots of new functions don't have unit tests. See inline comments for some of them.
2. The use of `@inline`s seems a bit arbitrary, although I could be missing something. In my experience, inlines are only needed to ensure constant propagation for symbols. Can we be more consistent about the use of inlines and remove any extraneous ones?

PR #591, opened 2022-01-06
Slice categories

kris-brown:
This PR is starting off with just slices, slice transformations, and limits in a slice category. It is parameterized by the Ob and Hom of some category. 

Once this part gets into shape, it should be straightforward to do colimits of slices as well as limits+colimits of co-slices.

### DIFF
--- a/src/categorical_algebra/Slices.jl
+++ b/src/categorical_algebra/Slices.jl
@@ -1,0 +1,92 @@
+module Slices
+export Slice, SliceMorphism,SliceLimit
+
+using ...GAT
+using ..FreeDiagrams, ..Limits, ..CSets
+using ...Theories: Category
+import ...Theories: dom, codom, compose, id
+import ..Limits: limit, universal
+import ..FinSets: force
+
+struct Slice{Ob,Hom}
+    slice::Hom
+end
+
+struct SliceMorphism{Ob, Hom}
+  dom::Slice{Ob, Hom}
+  codom::Slice{Ob, Hom}
+  f::Hom
+  function SliceMorphism{Ob,Hom}(dom::Slice{Ob, Hom},codom::Slice{Ob, Hom},
+                                 f::Hom) where {Ob, Hom}
+    p1, p2 = force.([dom.slice, compose(f,codom.slice)])
+    force(p1) == force(p2) ||
+       error("Slice condition failed: \ndom.slice $p1\ncomposite $p2\nf " *
+             "$(force(f))\ncodom $(force(codom.slice))")
+    return new(dom, codom, f)
+  end
+end
+
+
+dom(s::Slice)::StructACSet = dom(s.slice)
+codom(s::Slice)::StructACSet = codom(s.slice)
+force(s::Slice{Ob,Hom}) where {Ob,Hom} = Slice{Ob,Hom}(force(s.slice))
+force(s::SliceMorphism{Ob,Hom}) where {Ob,Hom} = SliceMorphism{Ob,Hom}(
+  force(dom(s)), force(codom(s)), force(s.f))
+
+
+struct SliceLimit{Ob_, Hom, Ob <: Slice{Ob_,Hom}, Diagram,
+                  Cone <: Multispan{Ob}} <: AbstractLimit{Ob,Diagram}
+  diagram::Diagram
+  cone::Cone
+  underlying::AbstractLimit
+end
+
+
+@instance Category{Slice, SliceMorphism} begin
+  dom(f::SliceMorphism) = f.dom
+  codom(f::SliceMorphism) = f.codom
+  id(A::Slice) = SliceMorphism(A, A, id(dom(A.slice)))
+  function compose(f::SliceMorphism, g::SliceMorphism)
+    codom(f) == dom(g) ||
+      error("Domain mismatch in composition: $(codom(f)) != $(dom(g))")
+    typeof(f)(dom(f), codom(g), compose(f.f, g.f))
+  end
+end
+
+"""
+Get the underlying diagram of a slice category diagram which has the object
+being sliced over explicitly, and arrows are ACSetTransformations
+
+Encode the base of slice as the first object in the diagram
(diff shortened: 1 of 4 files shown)


src/categorical_algebra/Slices.jl
```diff
@@ -0,0 +1,133 @@
...
+@auto_hash_equals struct Slice{Ob,Hom}
+  slice::Hom
+end
+
+@auto_hash_equals struct SliceMorphism{Ob, Hom}
+  dom::Slice{Ob, Hom}
+  codom::Slice{Ob, Hom}
+  f::Hom
+  function SliceMorphism{Ob, Hom}(dom, codom, f) where {Ob, Hom}
+    p1, p2 = force.([dom.slice, compose(f,codom.slice)])
```
epatters, line 21: Since the slice construction should work for any category, and not all category `@instances` in Catlab support or know about the `force` function, this constructor won't work in general.

There is a deeper issue here, which is that in Catlab we haven't taken a consistent position on whether `==` means mathematical equality or some intermediate notion of "data structure equality," which is weaker than `===` but sometimes stricter than mathematical equality. Often it actually it means the latter. 

We are not going to solve that problem here. For now, I'll just suggest that we follow the somewhat clunky pattern used [here](https://github.com/AlgebraicJulia/Catlab.jl/blob/35cb13bec84ea3137701948c307ecf325fb56681/src/categorical_algebra/Categories.jl#L243) and elsewhere of having keyword argument `strict` that controls whether the equality condition is checked. A side benefit is that when you call the constructor in `compose` below you can set `strict=false` and avoid a superfluous equality check. (That the check is not needed is what it means for composition in a slice category to be well-defined.)

src/categorical_algebra/Slices.jl
```diff
@@ -0,0 +1,133 @@
...
+    return new{Ob,Hom}(dom, codom, f)
+  end
+end
+
+SliceMorphism(a,b,f) = SliceMorphism{typeof(a).parameters...}(a,b,f)
+
+
+dom(s::Slice)::StructACSet = dom(s.slice)
+codom(s::Slice)::StructACSet = codom(s.slice)
+force(s::Slice{Ob,Hom}) where {Ob,Hom} = Slice{Ob,Hom}(force(s.slice))
```
epatters, line 34: Since calling `force` can change the concrete type of its argument (e.g., from `SetFunctionCallable` to `FinDomFunctionVector`), this should be simply:

```julia
force(s::Slice) = Slice(force(s.slice))
```

src/categorical_algebra/Slices.jl
```diff
@@ -0,0 +1,133 @@
...
+using ...Theories: Category
+import ...Theories: dom, codom, compose, id
+import ..Limits: limit, colimit, universal
+import ..FinSets: force, pushout_complement
+
+@auto_hash_equals struct Slice{Ob,Hom}
+  slice::Hom
+end
+
+@auto_hash_equals struct SliceMorphism{Ob, Hom}
```
epatters, line 16: Likewise about the type parameter `Ob` here. Furthermore, because the concrete morphisms types may differ, I think the struct definition should be:

```julia
struct SliceMorphism{Hom, Dom<:Slice, Codom<:Slice}
  dom::Dom
  codom::Codom
  f::Hom
end
```

Also, I think we should rename `SliceMorphism` to `SliceHom`, not for any particularly good reason, but just for consistency with other names in Catlab.

src/categorical_algebra/Slices.jl
```diff
@@ -0,0 +1,133 @@
+module Slices
+export Slice, SliceMorphism,SliceLimit,SliceColimit, pushout_complement
```
epatters, line 2: Since they are exported, the structs `Slice` and `SliceMorphism` should have docstrings.

Also, my guess is that `SliceLimit` and `SliceColimit` are implementation details that should not be exported. The philosophy here is that everything one can do with a (co)limit object is encapsulated by the accessor functions (`apex`, `legs`, etc) and the universal property (`universal`).

src/categorical_algebra/Slices.jl
```diff
@@ -0,0 +1,133 @@
...
+             "$(force(f))\ncodom $(force(codom.slice))")
+    return new{Ob,Hom}(dom, codom, f)
+  end
+end
+
+SliceMorphism(a,b,f) = SliceMorphism{typeof(a).parameters...}(a,b,f)
+
+
+dom(s::Slice)::StructACSet = dom(s.slice)
+codom(s::Slice)::StructACSet = codom(s.slice)
```
epatters, line 33: Return type here is not necessarily `StructACSet`.

src/categorical_algebra/Slices.jl
```diff
@@ -0,0 +1,133 @@
...
+
+using AutoHashEquals
+using ...GAT
+using ..FreeDiagrams, ..Limits, ..CSets
+using ...Theories: Category
+import ...Theories: dom, codom, compose, id
+import ..Limits: limit, colimit, universal
+import ..FinSets: force, pushout_complement
+
+@auto_hash_equals struct Slice{Ob,Hom}
```
epatters, line 12: From the PoV of dependent types, I understand the temptation to carry around the extra type parameter `Ob`, but ultimately I think we shouldn't because it's not in line with Julia's philosophy on types: it's not needed for dispatch or struct definition, and it's just extra stuff that has to be manually supplied and carried around.

Then you should also add a second constructor:

```julia
Slice(f::Hom) where Hom = Slice{Hom}(f)
```

epatters (changes requested):
Thanks Kris, generic support for slice categories is a long overdue feature!

I left a first round of feedback below. Happy to discuss further if it doesn't make sense.

epatters:
Sigh, GitHub seems to have nonsensically ordered my comments. It will make more sense if you read them in the order of the code.

kris-brown:
Feedback made sense and the code looks much better!

epatters (approved):

PR #592, opened 2022-01-06
ENH: Support for attributes in CatEl

bosonbaas:
This PR extends the current Category of Elements tooling to include Attributes. When generating the presentation for an attributed Category of Elements, it currently treats attributes as a special case and carries attributes along with their objects in the base instance.


jpfairbanks:
This sounds like the right approach to me.

PR #599, opened 2022-02-10
Add debugging info to is_functorial

### LINKED ISSUE #598: Improved error message on finfunctor failure
When using the `@finfunctor` macro, ill-defining a non-functorial functor yields the error message "Parsed functor is not functorial", along with the body of the functor.

This error message could be improved by stating which equations are not satisfied. Specifically, this would involve adding an optional `debug=false` flag to `parse_functor`.

lukem12345:
This pull request would close #598

It should be noted that the docstring for `is_functorial` states that the default value for `check_equations` is `true`, but the default value is actually `false`. As I am not sure which behavior is intended, I have left this discrepancy as is.

### DIFF
--- a/src/categorical_algebra/FinCats.jl
+++ b/src/categorical_algebra/FinCats.jl
@@ -323,20 +323,37 @@ not be possible.
 
 See also: [`is_natural`](@ref).
 """
-function is_functorial(F::FinDomFunctor; check_equations::Bool=false)
+function is_functorial(F::FinDomFunctor; check_equations::Bool=false, debug::Bool=false)
+  functorial = true
   C, D = dom(F), codom(F)
-  all(hom_generators(C)) do f
+  failed_hom_generators = filter(hom_generators(C)) do f
     g = hom_map(F, f)
-    dom(D,g) == ob_map(F, dom(C,f)) && codom(D,g) == ob_map(F, codom(C,f))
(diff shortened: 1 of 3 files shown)


src/categorical_algebra/FinCats.jl
```diff
@@ -323,20 +323,37 @@ not be possible.
 
 See also: [`is_natural`](@ref).
 """
-function is_functorial(F::FinDomFunctor; check_equations::Bool=false)
+function is_functorial(F::FinDomFunctor; check_equations::Bool=false, debug::Bool=false)
+  functorial = true
   C, D = dom(F), codom(F)
-  all(hom_generators(C)) do f
+  failed_hom_generators = filter(hom_generators(C)) do f
```
epatters, line 329: Should use `Iterators.filter`: no need to allocate when `debug=false`.

epatters (changes requested):
Thanks Luke! Besides the inline comment, I suggest the following changes:

- Rename `debug` to something like `return_failures`
- When that flag it set, return a list of generators failing the functorality or naturality check; don't print anything
- Add test cases checking that those lists are correct in cases were functorality/naturality fails

These changes should shorten the code, while also allowing the results to be used programmatically.

epatters:
Replaced by #762.

PR #602, opened 2022-02-11
surjective homomorphism constraint

kris-brown:
Just as we have a monic constraint option, the epic constraint keyword can be used to componentwise prune partial homomorphisms  in the search process that have no chance of becoming an epic homomorphism. 

By keeping track of how many elements have yet to be assigned in the domain and a vector for each element in the codomain (how many elements map to each element), we can check this.  When assigning a new element, we decrement the domain unassigned counter and we increment the counter on the element in the codomain (vice-versa when unassigning an element).


kris-brown:
Resubmitted as https://github.com/AlgebraicJulia/Catlab.jl/pull/910

PR #603, opened 2022-02-11
Loose acset homomorphism search

kris-brown:
Homomorphism searching by default assumes that attributes are related by an identity function. We can still have an ACSet homomorphism with a non-identity function, but it's not possible to determine this purely from the domain and codomain ACSets (we would at best only know the function's values for some finite set of points). However, if the user provides the Attr component of the homomorphism as starting data, we can check whether or not it is satisfied while performing the homomorphism search.

This returns a LooseACSetTransformation iff any non-Base.identity function is provided (note: some user defined x->x function is not equal to Base.identity).


src/categorical_algebra/CSets.jl
```diff
@@ -550,10 +553,12 @@ struct BacktrackingState{S <: SchemaDescType,
   dom::Dom
   """ Codomain ACSet: the "values" in the CSP. """
   codom::Codom
+  loose::LooseFun
 end
 
 function backtracking_search(f, X::StructACSet{S}, Y::StructACSet{S};
-                             monic=false, iso=false, initial=(;)) where {Ob, S<:SchemaDescType{Ob}}
+                             monic=false, iso=false, loose=(;), initial=(;),
```
epatters, line 560: I think we should rename the kw arg `loose` to `type_components`, since that's used elsewhere in this file.

src/categorical_algebra/CSets.jl
```diff
@@ -588,12 +596,22 @@ function backtracking_search(f, X::StructACSet{S}, Y::StructACSet{S};
...
 end
 
-function backtracking_search(f, state::BacktrackingState, depth::Int)
+function backtracking_search(f, state::BacktrackingState{S}, depth::Int) where {S}
   # Choose the next unassigned element.
   mrv, mrv_elem = find_mrv_elem(state, depth)
   if isnothing(mrv_elem)
     # No unassigned elements remain, so we have a complete assignment.
-    return f(ACSetTransformation(state.assignment, state.dom, state.codom))
+    if !isempty(state.loose) && any([f!=identity for f in state.loose])
```
epatters, line 604: Can replace both clauses of conditional with `any(!=(identity), state.loose)`.

src/categorical_algebra/CSets.jl
```diff
@@ -588,12 +596,22 @@ function backtracking_search(f, X::StructACSet{S}, Y::StructACSet{S};
...
+      ac = NamedTuple([
```
epatters, line 605: Using [this constructor](https://github.com/AlgebraicJulia/Catlab.jl/blob/12a6dbed89186757bb0168f3d1f8d662cd7b5c2c/src/categorical_algebra/CSets.jl#L324), can replace everything inside the conditional with

```julia
f(LooseACSetTransformation{S}(state.assignment, state.loose, state.dom, state.codom)
```

src/categorical_algebra/CSets.jl
```diff
@@ -659,8 +677,10 @@ be mutated even when the assignment fails.
 
     # Check attributes first to fail as quickly as possible.
     X, Y = state.dom, state.codom
-    $(map(out_attr(S, c)) do f
```
epatters, line 662: The accessor `out_attr` is not used elsewhere, so either update it or remove it.

epatters (changes requested):
Thanks Kris, this is a nice feature to have!

kris-brown:
Thanks, the code looks much nicer!

PR #609, opened 2022-03-18
(co)products/(co)equalizers for Fincats and Diagrams

kris-brown:
Main goal: implement the constructions described in [Peschke and Tholen](https://arxiv.org/abs/2006.10890) (2020) remark 2.8. This involves developing a lot of other prerequisite functionality.

Main contributions:
- Currying FinDomFunctors into C-Set as FinDomFunctors into Set
- pushout-based chase
    - Collage of functor
    - Conversion between C-Set and SpanC-Set (a model for C-Rel)
    - left kan extensions of FinDomFunctors into Set (or C-Set) + universal property
- Limits/colimits the category of FinCats
    - Products and coproducts of FinCat presentations (including treatment of equations)
    - No treatment of equations yet for equalizers/coequalizers
- Limits/colimits the category of FinCats
    - Products/coproducts of Diagram{id} and Diagram{op}, equalizers for the former and coequalizers for the latter
    - Universal properties for the above (co)limits
    - Left kan extension used to compute coequalizers for Diagram{id}

To-do:
- Left kan extension to compute equalizers for Diagram{op}
- Universal property of coequalizers (id) / equalizers (op). 

Caveats:
- Code for (co)limits of FinCats is stored in `Limits.jl` to avoid circular dependency. Tests are still in `test/.../FinCats.jl`
- For simplicity, FinCatPresentations are treated (but, ideally, code generically would work with FinCatGraphs as well)
- Because of the makeshift way dynamic acsets are implemented, rather than `MyCSetType()` it's safer to use `Base.invokelatest(MyCSetType)`
- Rebasing on master included a bunch of unrelated commits for some reason. The relevant files are `Chase.jl`, `Limits.jl`, `CSets.jl`, `Diagrams.jl`.
- Left kan extensions are abysmally slow (possibly in large part due to the dynamic acset hack)


kris-brown:
cleaned up commit history and resubmit as https://github.com/AlgebraicJulia/Catlab.jl/pull/615

PR #610, opened 2022-03-28
add hash method for diagram/diagram hom

kris-brown:
Diagrams/diagram homs do not use auto hash equals, so this must be defined manually.


epatters:
Thanks Kris!

PR #611, opened 2022-04-01
Replace half-baked function `roottype` with `Base.typename`

epatters:
It's still feels a bit hacky to have to do this at all, but this PR is definitely an improvement over what we had before. The function `Base.typename` is available in recent versions of Julia.


epatters:
This can also be considered a bug fix, since `Base.typename` handles `Union` types correctly, while the old implementation did not.

PR #614, opened 2022-04-12
Migrate from LightGraphs.jl to Graphs.jl

### LINKED ISSUE #613: Migrate graph interop from LightGraphs to Graphs
Due to a change of maintainership, [LightGraphs.jl](https://github.com/sbromberger/LightGraphs.jl) is EOL and [Graphs.jl](https://github.com/JuliaGraphs/Graphs.jl) is revived as the standard Julia graphs package.

epatters:
Closes #613. Also cleans up an addition to the Graphs interop API from #588.

Since (Light)Graphs is an *optional* dependency, it's arguable whether this change should be considered breaking. I'm going to argue that it's not just because I don't it merits a major release.


epatters:
The benchmarks CI failure is only due to the old benchmarks not running because they required package (LightGraphs) is not available!

PR #619, opened 2022-04-19
Homomorphism finding benchmarks

jhorn00:
I have a few supporting files that go along with the benchmark file, feel free to critique those as well. Let me know what needs to be changed and I'll get on it as soon as I can. I'm aware some of the functions have rather unsightly function call blocks in them. If this is a concern, let me know what you would prefer instead.

For when things are cleaned up, should I add anything to the benchmarks.jl file?

### DIFF
--- a/benchmark/Support/SSets.jl
+++ b/benchmark/Support/SSets.jl
@@ -142,11 +142,12 @@ quad_int = @acset SSet begin
 end
 
 function elements(f::ACSetTransformation{S}) where {S}
-  offY = Dict([o => let z = findfirst(==(i), elements(codom(f))[:πₑ])
+  X, Y = elements(dom(f)), elements(codom(f))
+  offY = Dict([o => let z = findfirst(==(i), Y[:πₑ])
     isnothing(z) ? 0 : z - 1
   end
                for (i, o) in enumerate(ob(S))])
   pts = vcat([collect(f[o]) .+ offY[o] for o in ob(S)]...)
-  hs = homomorphisms(elements(dom(f)), elements(codom(f)); initial=Dict([:El => pts]))
+  hs = homomorphisms(X, Y; initial=Dict([:El => pts]))
   return only(hs)
 end
--- a/benchmark/Support/SSets.jl
+++ b/benchmark/Support/SSets.jl
@@ -86,8 +86,6 @@ end
 function repeat1d(n::Int)
   res = @acset SSet begin
     V = 2 * n
-    #T = 2*(n-1)
-    #E = 4*(n-1)+1
   end
 
   # Create horizontal edges
--- a/benchmark/Support/SSets.jl
+++ b/benchmark/Support/SSets.jl
@@ -1,0 +1,159 @@
+# Authors: Kris Brown, Tyler Hanks
+# Original Repo: https://github.com/kris-brown/Computational-Category-Theoretic-Rewriting/blob/main/src/SSets.jl
+using Catlab.Graphs
+using Catlab.CategoricalAlgebra
+using Catlab.Present
+import Catlab.CategoricalAlgebra: elements
+
+# Define C-set
+@present ThSemisimplicialSet(FreeSchema) begin
+  (V, E, T)::Ob
+  (d1, d2, d3)::Hom(T, E)
+  (src, tgt)::Hom(E, V)
+end
+@acset_type SSet(ThSemisimplicialSet)
+
+# Define pattern
+"""
+V1--e1-→V2
+| t1  / |
+e2  /   e4
+↓ ↙  t2 ↓
+V3--e3-→V4
+"""
+quad = @acset SSet begin
+  T = 2
+  E = 5
+  V = 4
+  d1 = [1, 5]
+  d2 = [5, 3]
+  d3 = [2, 4]
+  src = [1, 1, 3, 2, 2]
+  tgt = [2, 3, 4, 4, 3]
+end
+
+# Define mesh
+"""
+Box order
+...
+n n+1 n+2 ...
+1  2   3 ...
+
+Edge order
+
+
+(i+1,j)  ------>(i+1,j+1)
+  |         /     |
+  v       /       v
+(i,j)----------> (i, j+1)
+"""
+function repeat(n::Int)
+  res = @acset SSet begin
+    E = (n * (n - 1)) * 2 + (n - 1)^2
+    V = n^2
+  end
+  coord = collect(Iterators.product(1:n, 1:n))
+  v_d = Dict([(x, y) => i for (i, (x, y)) in enumerate(coord)])
+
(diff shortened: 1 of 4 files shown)


epatters:
Hi James, thank you for the benchmarks! I'll to get someone (possibly me) to review the PR soon.

benchmark/Support/SSets.jl
```diff
@@ -0,0 +1,159 @@
...
+  end
+
+  return res
+
+end
+
+function repeat1d(n::Int)
+  res = @acset SSet begin
+    V = 2 * n
+    #T = 2*(n-1)
```
jpfairbanks, line 89: Do you want these triangles in here? Remove comment if not necessary.

benchmark/Support/autoplot.jl
```diff
@@ -0,0 +1,261 @@
...
+        print("Surjection complete.\n")
+    end
+    scatter([x1, x2], [y1, y2], title="Autoplotted Graph Vertices", xlabel="Number of \"From\" Vertices", ylabel="Single Hom Calculation Time (ns)", label=["Injective" "Surjective"])
+    savefig("autoVertex.png")
+    scatter([x3, x4], [y1, y2], title="Autoplotted Graph Edges", xlabel="Number of Edges", ylabel="Single Hom Calculation Time (ns)", label=["Injective" "Surjective"])
+    savefig("autoEdge.png")
+end
+
+# Less accurate... Much faster.
+function quickPlot(fromList, toList)
```
jpfairbanks, line 54: Is this faster because you only look at the zipped list of graphs to benchmark, rather than the Cartesian product of the lists? I would rather see this refractored to use `Base.Iterators.zip` and `Base.Iterators.product` and have an interface for just taking a sequence of pairs of graphs and running the benchmarks and making the scatter plot. Then the sequence of pairs of graphs could be created with different Iterators

benchmark/Support/autoplot.jl
```diff
@@ -0,0 +1,261 @@
...
+            append!(fromGraphs, [fromLarger])
+            append!(toGraphs, [toLarger])
+            fromLarger = apex(product(fromBase, add_loops(fromLarger)))
+            toLarger = apex(product(toBase, add_loops(toLarger)))
+        end
+    end
+end
+
+function autoShuffle(fromGraphs, toGraphs, vertexLimit, edgeLimit)
+    if vertexLimit < 1 && edgeLimit < 1
```
jpfairbanks, line 126: The Julia idiom for this is 

```
vertexLimit >= 1 && edgeLimit >= 1 || error("error message")
```

benchmark/Support/graph_bank.jl
```diff
@@ -0,0 +1,262 @@
+using Catlab, Catlab.Theories
+using Catlab.CategoricalAlgebra
+using Catlab.Graphs
+using Catlab.Graphs.GraphGenerators
+include("SSets.jl")
+
+# Graph Functions
+# add_loops
+add_loops!(g) = add_parts!(g, :E, nparts(g, :V), src=parts(g, :V), tgt=parts(g, :V))
```
jpfairbanks, line 9: This could be upstreamed into Catlab.Graphs. I would add an argument like `add_loops!(g, vs)` to allow people to only add loops for the vertices listed in vs.

benchmark/Support/autoplot.jl
jhorn00: For some reason I had forgotten those functions were still in that file. I need to adjust what I have in there since the plotting functions were used in the old plots I made toward the beginning of the semester.

That being said, I will work on them.

benchmark/Support/SSets.jl
```diff
@@ -0,0 +1,159 @@
...
+# Interface
+quad_int = @acset SSet begin
+  E = 4
+  V = 4
+  src = [1, 1, 3, 2]
+  tgt = [2, 3, 4, 4]
+end
+
+function elements(f::ACSetTransformation{S}) where {S}
+  X, Y = elements.([dom(f), codom(f)])
```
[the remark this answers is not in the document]
jhorn00: Definitely an oversight leaving that file in the state it was in. Honestly I just didn't review that file like I should have. I'll begin fixing the issues tomorrow (4/23 to clarify since I'm posting this late at night).

benchmark/Support/autoplot.jl
```diff
@@ -0,0 +1,261 @@
...
+    x2 = Int64[]
+    y2 = Float64[]
+    x3 = Int64[]
+    x4 = Int64[]
+    for i in 1:length(fromList)
+        print("Graph pair $i:   ")
+        f = fromList[i]
+        t = toList[i]
+        #injection
+        append!(y1, time(median(@benchmark homomorphism($f, add_loops($t)))))
```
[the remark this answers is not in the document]
jhorn00: That was the intention. Should I add a check for this instead? For example, check for self-referencing vertices and if they exist assume the codomain is already in the desired state?

benchmark/Support/autoplot.jl
```diff
@@ -0,0 +1,261 @@
...
+        end
+    end
+end
+
+function autoShuffle(fromGraphs, toGraphs, vertexLimit, edgeLimit)
+    if vertexLimit < 1 && edgeLimit < 1
+        println("autoShuffle() failed: There should be achievable limit parameters.")
+        return
+    end
+    if length(fromGraphs) != length(toGraphs)
```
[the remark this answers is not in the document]
jhorn00: From the docs it doesn't look like `@assert` supports interpolation. If I wish to use that, as you demonstrated with `error()`, would the following be acceptable?

`length(fromGraphs) == length(toGraphs) || error("Arguments...some interpolated stuff")`

benchmark/Support/SSets.jl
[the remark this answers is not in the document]
jhorn00: I'll continue making corrections after my Exam on Monday 4/25.

benchmark/Support/graph_bank.jl
jhorn00: This is now implemented. Let me know if another version/variation is desired.

benchmark/Support/graph_bank.jl
```diff
@@ -0,0 +1,262 @@
...
+add_loops(g) = begin
+    h = copy(g)
+    add_loops!(h)
+    return h
+end
+
+function box_product(g::T, h::T) where {T<:ACSet}
+    g₀, h₀ = T(nv(g)), T(nv(h))
```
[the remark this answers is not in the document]
jhorn00: Let me know if the docstring is sufficient.

jhorn00:
I made a change to how autoGen works (on top of the other fixes). This change provides the same functionality with much shorter (and cleaner) code. Are the other generators/shufflers necessary? If you would like to keep them I can leave them as-is. If you would rather they perform a bit differently I can make those changes.

PR #621, opened 2022-04-21
Representables and the Yoneda embedding

epatters:
This PR implements:

- Representable C-sets, subject to the limitations of our current implementation of left pushforward data migration
- The contravariant Yoneda embedding
- The oppositization 2-functor
- Various small fixes and improvements encountered along the way while using this stuff

### DIFF
--- a/src/categorical_algebra/DataMigrations.jl
+++ b/src/categorical_algebra/DataMigrations.jl
@@ -1,7 +1,8 @@
 """ Functorial data migration for attributed C-sets.
 """
 module DataMigrations
-export DataMigration, DeltaMigration, SigmaMigration, migrate, migrate!
+export DataMigration, DeltaMigration, SigmaMigration, migrate, migrate!,
+  representable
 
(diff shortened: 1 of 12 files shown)


src/categorical_algebra/DataMigrations.jl
```diff
@@ -287,9 +287,11 @@ function ob_map(ΣF::SigmaMigration, ::Type{T}, X::ACSet) where T <: ACSet
     if nparts(Y, nameof(dom(hom(diagramD, g)))) == 0
       continue
     end
+    src_colim, tgt_colim = colimX[src(diagramD, g)], colimX[tgt(diagramD, g)]
+    ϕ = hom(comma_cats, comma_hom_map[g])
```
epatters, line 291: @slibkind, I found a bug in the sigma migration code and implemented what seems to be a fix. However, I haven't taken the time to properly understand this code. Could you take a quick look and confirm that this make sense?

PR #631, opened 2022-05-05
Missing values in data attributes of acsets

epatters:
WIP.


epatters:
This was a hack to make a demo work. Varacsets provide a much better solution (#740).

PR #635, opened 2022-06-28
Move DPO related content to AlgebraicRewriting

kris-brown:
Rewriting is a self-contained application of categorical algebra that nothing else   in Catlab depends on. It makes sense to move this to its own [package](https://github.com/AlgebraicJulia/AlgebraicRewriting.jl). This PR would remove the rewriting content in Catlab to avoid code/name duplication when using both packages together.


epatters:
Thanks Kris! I think this moves makes sense.

We should do the merge before the next major release, which I'm thinking should be fairly soon. I'll coordinate separately with everyone about that.

epatters:
Preparing to merge. This functionality is now available in [AlgebraicRewriting.jl](https://github.com/AlgebraicJulia/AlgebraicRewriting.jl).

PR #646, opened 2022-07-25
LooseAcsetTransformation type component coercion 

kris-brown:
The type_components of LooseACSetTransformations can be given the wrong types due to a bug which mixes  `acodom_nums` and `attrtype_num`. This leads to a runtime error when one tries to call an incorrectly-typed function.

In CSets.jl line 334, `attrtype(S)` and `acodom_nums` shouldn't be zipped, as `acodom_nums` is for each attr, not each attrtype. The "i" we want is actually there is `attrtype_num(S,type)`


epatters:
Thanks Kris!

PR #650, opened 2022-07-27
Compile with generators

sjbreiner:
Added a dictionary field to SimpleCompileState and a new kwarg to the `compile` method. If a generator name is in the keys, that symbol is replaced by the associated value in the `generator_expr` method.


epatters:
Thanks Spencer! I'll take a look at this soon.

epatters:
I changed the mechanism slightly, which I think it makes this cleaner. Hopefully it still works for your purposes.

PR #651, opened 2022-07-30
Epi mono factorization in terms of (co)limits

kris-brown:
I've found this construction useful in some other contexts, but this might be of general use.

This factors a morphism into a surjective map into an image and then an injective map into the codomain, where 'image' is defined in terms of limits (so this could be potentially applied in more general contexts than just CSetTransformations). 

Also included are `is_injective` / `is_surjective` test functions for CSetTransformations. (I'm not sure what those should do for attributes)


src/categorical_algebra/CSets.jl
```diff
@@ -386,6 +386,20 @@ function is_natural(α::ACSetTransformation{S}) where {S}
   return true
 end
 
+function is_injective(α::CSetTransformation{S}) where {S}
```
olynch, line 389: This can be generalized to StrictACSetTransformation

olynch (approved):
LGTM

Issue #230, opened 2020-08-05
Theory of groupoids


epatters:
Define the generalized algebraic theory of groupoids, `Groupoid`, extending the theory of categories, `Category`, with another term constructor and two axioms.

Good first issue to learn the basics of GATs.

PR #663, opened 2022-08-22
Theory of groupoids, and syntax system for free groupoids

ssteakley:
This branch attempts to address Issue #230 and implement the generalized algebraic theory of groupoids, `ThGroupoid`. I relied on Catlab's pre-existing implementations of other theories for inspiration (as this is my first contribution), and in order to emulate those examples I took the additional step of implementing a syntax system, `FreeGroupoid`. I created a suite of unit tests for `FreeGroupoid` and included them in `test/theories/Theories.jl`.

The greatest difficulty of this implementation arose in connection to the syntax system, due to the need for suitable rewrite logic for the normalization of composition expressions. I assumed that it should be a requirement of `FreeGroupoid` that the inverse laws (as well as a few related equations) be satisfied by its composition expressions. I quickly found that the utilities that were available in `src/core/Rewrite.jl` could not fulfill this purpose on their own, and I further concluded that I could not even re-use them as parts of a new function. I found this somewhat unfortunate, because `FreeGroupoid` does need some of the same normalization behavior that those functions provide in regard to unitality and associativity. Ultimately, I chose to write a brand new rewriting function, with which, evidently, `FreeGroupoid` passes all the tests that I wrote, although I see some potential flaws in it.

One potential flaw is that I suspect I have broken from the intended pattern of the rewrite utilities, in that the pre-existing utilities each implement a rewriting operation that manipulates a given expression without "looking into" the expression's pieces. I found that I could not make the whole system work unless, in the case of a composition expression that fully cancels out into an identity, my rewrite function "looked at" the GAT `type_args` of the expression in order to determine which identity map must be returned. I put a comment with a "developer's note" in the body of the function (`associate_id_inverse`) which summarizes these thoughts.

I am submitting this as a draft so that it can be reviewed for feedback. Any and all comments are welcome. I am eager to learn more about this project in the process of attempting to contribute.


olynch:
Hi @ssteakley, this looks pretty good. Is there anything that you wanted to add before this gets merged? I'm noticing that it's still a draft PR.

ssteakley:
Hello @olynch, no, I don't have anything to add. I can mark it ready for review. I made it a draft originally just due to the reservations that I described above. Maybe I should have just marked it ready originally. Oops if so

ssteakley:
There is one thing though, which is that the `associate_id_inverse` function body contains a lengthy comment that was meant for the PR review process. So, maybe it should be removed before any merge.

epatters:
Thanks! I'll take a closer look at that before we merge.

epatters:
I shortened your comment, but I don't see a straightforward to avoid the problem that you raised. Ultimately, we want to move away from this manual/ad hoc approach to rewriting GAT exprs to one that is based on pattern matching and rewrite rules, so I'm not going to worry about it right now.

Thanks again for the nicely done PR!

Issue #634, opened 2022-06-07
connected_component_projection_bfs error


kris-brown:
The algorithm currently depends on the vertex order. For the graph with `V=2; E=1`, if the edge is `1->2` I am correctly told there is one connected component, but with `2->1` I am told there are two connected components.

PR #664, opened 2022-08-22
Fix connected_component_bfs #634

lcnhb:
This fixes issue #634 , just had to switch from neighbors to all_neighbors

I added tests aswell


epatters:
This is great, thanks!

PR #665, opened 2022-08-22
Refactor ACSets to use CompTime.jl

olynch:
This is a major change to how acsets are implemented.

First of all, the code for acsets is generated using CompTime.jl. This brings along the benefits of being able to have "dynamic acsets" and "static acsets" share the same code, where with dynamic acsets all of the code is executed at runtime but with static acsets some of the code gets executed ahead of time.

One other change is that indexing is now handled at a lower level than acsets, in a new file called "ACSetColumns.jl". Index management is now not code-generated; it is handled with normal Julia method dispatch. Hopefully this will ease the maintenance burden, and make innovation easier.

This won't pass checks until CompTime is a registered package, which should happen by the afternoon of August 22nd.


src/categorical_algebra/CSetDataStructures.jl
```diff
@@ -671,64 +656,107 @@ function Base.show(io::IO, acs::T) where {S,T<:StructACSet{S}}
...
+    set_subpart!(acs,:,@ct(f), rows[@ct f])
+  end
+  acs
+end
+
+function make_acset(T::Type{<:StructACSet{S}}, rows::NamedTuple) where {S}
+  _make_acset(Val{S}, T, rows)
+end
+
+macro acset(head, body)
```
epatters, line 739: This macro should have a docstring. I guess it didn't have one before but we may as well add one now.

src/wiring_diagrams/Directed.jl
```diff
@@ -205,6 +205,12 @@ mutable struct WiringDiagram{Theory, PortValue, WireValue, BoxValue} <: Abstract
     # Copy constructor for shallow copy
     new{T, PortValue, WireValue, BoxValue}(copy(f.diagram), f.value)
   end
+  function WiringDiagram{T, PortValue, WireValue, BoxValue}(
```
epatters, line 208: I know you spent a lot of time tracking down this bug, we don't actually need this new constructor, do we? If not, I'd suggest just removing it.

epatters (changes requested):
Thanks Owen, this refactor has turned out wonderfully! At this point, I don't have much to say, just a few small comments below.

src/wiring_diagrams/Directed.jl
epatters: Sorry, I guess I was wrong: the test failures suggest that you did need this constructor after all.

PR #675, opened 2022-09-08
Upgrade PrettyTables.jl to v2

epatters:


epatters:
On hold until DataFrames.jl upgrades as well (see https://github.com/JuliaData/DataFrames.jl/pull/3131 and https://github.com/JuliaData/DataFrames.jl/pull/3096). We can't have Catlab be incompatible with DataFrames.

ronisbr:
Hi @epatters 

In your case, the only change was two keywords. Both are set as deprecated (https://github.com/ronisbr/PrettyTables.jl/blob/6b5030845a5f011318d2542b11e630c2350672bc/src/private.jl#L325). Hence, I think you can use both versions 1.3 and 2.0 of PrettyTables.jl without problems here.

bkamins:
We will try to do 1.4 release in September 2022.

epatters:
Thanks both, very helpful!

Issue #656, opened 2022-08-11
No hash method for acsets


kris-brown:
The new way C-Sets are hashed does not make C-Sets which are literally equal have the same hash value.

```julia 
using Catlab.Graphs
hash(Graph()) == hash(Graph()) # false
```

kris-brown:
I can eventually do a PR with some tests with Sets / Dicts of C-sets.

epatters:
Thanks Kris, adding some tests would be great. The problem seems to be that we haven't even implemented `Base.hash` for acsets. We gotta fix that!

```julia
julia> methods(Base.:(==), [Graph,Graph])
# 1 method for generic function "==":
[1] ==(x1::T, x2::T) where T<:StructACSet in Catlab.CSetDataStructures at /home/epatters/.julia/packages/Catlab/9oZsA/src/categorical_algebra/CSetDataStructures.jl:199

julia> methods(Base.hash, [Graph,Graph])
# 0 methods for generic function "hash":
```

PR #677, opened 2022-09-14
Hash method for ACSets

kris-brown:
To avoid situations where `a==b` but `hash(a)!=hash(b)`, a hash method for C-Sets is added.

Addresses https://github.com/AlgebraicJulia/Catlab.jl/issues/656


epatters:
Thanks for the fix, Kris! Unfortunately, the CompTime refactor just got merged so you'll have to update this PR to fix the conflict.

epatters (approved):

PR #679, opened 2022-09-23
Small fix for Presentation -> SchemaDesc function 

kris-brown:
If the presentation has no homs/attrs, then the type of `doms` passed to `SchemaDesc` is `Dict{Any,Any}` when `Dict{Symbol,Symbol}` is required by `SchemaDesc`. Just explicitly marking the type of the dictionary solves this.


olynch (approved):
Looks good, lgtm

PR #681, opened 2022-09-27
Partiality in ACSets handled in a well-defined and well-understood way

olynch:
Currently, the story for partiality in ACSets is that we assume it doesn't exist, except for when it's convenient in building them.

There are also some subtle bugs, some of which we know exist (like int-valued attributes defaulting to zero) and some of which we are only fearful of (???) due to the janky way we currently handle partiality.

For this PR, we should have:
- A thorough control and understanding over where partiality will throw errors, where partiality will return default values, and where partiality will return "none" types.
- A coherent story for partial ACSets, where variables for attributes might be introduced
- A coherent story for ACSet transformations that could be partial.

As part of this work, acsets also have been migrated into a separate folder, and no longer depend on presentations.


docs/src/index.md
```diff
@@ -82,8 +82,44 @@ There are several core parts to the Catlab design, we start with a brief overvie
 
 4. [`@present`](@ref) enumerates a finite set of generators for a model of the GAT just like you would write out a group (model of the theory of groups) as list of generators and relations, the presentation lets you enumerate the objects and morphisms that generate a category.
 
+## Conventions
 
+In several places in Catlab, we use what we call "Abstract Field Convention". Instead of doing the following:
```
epatters, line 87: On second thought, I am wondering what is the motivation for this "abstract field convention." Usually my preference is to wrap fields with generic accessor functions, which allows for a uniform interface to different structs, and the Julia style guide suggests that one should [prefer exported methods over direct field access](https://docs.julialang.org/en/v1/manual/style-guide/#Prefer-exported-methods-over-direct-field-access). Is performance what you're worried about here?

src/acsets/Columns.jl
```diff
@@ -0,0 +1,393 @@
...
+    remove_mapping!(c.i, c.m[x], x)
+    maybe_unstore!(c.i, c.m[x])
+  end
+  store!(c.i, y)
+  add_mapping!(c.i, y, x)
+  c.m[x] = y
+end
+
+# This is to maintain compatibility with old convention for unique_indexing
+struct InconsistentPreimageFlag end
```
epatters, line 362: Some real shade being thrown with this name ;) But really, this should have a more descriptive name, maybe something like `UnboxUniquelyIndexedPreimages`

src/acsets/ColumnImplementations.jl
```diff
@@ -0,0 +1,592 @@
...
+
+Columns.isstored(m::PartialDictMap, x) = true
+Columns.store!(m::PartialDictMap, x) = nothing
+
+"""
+This is a dictionary-backed mapping that returns a default value
+when accessed at an undefined index.
+"""
+@struct_hash_equal struct DefaultDictMap{S,T,D<:Default{T}} <: DefaultMapping{S,T}
+  d::Dict{S,T}
```
epatters, line 235: Would it simplify the code to use `DefaultDict` from DataStructures (which we already depend on)?

src/acsets/ColumnImplementations.jl
```diff
@@ -0,0 +1,592 @@
...
+"""
+A column for a dict-backed injective hom with key type K
+"""
+struct SparseInjectiveAttr{K, T} <: Column{K,T}
+  m::PartialDictMap{K,T}
+  i::InjectiveIC{K, T, PartialDictMap{T, K}}
+end
+
+
+@data HomOrAttr begin
```
epatters, line 535: This whole bit for selecting the default column types is slick.

src/acsets/DenseACSets.jl
```diff
@@ -470,22 +364,41 @@ ACSetInterface.has_subpart(acs::DynamicACSet, f::Symbol) =
...
+domain(acs::DynamicACSet, f::Symbol) = runtime(_domain, acs, acs.schema, f)
 
-ACSetInterface.incident(acs::DynamicACSet, ::Colon, f::Symbol) where {S} =
-  runtime(_incident, acs, acs.schema, :, f)
+@ct_enable function _domain(acs, @ct(S), @ct(f))
+  @ct s = Schema(S)
+  parts(acs, @ct dom(s, f))
+end
+
+@inline ACSetInterface.incident(acs::SimpleACSet, part, f::Symbol; consistent=false) =
```
epatters, line 379: Again, rather than calling this flag `consistent`, we should call it something meaningful and descriptive.

src/acsets/LVectors.jl
```diff
@@ -0,0 +1,40 @@
+module LVectors
+export LVector
+
+using CompTime
+
+struct LVector{T,Labels} <: AbstractVector{T}
```
epatters, line 6: Perhaps add a docstring explaining that this type provides similar functionality to the corresponding type in LabeledArrays.jl but without all the dependencies.

src/theories/Schema.jl
```diff
@@ -36,169 +36,3 @@ abstract type AttrExpr{T} <: SchemaExpr{T} end
   compose(f::Hom, x::Attr) = associate_unit(new(f,x; strict=true), id)
 end
 
-# Type-level representation of Schema
```
epatters, line 39: Glad to see this functionality banished from `Catlab.Theories`.

test/categorical_algebra/CSetDataStructures.jl
```diff
@@ -55,9 +53,9 @@ for dds_maker in dds_makers
 
   @test has_subpart(dds, :Φ)
   @test !has_subpart(dds, :nonsubpart)
-  # @test_throws ArgumentError subpart(dds, 1, :nonsubpart)
-  # @test_throws ArgumentError incident(dds, 1, :nonsuppart)
-  # @test_throws ArgumentError set_subpart!(dds, 1, :nonsubpart, 1)
+  @test_throws Any subpart(dds, 1, :nonsubpart)
```
epatters, line 56: These still throw an except, right? If you don't want to be specific, you should use the abstract type `Exception` instead of `Any`.

test/categorical_algebra/CSets.jl
```diff
@@ -27,19 +27,19 @@ add_edges!(g, 2:4, 3:5)
 @test FinSet(g, :V) == FinSet(6)
 f = FinFunction(g, :V)
 @test collect(f) == 1:6
-@test is_indexed(f)
+# @test is_indexed(f)
```
epatters, line 30: Are these supposed to be commented?

epatters (changes requested):
Another clean and elegant refactor of the acsets machinery! This will open the door to all sorts of acset extensions, such as sparse or distributed acsets.

docs/src/index.md
```diff
@@ -82,8 +82,44 @@ There are several core parts to the Catlab design, we start with a brief overvie
 
 4. [`@present`](@ref) enumerates a finite set of generators for a model of the GAT just like you would write out a group (model of the theory of groups) as list of generators and relations, the presentation lets you enumerate the objects and morphisms that generate a category.
 
+## Conventions
```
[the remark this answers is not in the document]
olynch: I think we should definitely have a "for the developers" section of the docs, because most of the users of Catlab are also developers. Opened: https://github.com/AlgebraicJulia/Catlab.jl/issues/686#issue-1398437913

src/acsets/ColumnImplementations.jl
olynch: No, because this is done in a consistent way with DefaultVecMap, and if I used DefaultDict then this would be different.

src/acsets/ColumnImplementations.jl
olynch: Thanks!

test/categorical_algebra/CSets.jl
olynch: This is a temporary performance regression where indexing is not preserved; issue https://github.com/AlgebraicJulia/Catlab.jl/issues/687#issue-1400162233

PR #682, opened 2022-09-28
Gluing together Attributed CSets in limited scenario

kris-brown:
We currently support the ability to glue attributed C-sets where the merged parts agree completely on their attributes. This PR extends this colimit functionality with the ability to apply a fixed transformation to attributes. We require this to be known in advance and explicitly provided. 

This is called `glue` because it is not technically a colimit of LooseACSetTransformations, which is a shame because the code is 95% identical (perhaps a change can be made such that both `colimit` and `glue` share more code?).

There is an analogous operation (called `coglue` here) that performs a limit of LooseACSetTransformations where you can specify up front what the `type_components` of the resulting legs should be.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -8,7 +8,7 @@ export ACSetTransformation, CSetTransformation,
   isomorphism, isomorphisms, is_isomorphic,
   generate_json_acset, parse_json_acset, read_json_acset, write_json_acset,
   generate_json_acset_schema, parse_json_acset_schema,
-  read_json_acset_schema, write_json_acset_schema, acset_schema_json_schema
+  read_json_acset_schema, write_json_acset_schema, acset_schema_json_schema, glue
 
 using Base.Iterators: flatten
 using Base.Meta: quot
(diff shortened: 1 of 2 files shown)


epatters:
The part about colimits I refactored as #688. I don't understand the part about limits in a categorical way. How important is that part? Do you have a use case for it?

kris-brown:
Hi Evan, there is a very clear use case for the limits. In all of our labelled Petri net stratification demos, we define a labeled petri net `typed` (with transitions labelled, e.g., "infect", "disease", "strata") and then awkwardly do a slight-of-hand where a typed SIR model is a map into `typed'`, not `typed`, where `type'` is the same but has its attributes all rewritten to `nothing`. 

Moreover, in a recent [demo](https://github.com/AlgebraicJulia/ASKEM-demos/blob/kris-brown-patch-1/StratificationMadeEasy.ipynb) I wrote a function which abstracts away the tedium of adding "reflexive transitions" in order to stratify two models. You control which cross terms are included like so:
```julia
res = stratify(SIRD_typed=>[[:strata],[:strata],[:strata],[]], # S I R D
               Quarantine_typed=>[[:disease], [:disease,:infect]],# Q NQ 
               typed) 
```
It's natural here to make reference to the names of transitions of the type system. It's quite confusing/awkward to have to pass in two homomorphisms `A->Typed'`, `B->Typed'` and then again have to pass in `Typed` as a third argument just because the codomain of the typed models is "fake" in some sense.

The limit code here would also give us control over what the pullback apex attributes look like (maybe I want a symbol delimited with an underscore rather than a pair of symbols because I want to stay within `LabeledPetriNet{Symbol}`).

kris-brown:
More thoughts on the limit-y half of this code. The implementation in this PR is limited to `type_components` which you can take `preimage` of, so if you wanted to take a pullback of `Float64`s that multiplied them, this would be very hard to do using an arbitrary `SetFunction` function. A more robust solution is for a limit of `LooseACSetTransformation`s to optionally accept a `SetFunction` for each attrtype that takes an argument for each foot of the limit. This will naturally lend itself to computing the projection maps as `FinDomFunction`s. 

(another concern with actually multiplying out the rates in the pullback rather than just storing the pair of rates is that the `type_components` is a function on `Float64`, so `12` might map to `3*4` and to `2*6` by chance which would lead to a runtime error)

epatters:
See #740.

PR #688, opened 2022-10-07
Colimits of acsets with given type components

epatters:
Based on @kris-brown's #682, or at least the part of it concerning colimits. This stuff is a bit subtle but hopefully the fairly detailed docstrings explain adequately what is going on here.


kris-brown (approved):
looks good to me!

PR #693, opened 2022-10-16
Use intermediating AST in diagram macros

epatters:
This refactor required a lot more work than I expected, but the improved design should be make it easier to maintain and extend the `@migration` macro and related macros.


epatters:
Quickly self-merging because deadlines are on the horizon and my next PR will depend on this one.

PR #699, opened 2022-10-27
Typo in type declaration of `FinDomFunctionDict`

epatters:
I'm not sure why Julia didn't throw an error here given the type constraint on `FinDomFunction`, but the issue can be avoided entirely since `FinFunction` and `FinDomFunction` are just type aliases for `SetFunction` subject to certain constraints.

### DIFF
--- a/src/categorical_algebra/FinSets.jl
+++ b/src/categorical_algebra/FinSets.jl
@@ -396,7 +396,7 @@ The domain is a `FinSet{S}` where `S` is the type of the dictionary's `keys`
 collection.
 """
 @struct_hash_equal struct FinDomFunctionDict{K,D<:AbstractDict{K},Codom<:SetOb} <:
-    SetFunction{FinSet{AbstractSet{K},K},Codom}
+    SetFunction{FinSetCollection{Base.KeySet{K,D},K},Codom}
   func::D
   codom::Codom
 end
--- a/src/categorical_algebra/FinSets.jl
+++ b/src/categorical_algebra/FinSets.jl
@@ -237,7 +237,7 @@ The domain of this function is always of type `FinSet{Int}`, with elements of
 the form ``{1,...,n}``.
 """
 struct FinDomFunctionVector{T,V<:AbstractVector{T}, Codom<:SetOb{T}} <:
-    FinDomFunction{Int,FinSetInt,Codom}
+    SetFunction{FinSetInt,Codom}
   func::V
   codom::Codom
 end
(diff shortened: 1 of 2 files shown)


ssteakley:
Great, this makes more sense now. `myfunctiondict` is recognized as an instance of `FinDomFunction{AbstractSet{Char}}`.

There is still a certain quirk, though, which is that `dom(myfunctiondict)` doesn't agree with the type of `myfunctiondict` as a `SetFunction`. `myfunctiondict` is an instance of `SetFunction{FinSet{AbstractSet{Char}, Char}, FinSetCollection{Set{Int64}, Int64}}`, which suggests that the domain of `myfunctiondict` should be a `FinSet{AbstractSet{Char}, Char}`. However, when I run the corresponding type assertion, I get this:
```
dom(myfunctiondict) :: FinSet{AbstractSet{Char}, Char}
ERROR: TypeError: in typeassert, expected FinSet{AbstractSet{Char}, Char}, got a value of type FinSetCollection{Base.KeySet{Char, Dict{Char, Int64}}, Char}
```

I think a possible way to resolve this would be to add another parameter to the type declaration of `FinDomFunctionDict`, something like this:
```
@struct_hash_equal struct FinDomFunctionDict{K,S<:AbstractSet{K},D<:AbstractDict{K},Codom<:SetOb} <:
    SetFunction{FinSet{S,K},Codom}
  func::D
  codom::Codom

  function FinDomFunctionDict(f::D, codom::Codom) where {K,D<:AbstractDict{K},Codom<:SetOb}
    S = typeof(keys(f))
    new{K,S,D,Codom}(f, codom)
  end
end
```
This way, the domain of a `FinDomFunctionDict` (as a `SetFunction`) really is the type of the underlying dictionary's `keys` collection. Any thoughts?

epatters:
Fair point. Since all the dict types I can find (in Base and DataStructures) have `keys` return a `Base.KeySet`, I've simply used that type rather than adding an extra type parameter that must be extracted at runtime.

We now have:

```julia
julia> ℓ = FinFunction(Dict(:w => 2, :x => 1, :y => 1, :z => 4), FinSet(4))
FinFunction(Dict(:w => 2, :y => 1, :z => 4, :x => 1), FinSet(4))

julia> dom(ℓ) isa FinSet{<:AbstractSet{Symbol},Symbol}
true
```

ssteakley:
That makes sense to me!

epatters:
Great! Thanks for paying attention to these details.

PR #703, opened 2022-11-05
Clean up and tests for acset column system

kris-brown:
Internal improvements to system for acset columns:

- Allow general `AbstractVector` and `AbstractDict` instances in mappings and columns
- Constructors for mappings and columns
- Better names for default column types
- Unit tests for columns functionality that was previously only tested implicitly

This is preparation to refactor FinFunctions to use columns.

:train:


src/acsets/ColumnImplementations.jl
```diff
@@ -446,60 +453,63 @@ end
...
 """
-struct SparseIndexedHom{K} <: Column{K,K}
-  m::PartialDictMap{K, K}
+struct SparseIndexedFinColumn{K,D<:AbstractDict{K,K}} <: Column{K,K}
+  m::PartialDictMap{K, K, D}
   i::StoredPreimageCache{K, K, Set{K},
-                      DefaultDictMap{K, Set{K}, DefaultEmpty{Set{K}}}}
+                      DefaultDictMap{K, Set{K}, DefaultEmpty{Set{K}},
+                                     Dict{K,Set{K}}},
+                      }
```
epatters, line 469: Should clean up this formatting.

test/acsets/Columns.jl
```diff
@@ -0,0 +1,103 @@
...
+  @test Columns.isdefined(col, 3)
+  @test isempty(preimage(OneTo(3), col, 1))
+  if coltype <: SparseFinColumn || coltype <: SparseIndexedFinColumn
+    @test_skip preimage(OneTo(3), col, 2) == [1]
+  else 
+    @test preimage(OneTo(3), col, 2) == [1]
+  end
+end
+
+dfc = DenseFinColumn{UnitRange{Int}}(
```
epatters, line 96: This tests columns with an `AbstractVector` that is not a `Vector`. We should also test with an `AbstractDict` that is not a `Dict`.

epatters (changes requested):
Great to have some proper tests for columns!

epatters (approved):

Issue #231, opened 2020-08-05
Permutation groupoid


epatters:
The `Permutations` submodule currently provides some basic algorithms for decomposing permutations, which are used when converting wiring diagrams into expressions. We should extend this module to implement the [permutation groupoid](https://ncatlab.org/nlab/show/permutation+groupoid), also known as `Bij`, as an instance of the theory of groupoids (#230), following the general pattern established for `FinSet` and `FinRel`:

- an abstract type for morphisms in this category
- concrete types for particular computational representations, including at minimum the vector and cycle representations

ssteakley:
`FinSet` has been significantly overhauled since this issue was created. Any thoughts on whether `FinBij` should more follow `FinSet` in its current form, or the older pattern still held by `FinRel`?

PR #707, opened 2022-11-09
WIP - Permutation Groupoid

ssteakley:
This is a chunk of the work I've been doing toward implementation of #231. The main result here is the implementation of `FinBijectionCycles`: a concrete type for finite automorphisms that are backed by a set of cycles. `Cycle`s are implemented as `CircularVector`s, from `CircularArrays.jl`, which are wrapped so that we can provide appropriate implementations of `==` and `hash`.

The commit also introduces an abstract type `SetBijection <: SetFunction` and redeclares `IdentityFunction <: SetBijection`.

Potential Issues:
- Calculation of function values for `FinBijectionCycles`, as implemented here, is meaningfully slower than for vector-backed and dict-backed `FinFunction`s. If the underlying data of the function is truly going to be a set of cycles, then I'm not sure how the speed can be improved. It seems that some transformation would need to be computed.
- For `f :: FinBijectionCycles`, `show(f)` doesn't look great.
- This WIP was largely inspired by the preexisting implementations of the `FinFunction` types. Maybe this has been somewhat preempted, or superceded, by other ongoing work. Namely, the refactor of `FinFunction`s to use Columns.

Feedback would be greatly appreciated!


ssteakley:
After some more work on this, trying to speed up composition for `FinBijectionCycles`, I now feel that this approach may be inherently too slow. Representing a bijection by a set of `Cycle`s matches up intuitively with the mathematical cycle notation, but it is an unwieldy representation for computing function values and composites. This PR's algorithm for computing values, for example, is worst case O(n) for n the size of the domain. Would there be other advantages to this representation that would make it worth these instances of bad performance?

One alternative would be to store the mapping in a more usual way, either as a `Vector` or `Dict`, and then store the cycles themselves as metadata. Essentially, the mapping would be a wrapper around a `FinBijectionDict` or `FinBijectionVector` with some metadata. In the interest of performance, composition of two `FinBijectionCycles` probably should not automatically result in another `FinBijectionCycles`, since this would require the metadata (the new set of cycles, themselves) to be recomputed from scratch.

ssteakley:
More and more I am thinking that implementing `Bijection`s as wrapper types, with `SetFunction`s inside, is what makes the most sense. For me, this goes together with the insight that, for most purposes, if you are going to compute the cycles of a `FinBijection`, you may as well compute and store its inverse at the same time. Both require computing the same data in a loop over the mapping's domain. This suggests that the implementation of `FinBijectionCycles` could profitably be based on that of a type `BijectionBimap`, which would represent a bijection as a pair of (inverse) functions. Computing the inverse along with the cycles is, of course, a little slower, and takes more space. But as a consequence, the resulting object is perhaps more faithful to mathematical cycle notation: indeed, a written set of cycles displays the inverse mapping "alongside" the mapping itself, truly on equal footing.

ssteakley:
I have reimplemented `FinBijectionCycles` based on a system of wrapper types which represent bijections. Now, a `FinBijectionCycles` object is just a pair of inverse `SetFunction`s (i.e. a `BijectionBimap`) with the cycles themselves attached as ancillary data. Some odds and ends are still missing (e.g. an appropriate `show` method for `FinBijectionCycles`), but the code is passing tests for basic functionality.

As I explain in a `# developer's note` in src/CategoricalAlgebra/Sets.jl,  the extant implementation of `CompositeFunction` is disadvantageous for my implementation of composition with inverses. The crucial fact is that iterated composites result in a tree-like structure of nested `CompositeFunction`s, and the type information of the "root" of the tree does not convey the types of the leftmost and rightmost functions of the composite. This actually leads to an inevitable efficiency loss when using `compose` with any `CompositeFunction`. The obvious idea for a way forward would be to try refactoring `CompositeFunction` to store a list of `SetFunction`s instead of a pair, thus eliminating the need to form nested `CompositeFunction`s and allowing the type to expose the types of the leftmost and rightmost functions.

src/categorical_algebra/Sets.jl
```diff
@@ -97,21 +111,88 @@ codom(f::IdentityFunction) = f.dom
...
+unwrap(f::AbsBijectionWrap) = unwrap(f.func)
+unwrap(f::SetFunction) = f
+
+dom(f::AbsBijectionWrap) = dom(unwrap(f))
+codom(f::AbsBijectionWrap) = codom(unwrap(f))
+
+(f::AbsBijectionWrap)(x) = (unwrap(f))(x)
+
+# Developer's note: should I add in a `known_correct` parameter, asking
+# the caller to decide whether the SetFunction should be checked for bijectivity?
```
jpfairbanks, line 139: yes, that is usually how we enable optional correctness checking.

src/categorical_algebra/Sets.jl
```diff
@@ -97,21 +111,88 @@ codom(f::IdentityFunction) = f.dom
...
+@struct_hash_equal struct BijectionThinWrap{Dom<:SetOb, Codom<:SetOb,
```
jpfairbanks, line 140: `Thin` has a technical meaning in CT, that there is at most one morphism between two objects. Here I think you are just meaning "thin wrapper" like a lightweight wrapper type. It would be good to avoid the word thin for this non-jargon usage.

src/categorical_algebra/Sets.jl
```diff
@@ -97,21 +111,88 @@ codom(f::IdentityFunction) = f.dom
...
+    F<:SetFunction{Dom,Codom}} <: AbsBijectionWrap{Dom, Codom, F}
+  func::F
+end
+
+BijectionThinWrap(f::BijectionThinWrap) = f
+
+function Base.show(io::IO, f::BijectionThinWrap)
+  replacerules = (r"^SetFunction" => "Bijection", r"^FinFunction" => "FinBijection")
+  print(io, replace(repr(f.func), replacerules...))
```
jpfairbanks, line 149: Make sure to include a test for this so that we can detect breakage. I can imagine changing the "SetFunction" and this pattern not matching any more.

src/categorical_algebra/Sets.jl
```diff
@@ -191,12 +297,71 @@ end
...
+# composites are represented essentially as a tree of `SetFunction`s. As a
+# consequence, it is not possible to tell from the type of a `CompositeFunction`
+# whether the leftmost (or rightmost) item of the composite is a
+# `BijectionBimap`. Because of all this, the implementation introduces an
+# efficiency loss whenever using `compose` with a `CompositeFunction`, which
+# grows worse as the "tree" of composites grows deeper. To make things simpler
+# for `compose_inv`, the obvious idea would be to try refactoring
+# `CompositeFunction` to store a list of `SetFunction`s instead of a pair.
+
+function compose_inv(f::SetFunction, g::SetFunction)
```
jpfairbanks, line 325: I think this might be easier with MLStyle pattern matching and structural recursion.

src/categorical_algebra/FinSets.jl
```diff
@@ -553,6 +576,166 @@ Sets.do_compose(f::FinFunctionDict{K,D}, g::FinDomFunctionDict) where {K,D} =
...
+FinBijection(f::AbstractVector, a, args...) =
+  BijectionThinWrap(FinDomFunction(f, FinSet(a), args...))
+FinBijection(f::AbstractDict, args...) =
+  BijectionThinWrap(FinFunction(f, args...))
+
+function FinBijection(f::Union{AbstractDict{K,Int},AbstractVector{Int}}) where K
+  function minandmax(p::Tuple{<:Number, <:Number}, n::Int)::Tuple{Int,Int}
+    (Int(min(p[1], n)), Int(max(p[2], n)))
+  end
+  minval, maxval = reduce(minandmax, values(f), init=(Inf, -Inf))
```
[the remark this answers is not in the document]
ssteakley: Thanks, will change the sentinel value. About the ordering thing, if I'm not misunderstanding you, `minandmax` is already returning the values in that order, as is.

ssteakley:
Thanks for the comments @jpfairbanks ! For now, I decided not to worry about implementing the `known_correct` parameter and corresponding checkers; that can wait. I addressed the other comments, and made a couple more changes. If this is looking acceptable, then I could try to proceed with the rest of the code for the permutation groupoid.

Other needed steps:
- [ ] Write more tests

PR #710, opened 2022-11-12
Delta migration of acset transformations

kris-brown:
Functorial data migration should work on morphisms of ACSets too. It seems simple to work this out for delta - I'm not sure how other migrations might look. In the case of delta, it is straightforward to apply the functor to the domain and codomain and then copy the components (or type components) based on the data of the functor. 

One interesting thing I noticed is that it's not possible to apply a DeltaMigration to a LooseACSetTransformation `f: WeightedGraph{Int} -> WeightedGraph{Symbol}`, for example. This is because the data of a DeltaTransformation requires you to specify the concrete ACSet types, which means you can't apply the same DeltaMigration to both the domain and codomain of `f`.

### DIFF
--- a/src/categorical_algebra/Chase.jl
+++ b/src/categorical_algebra/Chase.jl
@@ -347,6 +347,26 @@ end
 
 # Sigma migration
 #################
+"""
+Convert a CSet morphism X->Y into a CSet on the schema C->C 
+(collage of id functor on C).
+"""
+collage(f::ACSetTransformation{S}) where S = collage(f, Presentation(S))
+function collage(f::ACSetTransformation, S::Presentation)
+  colim, col_pres = collage(id(FinCat(S)))
+  colimL, colimR = colim
+  res = AnonACSet(col_pres)
+  for o in ob(Schema(S))
+    add_parts!(res, nameof(ob_map(colimL,o)), nparts(dom(f), o))
+    add_parts!(res, nameof(ob_map(colimR,o)), nparts(codom(f), o))
+    set_subpart!(res, Symbol("α_$o"), collect(f[o]))
+  end
+  for h in homs(Schema(S); just_names=true)
+    set_subpart!(res, nameof(hom_map(colimL,h)), dom(f)[h])
+    set_subpart!(res, nameof(hom_map(colimR,h)), codom(f)[h])
+  end
+  colim, res
+end 
 
 """
 A collage of a functor is a schema encoding the data of the functor
--- a/src/categorical_algebra/FunctorialDataMigrations.jl
+++ b/src/categorical_algebra/FunctorialDataMigrations.jl
@@ -117,6 +117,25 @@ function migrate!(X::ACSet, Y::ACSet, FOb, FHom)
   migrate!(X, Y, M)
 end
 
+# On morphisms
+migrate(t::Type{T}, f::ACSetTransformation, M::DeltaMigration) where T <: ACSet = 
+  migrate(t, f, M.functor)
+
+migrate(t::Type{T}, f::ACSetTransformation, F::FinFunctor) where T <: ACSet = 
+  migrate(t, f, ob_map(F), hom_map(F))
+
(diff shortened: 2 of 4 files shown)


src/categorical_algebra/DataMigrations.jl
```diff
@@ -161,6 +162,30 @@ function migrate!(X::ACSet, Y::ACSet, FOb, FHom)
...
+function (::Type{T})(X::ACSet, F::FinFunctor) where T <: ACSet
+  Base.depwarn("Data migration constructor is deprecated, use `migrate` instead", :ACSet)
+  migrate(T, X, F)
+end
+
+function (::Type{T})(X::ACSet, FOb::AbstractDict, FHom::AbstractDict) where T <: ACSet
+  Base.depwarn("Data migration constructor is deprecated, use `migrate` instead", :ACSet)
+  migrate(T, X, FOb, FHom)
+end
+
```
KevinArlin, line 174: Why add new deprecated methods?

src/categorical_algebra/DataMigrations.jl
```diff
@@ -161,6 +162,30 @@ function migrate!(X::ACSet, Y::ACSet, FOb, FHom)
...
+  for (ob_dom,ob_codom) in pairs(FOb)
+    if Symbol(ob_codom) ∈ keys(components(f))
+      d[Symbol(ob_dom)] = f[Symbol(ob_codom)]
+    end
+  end
+  Fd, Fcd = migrate(T, dom(f), FOb, FHom), migrate(T, codom(f), FOb, FHom)
+  TightACSetTransformation(NamedTuple(d), Fd, Fcd)
+end
+
+
```
KevinArlin, line 188: Shouldn't there be methods that let you migrate using a `Functor` or a `DataMigration` too?

KevinArlin: More generally it would probably be nice to have the data flow for migrating morphisms match that for migrating ACSets, so eg having the `FOb,FHom` API go through the `Migration` API rather than short-circuiting straight to the end. If you wrote it this way with performance in mind, I'd say that users ought to build the migration once and for all if they're going to be migrating a lot along the same migration.

KevinArlin: Other than that, looks good; my only other thought is whether it might be nice to have a method `migrate(f,M,X,Y)` where `X` and `Y` are *already migrated* to the target category. This would be noticeably faster if you were migrating many morphisms among few acsets at some point. And the more usual `migrate` method could just migrate `X` and `Y` and then call into this one.

src/categorical_algebra/DataMigrations.jl
kris-brown: oops that looks like a rebase error! will delete those!

src/categorical_algebra/DataMigrations.jl
kris-brown: I'll add the things that didn't get covered in this PR as an issue to be addressed after MigrationFest2023!

src/categorical_algebra/DataMigrations.jl
```diff
@@ -161,6 +162,27 @@ function migrate!(X::ACSet, Y::ACSet, FOb, FHom)
   migrate!(X, Y, M)
 end
 
+
+# On morphisms
+
+migrate(t::Type{T}, f::ACSetTransformation, M::DeltaMigration) where T <: ACSet = 
+  migrate(t, f, M.functor)
```
KevinArlin, line 169: This will probably be able to be just any kind of migration, but maybe let's not worry about that until datamigrations.jl comes out.

KevinArlin (approved):

KevinArlin:
I think this is ready to merge @epatters

epatters:
This PR needs to rebased now that #846 has been merged.

epatters (approved):
Thanks both, LGTM!

PR #717, opened 2022-11-30
Option `product_attrs` for ACSet limits 

kris-brown:
We may want to take a limit, such as a pullback, on ACSets where we are actually only computing the limit of the *combinatorial* data. The question is how to handle the attributes in a canonical way: this is addressed by making the attributes of the apex of the limit to be the product of all* the attributes of objects in the diagram.

*: This is not strictly true: we only take the product of objects in the diagram which apex has legs into. Limits mathematically have legs into all objects of the diagram that we compute the limit of, but the `cone_obs` picks out specific objects for which we want to explicitly represent the legs to. This is typically the objects in the diagram which have no incoming morphisms.

### DIFF
--- a/test/categorical_algebra/CSets.jl
+++ b/test/categorical_algebra/CSets.jl
@@ -309,6 +309,20 @@ pb = ob(pullback(ϕ, ψ))
 @test (nv(pb), ne(pb)) == (8, 2)
 @test Set(pb[:weight]) == Set([(+1, -1), (-1, +1)])
 
+# Limit with LooseACSetTransformations 
+A = @acset VELabeledGraph{Symbol} begin V=2; vlabel=[:a,:b] end;
+B = @acset VELabeledGraph{Symbol} begin V=2; vlabel=[:q,:z] end;
+C = @acset VELabeledGraph{Symbol} begin V=2; vlabel=[:x,:y] end;
+ac = LooseACSetTransformation(
+  Dict([:V=>[1,2]]),Dict([:Label=>FinFunction(Dict(:a=>:x,:b=>:y))]),A,C);
+bc = LooseACSetTransformation(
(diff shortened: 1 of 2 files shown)


test/categorical_algebra/CSets.jl
```diff
@@ -309,6 +309,20 @@ pb = ob(pullback(ϕ, ψ))
 @test (nv(pb), ne(pb)) == (8, 2)
 @test Set(pb[:weight]) == Set([(+1, -1), (-1, +1)])
 
+# Limit with LooseACSetTransformations 
+A = @acset VELabeledGraph{Symbol} begin V=2; vlabel=[:a,:b] end;
+B = @acset VELabeledGraph{Symbol} begin V=2; vlabel=[:q,:z] end;
+C = @acset VELabeledGraph{Symbol} begin V=2; vlabel=[:x,:y] end;
+ac = LooseACSetTransformation(
+  Dict([:V=>[1,2]]),Dict([:Label=>FinFunction(Dict(:a=>:x,:b=>:y))]),A,C);
```
epatters, line 317: Not important, but you can save a few characters by replacing `Dict([:V=>[1,2]])` with `Dict(:V=>[1,2])`, etc.

epatters (changes requested):
Thanks Kris! Code looks great. The main thing needed is that the new keyword argument `product_attrs` should be document somewhere. I'm always a little unsure about how best to document methods for a generic function that already has lots of methods, but FWIW I documented the `type_components` keyword arg in the docstring for `LooseAcsetTransformation`. But maybe it would be better to just have dedicated docstrings for limits and colimits of acsets, since these are now acquiring more bells and whistles.

epatters:
As we discussed earlier, just calling `cone_objects` won't do what we want for general limit diagrams. But I'm OK with putting that off for now. I'll merge this as-is.

epatters (approved):

PR #723, opened 2022-12-07
Get Catlab version without assuming it is a dependency

aaguinal:


olynch:
Thanks Angeline! It's hard to write a test for this, but I'm assuming you tried it out yourself and it worked?

aaguinal:
> Thanks Angeline! It's hard to write a test for this, but I'm assuming you tried it out yourself and it worked?

No problem! I did test it out, it worked, and had no regressions. Yeah, it's a bit tough to figure out a good test for it because the version will change every so often. If anything, I would be like writing a unit test for the `Pkg.project()` function, which I would hope they write a test for.

epatters (approved):
Thanks Angeline!

PR #731, opened 2023-01-12
Reformat Tuple node labels in Graphviz graphs

slwu89:
This is a very small PR that just lets `to_graphviz` work with `node_labels` using labels that are Tuples. The reason this is useful is that some operations can change the types of labels on the resulting object. For example, taking the product of a `LabeledGraph{Symbol}` and `LabeledGraph{String}` results in a new label of type `Tuple{Symbol, String}`. This just makes the internal functions in the GraphvizGraphs module break those tuples down into a string for plotting, it's mostly useful for pedagogy.

### DIFF
--- a/src/graphics/GraphvizGraphs.jl
+++ b/src/graphics/GraphvizGraphs.jl
@@ -165,7 +165,6 @@ function to_graphviz_property_graph(g::AbstractGraph;
   )
 end
 
-# node_label(g, name::Symbol, v::Int) = Dict(:label => string(g[v, name]))
 node_label(g, name::Symbol, v::Int) = Dict(:label => node_label_to_string(g[v, name]))
 node_label(g, labels::Bool, v::Int) = Dict(:label => labels ? string(v) : "")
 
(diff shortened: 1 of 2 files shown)


epatters (changes requested):
Thanks Sean! I'm little leery of changing the display of standard Julia types, as it might seem arbitrary or unexpected to some users, but I can see the appeal when the attribute types come from product operations. So I'm inclined to go ahead with this, but for consistency, whatever transformation is applied to the node labels should also be applied to the edge labels, since edge attributes can just as easily be involved in products.

slwu89:
> Thanks Sean! I'm little leery of changing the display of standard Julia types, as it might seem arbitrary or unexpected to some users, but I can see the appeal when the attribute types come from product operations. So I'm inclined to go ahead with this, but for consistency, whatever transformation is applied to the node labels should also be applied to the edge labels, since edge attributes can just as easily be involved in products.

Thanks for pointing out that missing element. I added that to the edge labels too.

epatters (approved):
Thanks Sean, LGTM!

PR #732, opened 2023-01-14
Option to randomize C-set homomorphism search

kris-brown:
The current backtracking search algorithm looks at the parts of a C-Set in a deterministic order. A keyword which shuffles this order could allow us to find a *single* morphism randomly without having to search for *all* morphisms and then pick one randomly.


epatters (approved):
Thanks Kris, LGTM!

PR #733, opened 2023-01-26
Add debug flag to `is_natural`

kris-brown:
It can be very helpful when debugging to see why an ACSet transformation is unnatural. When the keyword `debug` is set to true, a list of violations is is returned. (This is type unstable, but `is_natural` shouldn't be used in a hot loop, anyways).

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -421,19 +421,19 @@ Xₕ ↓  ✓  ↓ Yₕ
   Xₙ --> Yₙ
      αₙ 
 
-If debug is true, return a list of violations, with elements such as: 
-  (h, index i in Xₐ, Yₕ(αₘ(i)), αₙ(Xₕ(i)))
+If `return_failures` is true, return a list of violations, with elements such as: 
+  (h, index i in Xₘ, Yₕ(αₘ(i)), αₙ(Xₕ(i)))
 
 This is type-unstable, so it should not be used in performance-sensitive code.
 """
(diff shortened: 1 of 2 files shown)


kris-brown:
inspiration from https://github.com/AlgebraicJulia/Catlab.jl/pull/599

src/categorical_algebra/CSets.jl
```diff
@@ -412,13 +412,36 @@ end
...
+Xₕ ↓  ✓  ↓ Yₕ
+  Xₙ --> Yₙ
+     αₙ 
+
+If debug is true, return a list of violations, with elements such as: 
+  (h, index i in Xₐ, Yₕ(αₘ(i)), αₙ(Xₕ(i)))
+
+This is type-unstable, so it should not be used in performance-sensitive code.
+"""
+function is_natural(α::ACSetTransformation{S}; debug::Bool=false) where {S}
```
epatters, line 429: I think this keyword arg should be called something like `return_failures` since `debug` is a bit ambiguous and this functionality might find uses other than debugging.

epatters (changes requested):
Thanks Kris! This will be helpful.

epatters:
Not required for this PR, but if you have a spare moment it would be great to reimplement #599 along these lines since that PR has been stale for a while.

mehalter (approved):
LGTM

src/categorical_algebra/CSets.jl
mehalter: 👍🏼

epatters (approved):

PR #734, opened 2023-01-27
Keyword args for `oapply` to expose (co)limit

olynch:
@epatters this should be merged ASAP and then a new release should be tagged, because the JuliaHub demo depends on it.


epatters (approved):
I made some minor changes. I will merge and release once the tests pass again.

Issue #735, opened 2023-01-30
ACset `map` `MethodError` running `convert`


mehalter:
I'm running into an issue with `map` in AlgebraicPetri even though the types align correctly. Here is an example to replicate the bug (this example does require the `AlgebraicPetri` package):

```julia
using AlgebraicPetri

# Create a labelled reaction net with integer concentrations and rates
# and tuples of symbols
tuple_rxn = AlgebraicPetri.LabelledReactionNetUntyped{Int,Int,Tuple{Symbol,Symbol}}()
add_species!(tuple_rxn, 3, sname=((:U, :S), (:U, :I), (:U, :R)), concentration=(1, 1, 1))
add_transitions!(tuple_rxn, 2, tname=((:Q, :inf), (:Q, :rec)), rate=(1, 1))

# Create a simple function to convert nested tuples of symbols to a single Symbol
flat_symbol(sym::Symbol)::Symbol = sym
flat_symbol(sym::Tuple)::Symbol = mapreduce(x -> isa(x, Tuple) ? flat_symbol(x) : x, (x, y) -> Symbol(x, "_", y), sym)

# Map this `flat_symbol` function to all Name types
# (`tname` and `sname` parts)
map(tuple_rxn, Name=flat_symbol)
```

*Also to note* I get the same behavior even if I specify the parts in particular with

```julia
map(tuple_rxn, tname=flat_symbol, sname=flat_symbol)
```

This results in the following error:

```
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type Tuple{Int64, Int64, Tuple{Symbol, Symbol}}
Closest candidates are:
  convert(::Type{T}, ::T) where T<:Tuple at essentials.jl:331
  convert(::Type{T}, ::Tuple{Vararg{Any, N}}) where {N, T<:Tuple} at essentials.jl:332
  convert(::Type{T}, ::CartesianIndex) where T<:Tuple at multidimensional.jl:131
  ...
Stacktrace:
  [1] setindex!(A::Vector{Tuple{Int64, Int64, Tuple{Symbol, Symbol}}}, x::Int64, i1::Int64)
    @ Base ./array.jl:966
  [2] getindex(#unused#::Type{Tuple{Int64, Int64, Tuple{Symbol, Symbol}}}, vals::Int64)
    @ Base ./array.jl:407
[...]

PR #736, opened 2023-01-31
Simplified map

olynch:
The `map` function now is not generated, which means that it will do some more work every time, but is much less janky, and also potentially will have faster compile times.


epatters:
This PR is reported to fix #735.

epatters (approved):

PR #740, opened 2023-02-03
Variables in ACSets

kris-brown:
This PR introduces a finite set for each ACSet attribute type, called "variables". We can refer to ACSets with this set being nonempty as VarACSets. Not all features supported for ACSets generally will work for VarACSets (e.g. data migrations, limits). The key features now are homomorphism search and colimits.

There is subtlety in ACSetTransformations because there are now two possible maps associated with each attribute type (suppose it is associated with Julia type `T`): a map `FinSet -> FinSet + T` which assigns values for each variable, and a map `T-> T'` for `LooseACSetTransformations`. This is a breaking change because the ACSetTransformation constructor could unambiguously process components and type_components all at once, but now that there are, e.g., two different `Weight` maps for a `WeightedGraph` transformation, we must explicitly pass in `type_components=(Weight=...)` in order to construct a LooseACSetTransformation.

In my experience, many tasks that relied on LooseACSetTransformations (especially homomorphism search/limits/colimits with user-provided type_components) are more naturally handled via VarACSets.

This PR also introduces the pushout-based sigma migration.

VarACSet code hits a bug in Julia type inference which has been addressed in 1.9. This sometimes makes the program crash or hang for incredible amounts of time when doing colimits of VarACSets. For this reason, the minimum Julia version is bumped to 1.9.

This PR contains just basic features - there are a quite a few others in an AlgebraicRewriting branch that might be important to upstream
- "abstracting" an ACSet `X` to producing a map `Y-->X` (where `Y` has only variables)
- substituting variables in an ACSet `X` via a `Dict` to produce a map `X-->Y`.
- limits of ACSets (variables in place of attributes in the apex)
- removing free floating variables


src/acsets/DenseACSets.jl
```diff
@@ -212,7 +215,7 @@ struct AnonACSet{S,Ts,Parts,Subparts} <: StructACSet{S,Ts}
 
   function AnonACSet{S,Ts,Parts,Subparts}() where {S,Ts,Parts,Subparts}
     new{S,Ts,Parts,Subparts}(
-      Parts(zeros(Int, length(S.parameters[2].parameters))),
+      Parts(zeros(Int64, length(S.parameters[2].parameters)+length(Ts.parameters))),
```
mehalter, line 218: This should just be `Int` to help with 32 bit architectures.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -438,9 +383,24 @@ function colimit_representables(F::DeltaSchemaMigration, y)
...
   C = dom(F)
+  ACS = constructor(ob_map(y,first(ob_generators(dom(y)))))
   colimits = make_map(ob_generators(C)) do c
-    Fc = ob_map(F, c)
-    colimit(compose(op(Fc), y))
+    Fc = ob_map(F, c) # e.g. I
+    clim_diag = deepcopy(compose(op(Fc), y))
+    # modify the diagram we take a colimit of to concretize some vars
+    params = Fc isa SimpleDiagram ? Dict() : Fc.params
+    G = dom(clim_diag.diagram).cat.graph
```
epatters, line 392: Use the accessor functions to get these rather than digging through the fields, which again are a private implementation detail.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -438,9 +383,24 @@ function colimit_representables(F::DeltaSchemaMigration, y)
...
+    for (i,val) in collect(params)
+      v = add_vertex!(G; vname=Symbol("param$i"))
+      add_edge!(G, v, i; ename=Symbol("param$i"))
+      at = nameof(ob_map(Fc, i)) # attribute type name 
+      h = only(homomorphisms(ob_map(clim_diag,i), ACS(); initial=Dict(at=>[val])))
+      push!(clim_diag.diagram.ob_map, ACS())
+      push!(clim_diag.diagram.hom_map, h)
```
epatters, line 399: These private fields are not supposed to be mutable. Rather than deep copying the diagram `clim_diag` and mutating it, we should copy the indexing graph and mutate that, then construct a new diagram using that.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -438,9 +383,24 @@ function colimit_representables(F::DeltaSchemaMigration, y)
...
+    end
+    is_functorial(clim_diag.diagram) || error("here")
```
epatters, line 401: Can this ever fail or was this left over from debugging? If the former, should have a proper error message.

src/categorical_algebra/FinCats.jl
```diff
@@ -124,6 +124,14 @@ abstract type FinCatGraph{G,Ob,Hom} <: FinCat{Ob,Hom} end
 """
 graph(C::FinCatGraph) = C.graph
 
+graph(C::OppositeCat) = op(graph(op(C)))
+function op(g::G) where G<:HasGraph
```
epatters, line 128: We should implement `Base.reverse` and `Base.reverse!` in our graphs module, following [Graphs.jl](https://juliagraphs.org/Graphs.jl/dev/core_functions/operators/#Base.reverse), and then have

```julia
graph(C::OppositeCat) = reverse(graph(op(C)))
```

This could be deferred until later, but if so, let's make an issue to track it.

src/categorical_algebra/FinCats.jl
```diff
@@ -406,20 +416,41 @@ not be possible.
 
 See also: [`is_natural`](@ref).
 """
-function is_functorial(F::FinDomFunctor; check_equations::Bool=false)
+function is_functorial(F::FinDomFunctor; return_failures::Bool=false,
```
epatters, line 419: Unfortunately this now conflicts with #762, so you'll have to rebase to use that API.

src/categorical_algebra/FinSets.jl
```diff
@@ -1103,6 +1195,57 @@ function colimit(d::BipartiteFreeDiagram{<:FinSet{Int}})
...
+  n_var = VarSet{T}(length(roots)) # number of resulting variables
+  # Construct maps into apex
+  csp = Multicospan(n_var, map(var_indices[(1:n₂).+n₁]) do v_is 
+    VarFunction{T}(map(v_is) do v_i 
+      r = find_root!(vars, v_i)
+      haskey(concretes,r) ? concretes[r] : AttrVar(findfirst(==(r), roots))
+    end, FinSet(n_var.n))
+  end)
+  # Cocone diagram
+  return Colimit(d, csp) # FIXME why is this a Colimit{FinSet{Int}, Diagram{VarSet}} ?
```
epatters, line 1232: What's the story with this FIXME? Is there something to fix?

src/programs/DiagrammaticPrograms.jl
```diff
@@ -1103,6 +1123,9 @@ end
 ############################
 
 # Types that can be Julia literals.
+struct Lit 
+  val::Any
+end
```
epatters, line 1128: Is this struct used anywhere? I don't see it being used.

test/acsets/CSetDataStructures.jl
```diff
@@ -242,7 +241,7 @@ for (dgram_maker, ldgram_maker) in dgram_makers
 
   # Allow type inheritance for data attributes.
   d_abs = dgram_maker(Number)
-  add_parts!(d_abs, :X, 2, height=[10.0, 4])
+  add_parts!(d_abs, :X, 2, height=Number[10, 4])
```
epatters, line 244: Is this a varacsets thing or a general problem with our acsets API? It's not in the spirit of Julia to be so strict about the types. A vector of integers should be implicitly treated as vector of numbers.

test/categorical_algebra/FinCats.jl
```diff
@@ -2,6 +2,7 @@ module TestFinCats
 using Test
 
 using Catlab, Catlab.Theories, Catlab.CategoricalAlgebra, Catlab.Graphs
+using Catlab.CategoricalAlgebra.FinSets: VarSet
```
epatters, line 5: Looks like some accidental/leftover edits to this file here and in the commented lines below.

test/programs/RelationalPrograms.jl
```diff
@@ -49,6 +49,12 @@ d = RelationDiagram(0)
 add_box!(d, 0, name=:A)
 @test parsed == d
 
+sird_uwd = @relation () where (S::Pop, I::Pop, R::Pop, D::Pop) begin
+  infect(S,I,I,I) # inf
+  disease(I,R) # recover
+  disease(I,D) # die
+end
+
```
epatters, line 57: Is this supposed to be here? I don't see any tests of the constructed diagram and it doesn't seem relevant to the PR.

epatters (changes requested):
It's great to see this shaping up! I left a final round of mostly minor comments below.

test/programs/RelationalPrograms.jl
kris-brown: The test here is just making sure that this doesn't error. 

In the course of making this PR I think there was a change which made constructing a UWD with no outer ports fail, apparently that problem has been resolved though.

test/acsets/CSetDataStructures.jl
kris-brown: I reverted the change and there was no error, so this is no longer a problem!

src/acsets/DenseACSets.jl
```diff
@@ -385,19 +406,29 @@ end
...
-  if !unbox_injective
-    preimage_multi(dom_parts(acs, f), acs.subparts[f], parts)
-  else
-    preimage_multi(dom_parts(acs, f), acs.subparts[f], parts, UnboxInjectiveFlag())
-  end
+"""
+Performance regression by avoiding preimage_multi, which does not work for 
+attributes.
+"""
+@inline function ACSetInterface.incident(acs::SimpleACSet,
```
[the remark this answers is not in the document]
kris-brown: I've made it clear now that this is something to be fixed in the future.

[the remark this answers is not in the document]
kris-brown: created an [issue](https://github.com/AlgebraicJulia/Catlab.jl/issues/776)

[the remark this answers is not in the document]
epatters: Perfect, thanks!

src/categorical_algebra/Chase.jl
```diff
@@ -0,0 +1,520 @@
...
+  return NamedTuple(init)
+end 
+
+# Colimit of ACset FinCats
+##########################
+"""
+Preserves the original name of the inputs if it is unambiguous, otherwise
+disambiguates with index in original input. E.g. (A,B)⊔(B,C) → (A,B#1,B#2,C)
+"""
+function coproduct(Xs::AbstractVector{<: FinCatPresentation{ThSchema}}; kw...)
```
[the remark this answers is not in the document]
kris-brown: This thread lives on in this issue https://github.com/AlgebraicJulia/Catlab.jl/issues/778

src/categorical_algebra/Chase.jl
```diff
@@ -0,0 +1,520 @@
...
+  end
+  new_codom = FinCat(p)
+  ls = map(legs(C)) do l
+    FinFunctor(l.ob_map, l.hom_map, dom(l), new_codom)
+  end
+  Colimit(DiscreteDiagram(Xs), Multicospan(ls)) => p
+end
+
+
+# Extending morphisms (should this be in CSets.jl)
```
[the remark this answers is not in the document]
epatters: This is issue #774.

epatters (approved):

PR #748, opened 2023-03-15
Acset transformations between dynamic acsets

kris-brown:
This extends the Struct vs Dynamic acset paradigm to include morphisms. Many features are still not covered for DynamicACSet categories, but this is a starting point. The goal of this PR was to be able to construct and compose maps between DynamicACSets.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -1,7 +1,8 @@
 """ Categories of C-sets and attributed C-sets.
 """
 module CSets
-export ACSetTransformation, CSetTransformation,
+export ACSetTransformation, CSetTransformation,StructACSetTransformation,
+  StructTightACSetTransformation,
   TightACSetTransformation, LooseACSetTransformation, SubACSet, SubCSet,
   ACSetHomomorphismAlgorithm, BacktrackingSearch, HomomorphismQuery,
   components, force, is_natural, homomorphism, homomorphisms, is_homomorphic,
@@ -66,6 +67,8 @@ end
 """
 @inline TypeSet(::StructACSet{S,Ts}, type::Symbol) where {S,Ts} =
   type_set(Val{S}, Val{Ts}, Val{type})
+
+TypeSet(X::DynamicACSet, type::Symbol) = runtime(type_set, X.schema, X.type_assignment,type)
 
 @ct_enable function type_set(@ct(S), @ct(Ts), @ct(type))
   @ct begin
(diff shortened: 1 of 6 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -224,48 +227,36 @@ acsets, loose morphisms are usually preferable. For more information about
...
-    error("Not all names in $(keys(components)) are objects or attribute types")
-  if isempty(type_components)
-    TightACSetTransformation{S}(ob_components, X, Y)
-  else
-    LooseACSetTransformation{S}(ob_components, type_components, X, Y)
-  end
-end
+abstract type ACSetTransformation end
+abstract type TightACSetTransformation <:ACSetTransformation end
+abstract type LooseACSetTransformation <:ACSetTransformation end
```
epatters, line 232: These three top-level types should each have docstrings because they are exported and because most users will interact with these rather than specific concrete subtypes.

src/categorical_algebra/CSets.jl
```diff
@@ -224,48 +227,36 @@ acsets, loose morphisms are usually preferable. For more information about
...
 force(α::ACSetTransformation) = map_components(force, α)
 
-""" Transformation between C-sets.
 
-Recall that a C-set homomorphism is a natural transformation: a transformation
-between functors C → Set satisfying the naturality axiom for every morphism, or
-equivalently every generating morphism, in C.
+# Dynamic ACSet transformations
+@struct_hash_equal struct DynamicTightACSetTransformation <: TightACSetTransformation
+  components::NamedTuple
```
epatters, line 241: Not the most important thing, but IMO the most natural type for `components` would be `AbstractDict{Symbol}` rather than `NamedTuple` because in the latter, the key names are static (contained in a type parameter of `NamedTuple`).

src/categorical_algebra/CSets.jl
```diff
@@ -305,19 +322,21 @@ end
 coerce_component(ob::Symbol, f, dom_size::Int, codom_size::Int) =
   FinFunction(f, dom_size, codom_size)
 
-function Base.getindex(α::TightACSetTransformation{S}, c) where S
+# Other
```
epatters, line 325: Did this comment `# Other` creep in by accident?

epatters (changes requested):
Nice and clean! I left a few comments.

src/categorical_algebra/CSets.jl
kris-brown: There are some issues with code that is current generic w/r/t `Struct` vs `Loose` that gets broken if this is changed to an `AbstractDict`, so perhaps this can be changed in a later PR?

epatters: Postponed to #750

epatters (approved):

PR #752, opened 2023-03-24
Move to organization level julia ci/cd

mehalter:


epatters (changes requested):
Thanks Micah, this will be great!

Your code editor seems to have auto-reformatted the code in several files. I am very deliberate about how I format code and generally like it the way it is. Can you please revert these changes?

mehalter:
Ah yeah @epatters I was just using the built in formatting commands from the Julia language servers and the standard Julia formatting specification. I'll revert them and just do the changes that I made to fix the 32 bit support

mehalter:
One that gets changed, is this good to merge? Want to have time to get this merged in tomorrow morning if possible

mehalter:
This will be an active blocker getting any other AlgebraicJulia package moved over to the new setup since they will need these fixes to be 32 bit compatible

mehalter:
We might want to look into formalizing the formatting of AlgebraicJulia packages to avoid this in the future and also clean up our code. This will make it easier for outside collaborators to contribute as well. There are currently a lot of mixed format standards such as inconsiderate spacing of parameters/list items and inconsistent uses of single/double quotes are a couple that the formatter seems to really pick up on. The biggest thing we would have to figure out is probably just disabling indentation checking by a formatter since the indentation used in Catlab isn't able to be done programmatically.

epatters:
Yes, once the formatting changes are reverted, this is good to merge. Thanks Micah!

PR #753, opened 2023-03-25
Colimits of dynamic acsets

kris-brown:
This PR makes minimal changes in order to get working pushouts for dynamic acsets. `@ct_enable` is applied to `pack_colimit`,`colimit_attrs!`, and the top-level `colimit` call for a `DynamicACSet` first looks for an object in the diagram and then gets type information from that. For example:

```
diagram_ob(m::Multispan{DynamicACSet}) = apex(m)
diagram_ob(m::Multicospan{DynamicACSet}) = apex(m)
# etc., add as needed
```
Right now only these diagrams are covered, but to support general colimits, every type that can serve as a diagram should be covered.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -980,7 +980,7 @@ end
 
 
 function colimit(::Type{<:Tuple{DynamicACSet,TightACSetTransformation}}, diagram) 
-  X = first(cocone_objects(d))
+  X = first(cocone_objects(diagram))
   S = acset_schema(X)
   ACS = ()->DynamicACSet(X.name,S)
   colimits = map(colimit, unpack_diagram(diagram))
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -34,6 +34,7 @@ import ..Subobjects: Subobject, implies, ⟹, subtract, \, negate, ¬, non, ~
 import ..Sets: SetOb, SetFunction, TypeSet
 import ..FinSets: FinSet, FinFunction, FinDomFunction, force, predicate, is_monic, is_epic
 import ..FinCats: FinDomFunctor, components, is_natural
+using ...DenseACSets: indices, unique_indices
 
 # Sets interop
 ##############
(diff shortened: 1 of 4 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -102,6 +102,9 @@ For morphisms, the result is a `FinFunction`; for attributes, a
 @inline SetFunction(X::StructACSet{S}, name::Symbol) where {S} =
   set_function(X, Val{S}, Val{name})
 
+@inline SetFunction(X::DynamicACSet, name::Symbol)  =
```
epatters, line 105: This one shouldn't be inlined. The inlining is only required to propagate constants through `Val` types.

src/categorical_algebra/CSets.jl
```diff
@@ -121,6 +124,10 @@ Indices are included whenever they exist.
 @inline FinFunction(X::StructACSet{S}, name::Symbol) where {S} =
   set_function(X, Val{S}, Val{name})
 
+@inline FinFunction(X::DynamicACSet, name::Symbol)  =
```
epatters, line 127: Ditto.

src/categorical_algebra/CSets.jl
```diff
@@ -1083,6 +1106,22 @@ function unpack_components(αs::AbstractVector{<:StructACSetTransformation{S}};
   NamedTuple(c => map(α -> α[c], αs) for c in names)
 end
 
+"""
+Every (non-empty) diagram has a representative object we can use to get  
+dynamic type information from
+"""
+diagram_ob(m::Multispan{DynamicACSet}) = apex(m)
```
epatters, line 1113: Since it doesn't matter which object this is (as long as the user hasn't mixed up their acsets!), I think you just define this as `first(cocone_objects(d))` for all colimits, which would fail only for empty diagrams.

src/categorical_algebra/CSets.jl
```diff
@@ -985,30 +1001,37 @@ function colimit(::Type{<:Tuple{ACS,LooseACSetTransformation}}, diagram;
...
-                              type_components=type_components), Xs)
+  colimit_attrs!(pack_colimit(Val{ColimitACS}, Val{S}, diagram, Xs, colimits, type_components), Xs)
 end
 
-""" Make colimit of acsets from colimits of sets, ignoring attributes.
-"""
-function pack_colimit(::Type{ACS}, diagram, Xs, colimits; kw...) where
-    {S, ACS <: StructACSet{S}}
+#""" Make colimit of acsets from colimits of sets, ignoring attributes."""
+@ct_enable function pack_colimit(@ct(ACS), @ct(S), diagram, Xs, colimits, type_components)#; kw...) where {S, ACS <: StructACSet{S}}
```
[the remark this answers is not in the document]
olynch: Yeah, I'd be inclined to agree.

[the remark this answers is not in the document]
kris-brown: thanks, that's fixed now!

olynch:
:fireworks: Woohoo!

PR #760, opened 2023-04-05
Added type annotations to acset part mutators

### LINKED ISSUE #728: Unexpected error message when adding parts
```
g = Graph 
add_part!(g, :V)
```
this is a user error (should have been `Graph()`) results in a `StackOverflowError` in contrast to things like `rem_part!` (which results in a method not found error). If there's any way to avoid this it would make things more user friendly!

KevinArlin:
Fixed #728  by requiring that the parameter `acs` of acset part mutators be an actual `ACSet`.

### DIFF
--- a/src/acsets/ACSetInterface.jl
+++ b/src/acsets/ACSetInterface.jl
@@ -140,7 +140,7 @@ Returns the ID of the added part.
 
 See also: [`add_parts!`](@ref).
 """
-@inline function add_part!(acs, type::Symbol, kw)
+@inline function add_part!(acs ::ACSet , type::Symbol, kw)
   part = only(add_parts!(acs,type,1))
   try
     set_subparts!(acs, part, kw)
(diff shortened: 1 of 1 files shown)


src/acsets/ACSetInterface.jl
```diff
@@ -140,7 +140,7 @@ Returns the ID of the added part.
 
 See also: [`add_parts!`](@ref).
 """
-@inline function add_part!(acs, type::Symbol, kw)
+@inline function add_part!(acs ::ACSet , type::Symbol, kw)
```
epatters, line 143: The usual style in Julia is to not put spaces on either side of `::`, e.g., this should line be `acs::ACSet`.

epatters (changes requested):
Thanks Kevin! I assume you checked that this resolves the stack overflow reported in the original issue?

KevinArlin:
> Thanks Kevin! I assume you checked that this resolves the stack overflow reported in the original issue?

Yep, this replaces Kris's stack overflow with a ```MethodError```.

epatters (approved):
Awesome, thanks!

PR #762, opened 2023-04-11
Improve finite functor error messages

### LINKED ISSUE #598: Improved error message on finfunctor failure

KevinArlin:
Fixes #598 by having `@finfunctor` tell the user which morphisms are mapped inconsistently with the object map. There is code ready to update `@finfunctor` to check preservation of equations between morphisms in the domain category, too, once we have some ability to check equality of homs.

### DIFF
--- a/src/programs/DiagrammaticPrograms.jl
+++ b/src/programs/DiagrammaticPrograms.jl
@@ -263,7 +263,7 @@ function parse_functor(C::FinCat, D::FinCat,check_equations::Bool, ast::AST.Mapp
   ob_map, hom_map = make_ob_hom_maps(C, ast)
   F = FinFunctor(mapvals(x -> parse_ob(D, x), ob_map),
                  mapvals(f -> parse_hom(D, f), hom_map), C, D)
-  failures = get_nonfunctorialities(F,check_equations=check_equations)
+  failures = functoriality_failures(F,check_equations=check_equations)
   if !all(isempty,failures)
     if !check_equations
       doms, cods = failures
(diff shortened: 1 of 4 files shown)


olynch:
@mehalter @KevinArlin, what's our feeling about the failing tests on nightly? Is this a bug in nightly? Should we disable nightly CI until it's fixed? I don't want to merge this with failing tests.

src/programs/DiagrammaticPrograms.jl
```diff
@@ -1153,4 +1185,18 @@ function destructure_unary_call(expr::Expr)
...
 end
 
+"""
+Return the right-hand side of the assignment in an expression of the form
+`:(var=val)`.
+"""
+function get_keyword_arg_val(expr::Expr)
+  @match expr begin
+    Expr(:(=),var,x) => x
+    Expr(_...) => error("""
```
olynch, line 1195: This can just be `_`, not `Expr(_...)`

olynch (approved):
Overall looks good; I saw one minor thing that you might want to change but it doesn't really matter it's just something good to know.

epatters (changes requested):
Thanks Kevin! In addition to Owen's comment, I have a few small comments too.

Beyond that, the main thing that's bothering me is that naturality and functorality testing now have completely different APIs: the former uses a keyword argument `return_failures` and the latter has separate function `get_nonfunctoralities`.

It doesn't have to be in this PR, but before doing a release I'd like to make this consistent. My suggestion would be to have no keyword arguments and use:

- `is_natural` and `naturality_failures`
- `is_functorial` and `functorality_failures`

I'm open to other conventions if you have ideas. I would suggest not prefixing with `get_` since this is generally (though not always) avoided in Julia APIs.

epatters:
Oh, @KevinArlin, I see that #765 you also add a `get_unnaturalities` function. I think I'd prefer a different naming convention but, more importantly, this does address my complaint about having inconsistent APIs for naturality and functorality testing.

epatters (approved):

PR #765, opened 2023-04-21
New macro `@acset_transformation`

### LINKED ISSUE #623: Better error message when unnatural homomorphisms get used
It can be confusing when a new user gets `Quotient map of colimit is ill-defined` nested deep in colimit internals as an error, when often the cause of this is passing in an unnatural transformation to something which invokes colimits (e.g. DPO rewriting). Would it make sense for some higher-level function to catch this error and then test user input for naturality? (returning the original error if the inputs were natural, but otherwise a better error message saying which input was unnatural)

KevinArlin:
Closes #623 by defining macros @acset_transformation(s) that provide guardrails and syntax for construction of AcsetTransformations. Also refactors `is_natural` to be type-stable.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -693,6 +693,8 @@ either Julia functions on the appropriate types or FinFunction(Dict(...)).
 
 Use the keyword argument `alg` to set the homomorphism-finding algorithm. By
 default, a backtracking search algorithm is used ([`BacktrackingSearch`](@ref)).
+Use the keyword argument error_failures = true to get errors explaining 
+any immediate inconsistencies in specified initial data.
 
 See also: [`homomorphisms`](@ref), [`isomorphism`](@ref).
 """
@@ -706,44 +708,6 @@ function homomorphism(X::ACSet, Y::ACSet, alg::BacktrackingSearch; kw...)
     result = isa(α,ACSetTransformation) ? α : nothing; return true
   end
   result
-end
-
-"""
-Searches for a homomorphism,  return inconsistencies in the givens
-if the search fails before recursion starts. TODO: partialacsettransformation
-type to allow checking for naturality of initial assignment?
-"""
-function homomorphism_error_failures(X,Y;kw...)
-  result = nothing
-  backtracking_search(X,Y;kw...) do α
-      result = α; return isa(α,ACSetTransformation)
-  end 
-  if isa(result,Tuple)
-    isoFailures, monoFailures = result
-    isoFailures,monoFailures = collect(isoFailures),collect(monoFailures)
-    error("""
-          Cardinalities inconsistent with request for...
-            iso at object(s) $isoFailures
-            mono at object(s) $monoFailures
-          """)
-  end
-  if isa(result,Dict)
-    error(show_unnaturalities(result))
-  end
-  if isnothing(result)
-    error("""
-          No homomorphism exists extending given data.
-          """)
-  end
-  if result == false return nothing end
-  return result
-end
-#TODO: This is currently up to twice as slow as `homomorphisms`, in the case 
-#when there's exactly one solution. Needs refactoring of 
-#`backtracking_search`, which is like a whole dang thing right now.
-function homomorphisms_error_failures(X,Y;kw...)
-  homomorphism_error_failures(X,Y;kw...)
-  homomorphisms(X,Y;kw...)
 end
 
 """ Find all homomorphisms between two attributed ``C``-sets.
@@ -839,9 +803,9 @@ struct BacktrackingState{
   type_components::LooseFun
 end
 
-#f needs to return bool
 function backtracking_search(f, X::ACSet, Y::ACSet;
-    monic=false, iso=false, random=false, type_components=(;), initial=(;))
+    monic=false, iso=false, random=false, 
+    type_components=(;), initial=(;), error_failures=false)
   S, Sy = acset_schema.([X,Y])
   S == Sy || error("Schemas must match for morphism search")
   Ob = Tuple(objects(S))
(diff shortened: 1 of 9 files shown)


test/categorical_algebra/CSets.jl
```diff
@@ -114,13 +114,17 @@ g, h = path_graph(Graph, 4), cycle_graph(Graph, 2)
 @test components(α′′) == components(α)
 
 # Naturality.
+d = get_unnaturalities(α)
+@test [collect(d[a]) for a in keys(d)] == [[],[]]
 @test is_natural(α)
 β = CSetTransformation((V=[1,2,1,2], E=[1,1,1]), g, h)
+d = get_unnaturalities(β)
+@test [collect(d[a]) for a in keys(d)] == [[(2,1,2)],[(2,2,1)]]
```
epatters, line 122: This test is failing some versions of Julia because the order of iteration of dictionary keys is not guaranteed to be stable between different versions. This should be:

```julia
sort([collect(v) for v in values(d)]) == [[(2,1,2)],[(2,2,1)]]
```

epatters (changes requested):

src/categorical_algebra/CSets.jl
```diff
@@ -538,30 +538,74 @@ For each hom in the schema, e.g. h: m → n, the following square must commute:
...
      αₙ 
-
-If `return_failures` is true, return a list of violations, with elements such as: 
-  (h, index i in Xₘ, Yₕ(αₘ(i)), αₙ(Xₕ(i)))
-
-This is type-unstable, so it should not be used in performance-sensitive code.
+You're allowed to run this on a named tuple partly specifying an ACSetTransformation,
+though at this time the domain and codomain must be fully specified ACSets.
+"""
+function is_natural(α::ACSetTransformation)
```
epatters, line 544: This function should dispatch on `TightACSetTransformatio` vs `LooseACSetTransformation` rather than doing a type check using `isa` in the body.

src/categorical_algebra/CSets.jl
```diff
@@ -538,30 +538,74 @@ For each hom in the schema, e.g. h: m → n, the following square must commute:
...
+    Xf,Yf,α_c,α_d = subpart(X,f),subpart(Y,f), α[c], α[d]
+    Pair(f,
+    Iterators.map(i->(i,Yf[α_c(i)],α_d(Xf[i])),
+      Iterators.filter(dom(α_c)) do i
+        Xf[i] in dom(α_d) && Yf[α_c(i)] != α_d(Xf[i])
+      end))
+  end
+  Dict(ps)
+end
+function naturality_failures(α::ACSetTransformation)
```
epatters, line 582: Ditto for this function.

src/categorical_algebra/CSets.jl
```diff
@@ -538,30 +538,74 @@ For each hom in the schema, e.g. h: m → n, the following square must commute:
...
+      if the given nat is α:X -> Y, f's naturality square 
+      fails to commute at i ∈ X(c), 
+      with Y(f)(α_c(i))=j and α_d(X(f)(i))=k.
+      
+
+      """ 
+  for f in keys(d)
+    isempty(d[f]) || begin
+      failures = collect(d[f])
+      s *= reduce(*,[["$f: "];["$failure" for failure in failures];["\n"]])
```
epatters, line 602: More idiomatic to use `join` rather than `reduce(*, ...)` here.

src/categorical_algebra/CSets.jl
```diff
@@ -643,20 +686,26 @@ that the vertex map is injective but imposes no constraints on the edge map.
...
-homomorphism(X::ACSet, Y::ACSet; alg=BacktrackingSearch(), kw...) =
+homomorphism(X::ACSet, Y::ACSet; alg=BacktrackingSearch(), kw...) = begin 
   homomorphism(X, Y, alg; kw...)
+end
 
 function homomorphism(X::ACSet, Y::ACSet, alg::BacktrackingSearch; kw...)
   result = nothing
   backtracking_search(X, Y; kw...) do α
-    result = α; return true
+    result = isa(α,ACSetTransformation) ? α : nothing; return true
```
epatters, line 708: Why is this check needed? My guess is that this is leftover from a previous design.

src/categorical_algebra/CSets.jl
```diff
@@ -672,7 +721,7 @@ homomorphisms(X::ACSet, Y::ACSet; alg=BacktrackingSearch(), kw...) =
 function homomorphisms(X::ACSet, Y::ACSet, alg::BacktrackingSearch; kw...) 
   results = []
   backtracking_search(X, Y; kw...) do α
-    push!(results, map_components(deepcopy, α)); return false
+    if isa(α,ACSetTransformation) push!(results, map_components(deepcopy, α)) end; return false
```
epatters, line 724: Same question here.

src/categorical_algebra/CSets.jl
```diff
@@ -774,16 +824,25 @@ function backtracking_search(f, X::ACSet, Y::ACSet;
...
   end
-  for c in iso
-    nparts(X,c) == nparts(Y,c) || return false
-  end
   if monic isa Bool
     monic = monic ? Ob : ()
   end
-  # Injections between finite sets are bijections, so reduce to that case.
+  isoFailures = Iterators.filter(c->nparts(X,c)!=nparts(Y,c),iso)
+  monoFailures = Iterators.filter(c->nparts(X,c)>nparts(Y,c),monic)
```
epatters, line 831: Our convention is to use snake case for variable names, so these should be `iso_failures` and `mono_failures`.

src/categorical_algebra/FinCats.jl
```diff
@@ -65,6 +65,7 @@ Because object generators usually coincide with objects, the default method for
 function ob_generator end
 
 ob(C::FinCat, x) = ob_generator(C, x)
+hom(C::FinCat,x::Nothing) = x
```
epatters, line 68: Not clear why this line was added and it's not being called in the tests, so just delete it?

src/categorical_algebra/FinCats.jl
```diff
@@ -220,6 +219,7 @@ compose(C::FinCatPathGraph, fs...) =
   reduce(vcat, coerce_path(graph(C), f) for f in fs)
 
 hom(C::FinCatPathGraph, f) = coerce_path(graph(C), f)
+hom(C::FinCatPathGraph, f::Nothing) = nothing
```
epatters, line 222: Ditto for this line, and several others below. Please check this.

src/categorical_algebra/FinCats.jl
```diff
@@ -239,7 +239,6 @@ end
 FinCatGraph(g::HasGraph) = FreeCatGraph(g)
 
 is_free(::FreeCatGraph) = true
-equations(::FreeCatGraph) = []
```
epatters, line 242: Did you mean to delete this line, and the other one below? I'm surprised this didn't cause an error. Our test coverage of these might be incomplete.

epatters (changes requested):
Thanks Kevin! Some minor comments below.

src/categorical_algebra/FinCats.jl
KevinArlin: Oops, yeah, these were probably from a draft that used nothing where I really ended up using GatExpr{:nothing}

epatters (approved):
Thanks Kevin! Merging now.

PR #768, opened 2023-05-01
Cascading delete in acsets

kris-brown:
Removing a part with `recurse=true` removes all incident parts recursively

This can be redone much nicer after ACSets support 'mark as deleted'. Because deleting a single part changes the ids of other parts, a full pass must be made to find everything that must be deleted and then the deletion follows afterwards.


kris-brown:
Since Variables can in principle be dangling, it's unclear what this should do once this PR combines with the VarACSets PR

PR #775, opened 2023-05-13
Use native package extensions in Julia 1.9

mehalter:
This PR moves Catlab to using only built in package extensions rather than having a dependency on Requires.jl.


mehalter:
Actually, reverting this back to a draft for now. Found some issues with this hacky approach to multiple package dependencies 😢

epatters (approved):
Thanks Micah, this is great!

I forgot how janky the Graphviz JLL is. I tested that locally and confirmed that it returns the right path when the JLL is loaded.

I didn't test the SCS extension but based on your comment, it sounds like you did.

PR #777, opened 2023-05-16
Maximum common sub-C-set

mehalter:
Implements a general purpose algorithm for calculating Most Common Sub C-Set

### DIFF
--- a/src/acsets/DenseACSets.jl
+++ b/src/acsets/DenseACSets.jl
@@ -516,18 +516,13 @@ Note: the correctness is dependent on the implementation details of `rem_parts!`
 """
 function delete_subobj!(X::ACSet, delparts)
   dels = delete_subobj(X, delparts)
-  return NamedTuple(Dict(map(ob(acset_schema(X))) do o
+  return NamedTuple(map(ob(acset_schema(X))) do o
     ps = collect(parts(X,o))
     rem_parts!(X, o, dels[o])
     return o => map(parts(X,o)) do i 
-      if i ∈ dels[o]
-        return pop!(ps) # if idx is deleted, it's replaced with last idx
-      else 
-        return i 
-      end
+      return i ∈ dels[o] ? pop!(ps) : i
     end
-  end))
-  dels
+  end)
 end
 
 ACSetInterface.cascading_rem_parts!(acs::ACSet, type, parts) =
--- a/src/acsets/DenseACSets.jl
+++ b/src/acsets/DenseACSets.jl
@@ -509,10 +509,22 @@ function delete_subobj(X::ACSet, delparts)
   return Dict([k => sort(collect(v)) for (k,v) in pairs(delparts)])
 end
 
+"""Return a mapping of from parts of updated X to the old X"""
 function delete_subobj!(X::ACSet, delparts)
-  for (type, parts) in delete_subobj(X, delparts)
-    rem_parts!(X, type, parts)
-  end 
+  dels = delete_subobj(X, delparts)
+  return NamedTuple(Dict(map(ob(acset_schema(X))) do o
+    ps = collect(parts(X,o))
+    rm = rem_parts!(X, o, dels[o])
+    res = map(parts(X,o)) do i 
+      if i ∈ dels[o]
+        return pop!(ps)
+      else 
+        return i 
+      end
+    end
+    return o => res
+  end))
+  dels
 end
 
 ACSetInterface.cascading_rem_parts!(acs::ACSet, type, parts) =
--- a/src/acsets/DenseACSets.jl
+++ b/src/acsets/DenseACSets.jl
@@ -509,20 +509,23 @@ function delete_subobj(X::ACSet, delparts)
   return Dict([k => sort(collect(v)) for (k,v) in pairs(delparts)])
 end
 
-"""Return a mapping of from parts of updated X to the old X"""
+"""
+Return a mapping of from parts of updated X to the old X
+
+Note: the correctness is dependent on the implementation details of `rem_parts!`
+"""
 function delete_subobj!(X::ACSet, delparts)
   dels = delete_subobj(X, delparts)
   return NamedTuple(Dict(map(ob(acset_schema(X))) do o
     ps = collect(parts(X,o))
-    rm = rem_parts!(X, o, dels[o])
-    res = map(parts(X,o)) do i 
+    rem_parts!(X, o, dels[o])
+    return o => map(parts(X,o)) do i 
       if i ∈ dels[o]
-        return pop!(ps)
+        return pop!(ps) # if idx is deleted, it's replaced with last idx
       else 
         return i 
       end
     end
-    return o => res
   end))
   dels
 end
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -1723,76 +1723,17 @@ If there are attributes, we ignore these and use variables in the apex of the
 overlap.
 """
 function mca(Xs::Vector{T}) where T <: ACSet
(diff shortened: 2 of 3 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -1429,6 +1433,271 @@ end
...
+  Y == codom(f) ? top(dom(f)) : error("Cannot apply inverse of $f to $Y")
+
+# VarACSets
+###########
+
+"""
+For any ACSet, X, a canonical map A→X where A has distinct variables for all
+subparts.
+"""
+function abstract(X::ACSet)
```
epatters, line 1478: I'm actually surprised that `abstract` isn't a reserved word in Julia (I guess only `abstract type` is) but in any case this name seems a bit too generic. Maybe `abstract_attributes`?

src/categorical_algebra/CSets.jl
```diff
@@ -1429,6 +1433,271 @@ end
...
+  S = acset_schema(X)
+  A = deepcopy(X);
```
epatters, line 1480: Not very important, but this need only be `A = copy(X)` since no mutable data potentially stored in attributes with be modified.

src/categorical_algebra/CSets.jl
```diff
@@ -1429,6 +1433,271 @@ end
...
+""" Compute the size of a C-Set
+
+Defintion: let 𝐺: C → 𝐒et be a C-set, we define the _size_ of 𝐺 to be ∑_{c ∈ C}
+|𝐺c|.  For example, under this definition, the size of:
+  * a graph G is |GE| + |GV| (num edges + num vertices)
+  * a Petri net P is |PT| + |PS| + |PI| + |PO| (num transitions + num species +
+    num input arcs + num output arcs).
+"""
+size(X::ACSet) = mapreduce(oₛ -> nparts(X, oₛ), +, objects(acset_schema(X)); init=0)
+size(X::ACSetTransformation) = size(dom(X))
```
epatters, line 1507: I guess it's good that these aren't exported since, by analogy with multidimensional arrays, I would expect `Base.size` to do something different on acsets. But I'm fine with this as long as we aren't committing to it as a public interface.

src/categorical_algebra/CSets.jl
```diff
@@ -1429,6 +1433,271 @@ end
...
+
+"""
+Enumerate subobjects of an ACSet in order of largest to smallest 
+(assumes no wandering variables).
+"""
+struct SubobjectIterator
+  top::ACSet
+end
+
+function Base.collect(it::SubobjectIterator)
```
epatters, line 1517: Are you sure that this method needs to be implemented? I would expect `collect` to work automatically for any iterable, and the [iterators interface](https://docs.julialang.org/en/v1/manual/interfaces/) doesn't mention it.

src/categorical_algebra/CSets.jl
```diff
@@ -1429,6 +1433,271 @@ end
...
+monic span of ACSets ``a₁ ← a → a₂``. There may be many maps from this overlap 
+into each of the inputs, so a Vector{Vector{ACSetTransformations}} per overlap 
+is returned (a cartesian product can be taken of these to obtain all possible 
+multispans for that overlap). If there are multiple overlaps of equal size, 
+these are all returned.
+
+If there are attributes, we ignore these and use variables in the apex of the 
+overlap.
+"""
+function mca(Xs::Vector{T}) where T <: ACSet
```
epatters, line 1687: Can we call this `maximum_common_subobject`? It's a longer name but `mca` is cryptic.

epatters (changes requested):
Really nice PR! I especially appreciate the thorough tests.

I left a few small comments below.

src/acsets/DenseACSets.jl
```diff
@@ -509,10 +509,25 @@ function delete_subobj(X::ACSet, delparts)
...
+Return a mapping of from parts of updated X to the old X
+
+Note: the correctness is dependent on the implementation details of `rem_parts!`
+"""
 function delete_subobj!(X::ACSet, delparts)
-  for (type, parts) in delete_subobj(X, delparts)
-    rem_parts!(X, type, parts)
-  end 
+  dels = delete_subobj(X, delparts)
+  return NamedTuple(Dict(map(ob(acset_schema(X))) do o
```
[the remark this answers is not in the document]
mehalter: I resolved this I believe, please take a look at the new code. Basically just removes the `Dict` creation, should keep the elements in the expected order when creating the NamedTuple.

src/acsets/DenseACSets.jl
```diff
@@ -509,10 +509,25 @@ function delete_subobj(X::ACSet, delparts)
...
+    rem_parts!(X, o, dels[o])
+    return o => map(parts(X,o)) do i 
+      if i ∈ dels[o]
+        return pop!(ps) # if idx is deleted, it's replaced with last idx
+      else 
+        return i 
+      end
+    end
+  end))
+  dels
```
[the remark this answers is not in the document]
mehalter: Resolved in latest commit.

src/categorical_algebra/CSets.jl
```diff
@@ -1429,6 +1433,271 @@ end
     end...))
 end
 
+
+"""A map f (from A to B) as a map of subobjects of A to subjects of B"""
+(f::ACSetTransformation)(X::SubACSet) = begin
+  codom(hom(X)) == dom(f) || error("Cannot apply $f to $X")
```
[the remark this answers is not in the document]
mehalter: Resolved in latest commit.

src/categorical_algebra/CSets.jl
```diff
@@ -1429,6 +1433,271 @@ end
...
+# i.e. get the image of f as a subobject of B
+"""
+(f::ACSetTransformation)(X::StructACSet) =
+  X == dom(f) ? f(top(X)) : error("Cannot apply $f to $X")
+
+"""    preimage(f::ACSetTransformation,Y::Subobject)
+Inverse of f (from A to B) as a map of subobjects of B to subjects of A.
+"""
+function preimage(f::ACSetTransformation,Y::SubACSet)
+  codom(hom(Y)) == codom(f) || error("Cannot apply $f to $X")
```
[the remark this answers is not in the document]
mehalter: Resolved in latest commit.

src/categorical_algebra/CSets.jl
mehalter: Resolved in latest commit.

src/categorical_algebra/CSets.jl
mehalter: Resolved in latest commit.

src/categorical_algebra/CSets.jl
mehalter: I looked into this and it appears that there is a much cleaner way of doing this. I have fixed our iterator definition to properly use the Iterator interface. Basically we just needed to tell the interface that the size is unknown. Apparently by default the iterators interface sets the size as "HasSize" but then if you try to get the size it says "size unknown". This seems like a pretty bad default for "optional" fields but that's a separate conversation outside of this repo.

src/categorical_algebra/CSets.jl
mehalter: Yeah I agree that this very much shouldn't be exported. Do you have a different name that might make more sense in here just to make it clear that this *isn't* supposed to be `Base.size`?

mehalter: I'm going to rename this to `total_parts`

epatters: Seems reasonable. Thanks!

epatters (approved):

PR #783, opened 2023-05-22
Fix implicit conversion for union types with `AttrVar`

mehalter:


epatters (changes requested):
Thanks Micah! Can you please add a test case that would fail without this fix?

mehalter:
@epatters I updated one of the tests to test this. I also verified that with the test updates the tests fail on the current `master` branch

epatters (approved):
Thanks Micah!

PR #784, opened 2023-05-23
Add check if review checklist has already been commented

mehalter:
I'm pretty sure this is correct, sadly I have no way of testing this before it gets merged to `master` 😅


epatters (approved):
Thanks Micah, I guess we'll find out!

PR #785, opened 2023-05-24
delete_subobj! bugfix

kris-brown:
delete_subobj! deletes part of an ACSet while returning the data needed to construct a morphism from the new into the old ACSet. This requires simulating how the indices change, and through more thorough testing I found an unnatural morphism being produced from the old code. 

This new version works for the new examples, and it appears to be more correct.


mehalter (approved):
LGTM!

epatters (approved):
Thanks Kris!

PR #786, opened 2023-05-24
Subobject classifier + Internal hom for C-Sets

kris-brown:
Combines subacset search with computation of representables.

Tested on DDS (with equations to make the repr finite), Grph, and ReflGrph.

A change to SigmaMigration was needed because the previous implementation threw away the diagram map data (*how* the input CSet is related to the output CSet). Maybe should return a proper DiagramHom, but for now it's just the result ACSet + a dictionary of the diagram map components.

Shoutout to @KevinArlin for helping me work out where the homs go in the subobject classifier + internal hom constructions!


src/categorical_algebra/CSets.jl
```diff
@@ -1621,6 +1621,16 @@ function Base.push!(S::SubobjectIteratorState,h::ACSetTransformation)
   push!(S.seen, h); push!(S.to_recurse, h); push!(S.to_yield, h)
 end
 
+"""This would be much more efficient with canonical isomorph"""
+function seen(S::SubobjectIteratorState, f::ACSetTransformation)
```
epatters, line 1625: A small thing, but I like the Julia convention of prefacing the names of predicates with `is` or `has`, so this might be `is_seen`.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -305,7 +313,13 @@ function (F::SigmaMigration)(d::ACSet; n=100)
...
     end
   end
-  return res
+  diagram_map = Dict(map(types(S)) do o
+    s, t= add_srctgt("α_$o")    
+    m = last.(sort(collect(zip([rel_res[x] for x in [s,t]]...))))
+    ff = o ∈ ob(S) ? FinFunction : VarFunction{attrtype_type(D,o)}
+    o => ff(m, nparts(rel_res, nameof(ob_map(F.F⋅i2,o))))
+  end)
+  return res, diagram_map
```
epatters, line 322: @KevinArlin and I had a discussion this morning about how we are going to refactor the types around data migration. We don't need to worry about it for this PR, but just heads up that the API, including return conventions, for the migrations will change. Luckily no except us should depend on sigma migrations since until recently it only worked in a very special case.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -341,6 +355,84 @@ to recover the presentation from the datatype. Thus, this method for
...
+      findfirst(==(composite), maps[b])
+    end
+  end
+  return Fᴳ, homs
+end
+
+"""
+Any hom f: A->B induces a map from the representable of B to the repr of A
+"""
+function repr_map(S, reprs, f::Symbol)
```
epatters, line 429: `repr` is already a Julia function with an unrelated meaning. Would call this `representable_map` for clarity.

Or better, now that I think of it: since this is just the action of the Yoneda embedding on morphisms and that seems generally useful, it would be nice to expose this through a public API, perhaps as a method for the generic function `yoneda` if that makes sense.

test/categorical_algebra/DataMigrations.jl
```diff
@@ -318,12 +318,26 @@ X = @acset UndirectedBipartiteGraph begin
...
+  Dict(VG => VG, EG => EG, :Weight=>:Weight),
+  Dict(srcG => srcG, tgtG => tgtG, :weight=>:weight),
+  SchWeightedGraph, SchWeightedGraph
+)
+Y = @acset WeightedGraph{Symbol} begin 
+  V=2; E=2; Weight=1; src=1; tgt=[1,2]; weight=[AttrVar(1), :X] 
+end
+@test_throws ErrorException SigmaMigration(idF, WeightedGraph{Symbol}, WeightedGraph{Symbol})
+ 
+# @test ΣF(Y) == Y # TODO allow homs between attrtypes
```
epatters, line 340: Should this be an issue?

src/categorical_algebra/DataMigrations.jl
```diff
@@ -268,6 +273,9 @@ SigmaMigration(f,d::Type{T′},::Type{T}) where {T<:StructACSet, T′<:StructACS
 """
 Create a C-Set for the collage of the functor. Initialize data in the domain 
 portion of the collage, then run the chase.
+
+The true result data is a diagram morphism. We return the result ACSet and the 
+diagram map data as a dictionary with FinFunctions for each Ob.
```
epatters, line 278: I believe this is the unit of the adjunction between the sigma migration and delta migration functors. In particular, it is the data of an acset transformation on the domain schema C of the functor F: C -> D. We should document that.

Ideally we would return this as a diagram morphism from C-Set to D-Set whose functor part is F, rather than as a dict. Would that be impractical?

epatters (commented):
Thanks Kris! Wonderful to see how slickly we can now implement these features.

I left a few comments below.

Also, I would like to have a generic interface for subobject classifiers and internal homs, which we would then implement for FinSet as well as C-Set. However, that can wait for a future PR.

test/categorical_algebra/DataMigrations.jl
kris-brown: Created one! https://github.com/AlgebraicJulia/Catlab.jl/issues/788

test/categorical_algebra/DataMigrations.jl
```diff
@@ -318,21 +318,38 @@ X = @acset UndirectedBipartiteGraph begin
...
+idF = FinFunctor(
+  Dict(VG => VG, EG => EG, :Weight=>:Weight),
+  Dict(srcG => srcG, tgtG => tgtG, :weight=>:weight),
+  SchWeightedGraph, SchWeightedGraph
+)
+Y = @acset WeightedGraph{Symbol} begin
+  V=2; E=2; Weight=1; src=1; tgt=[1,2]; weight=[AttrVar(1), :X]
+end
+ΣF = SigmaMigrationFunctor(idF, WeightedGraph{Symbol}, WeightedGraph{Symbol})
+@test_skip ΣF(Y) == Y
```
epatters, line 344: @KevinArlin, I rebased this PR off of main, which includes #829. I expected that this test would now work, but it fails with an error. Can you please investigate?

KevinArlin: Yeah, I'll look into it.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -319,9 +319,8 @@ function (M::SigmaMigrationFunctor)(d::ACSet; n=100, return_unit::Bool=false)
   for (k,v) in datatypes(D)  atypes[Symbol(ob_map(i1,k))] = v end
   for (k,v) in datatypes(CD) atypes[Symbol(ob_map(i2,k))] = v end
   col_type = crel_type(presentation(apex(col)); types=atypes, name="Sigma")()
-  for o in ob(S)
+  for o in types(S)
     add_parts!(col_type, Symbol(ob_map(i1,o)), nparts(d,o))
-    #add parts for attrvars?
```
KevinArlin, line 318: The answer was "Yes". :)

test/categorical_algebra/DataMigrations.jl
KevinArlin: OK, that should be fixed. It looks like I thought of this attrvar issue briefly while writing this but then forgot to test on it.

src/categorical_algebra/DataMigrations.jl
epatters: Haha, great! Thanks for the fix.

KevinArlin: OK, fixed in a mathematically-correct, code-unprincipled way by borrowing @kris-brown's function for stripping free variables from AlgebraicRewriting. He said he might have a minute to think about how that function should really be rewritten and where it should go (in ACSets.jl, I suppose) but I'll do that if he doesn't get to it.

kris-brown: Why is `remove_freevars` needed here? The chase shouldn't introduce any free variables.

KevinArlin: When you migrate up to the collage, all the attributes are initially valued as attrvars, and then it's post-chase that we go back and re-impose naturality on attributes. That can leave some of those original attrvars floating.

PR #792, opened 2023-05-26
Breaking change to data migration API

KevinArlin:
This PR separates `DataMigration`s from `DataMigrationFunctor`s. The former is currently just a functor from a schema to another schema or a category of diagrams in (diagrams in) a schema. For upcoming functionality, a `DataMigration` `M` contains a dictionary `params` intended to contain some Julia functions describing the attributes `migrate(X,M)` should have in terms of the attributes of `X`. The semantic reason for this change is that a data migration is not the same thing as a functor--the functor tells you how to migrate, but isn't identical to the migration. But it was forced, practically, by the need for those `params`. The existing `QueryDiagram`s handle the need for `param`s just fine when we're doing, well, queries, but not when we're doing mutations, as I'm working on elsewhere right now.

In more detail, we have a new tree of types rooted at `AbstractDataMigration`, which always has a `func` method getting the underlying functor, with children `AbstractContravariantMigration` and `AbstractCovariantMigration`. Depending on the codomain of the underlying functor, an `AbstractContravariantMigration` might be (identical to, not a supertype of) a `DeltaSchemaMigration`, a `ConjSchemaMigration`, etc. The only `struct` subtyping `AbstractContravariantMigration` thus far is `DataMigration`, which contains a functor and the aforementioned dictionary of parameters. If the dictionary type's values are in `Union{}`, you have a `TotalDataMigration`, which might in particular be a `TotalDeltaMigration`, and which can be constructed using just a functor.

There's also some stuff for `SigmaMigration`s and `DataMigrationFunctor`s, where there's not too much going on of interest.

The `@migration` macros now always return a `DataMigration`, and all methods of `migrate` now expect an `AbstractDataMigration` rather than a `Functor`. It would be easy and might be good quality of life to add back in methods accepting just the underlying functor for `TotalDataMigration`s, but that won't be a breaking change so I've ignored it for now.

The trickiest thing to decide on here was how to handle the relationship between `DataMigration`s and `QueryDiagram`s, which are *almost* isomorphic structs. The only objective difference is that `QueryDiagram`s include a parameter identifying which diagram category they live in. We were originally planning to eliminate `QueryDiagram`s entirely and make even the output of `DiagrammaticPrograms.make_query` be a `DataMigration`, but that would require making `DataMigration`s carry around a variance parameter and some fiddling with implementing `limit` etc for `DataMigration`s, so I've left `QueryDiagram`s alive. Another option would be to just add the variance parameter so that `DataMigration==QueryDiagram``, and probably avoid all the new named types, but that seems unaesthetic and a bit disorganized to me. Happy for thoughts on this issue, though.

A side note: this PR involved the first error message I've ever seen that overflowed my entire terminal! That is, I had to pipe it to a text file to be able to scroll all the way to the top. These unergonomic error messages are a pretty nontrivial time cost. I think we've discussed and agreed that only Julia can do much about this issue, just wanted to note it again.

### DIFF
--- a/src/categorical_algebra/DataMigrations.jl
+++ b/src/categorical_algebra/DataMigrations.jl
@@ -2,7 +2,7 @@
 """
 module DataMigrations
 export DataMigration, SigmaMigration, DeltaMigration, SigmaMigrationFunctor, 
-  DeltaMigrationFunctor, DataMigrationFunctor, func, migrate, migrate!,  representable, yoneda, colimit_representables
+  DeltaMigrationFunctor, DataMigrationFunctor, functor, migrate, migrate!,  representable, yoneda, colimit_representables
 
 using ACSets
 using ACSets.DenseACSets: constructor, datatypes
@@ -11,6 +11,7 @@ using ...Theories: ob, hom, dom, codom, attr, AttrTypeExpr
 using ..Categories, ..FinCats, ..Limits, ..Diagrams, ..FinSets, ..CSets
 using ...Graphs, ..FreeDiagrams
 import ..Categories: ob_map, hom_map
+import ...Syntax: functor
 using ..FinCats: make_map, mapvals
 using ..Chase: collage, crel_type, pres_to_eds, add_srctgt, chase
 using ..FinSets: VarSet
@@ -51,9 +52,9 @@ duc, or gluc (contravariant) data migration. Alternatively, ``F`` may
 be a schema functor specifying a ``Σ`` migration in the covariant direction.
 """
 abstract type AbstractDataMigration{F<:FinDomFunctor} end
-func(M::AbstractDataMigration) = M.functor
-abstract type AbstractContravariantMigration{F<:FinDomFunctor} <: AbstractDataMigration{F} end
-abstract type AbstractCovariantMigration{F<:FinDomFunctor} <: AbstractDataMigration{F} end
+functor(M::AbstractDataMigration) = M.functor
+abstract type ContravariantMigration{F<:FinDomFunctor} <: AbstractDataMigration{F} end
+abstract type CovariantMigration{F<:FinDomFunctor} <: AbstractDataMigration{F} end
 
 """
 A contravariant data migration whose functor ``F`` may not be fully defined. Instead,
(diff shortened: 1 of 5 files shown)


src/categorical_algebra/DataMigrations.jl
```diff
@@ -41,102 +41,107 @@ a "duc query" (disjoint union of conjunctive queries).
...
+migration may be a pullback data migration ([`DeltaMigration`](@ref)), specified by a
+functor ``D → C`` between the schemas, or a conjunctive, gluing,
+duc, or gluc (contravariant) data migration. Alternatively, ``F`` may
+be a schema functor specifying a ``Σ`` migration in the covariant direction.
 """
-const DeltaSchemaMigration{D<:FinCat,C<:FinCat} = FinFunctor{D,C}
+abstract type AbstractDataMigration{F<:FinDomFunctor} end
+func(M::AbstractDataMigration) = M.functor
+abstract type AbstractContravariantMigration{F<:FinDomFunctor} <: AbstractDataMigration{F} end
+abstract type AbstractCovariantMigration{F<:FinDomFunctor} <: AbstractDataMigration{F} end
```
epatters, line 56: We don't have to prefix all abstract types with `Abstract`. I'd mildly prefer to shorten these to `ContravariantMigration` and `CovariantMigration`.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -242,37 +241,67 @@ function migrate(X::FinDomFunctor, F::GlucSchemaMigration)
...
-# Sigma migration
-#################
+""" Abstract type for a data migration functor.
 
-""" Sigma or left pushforward functorial data migration between acsets.
+This allows the user to have an actual model of the theory of 
+functors, once one knows the exact concrete types Dom and Codom
+of the relevant acsets. 
+"""
+abstract type MigrationFunctor{Dom<:ACSet,Codom<:ACSet} <:
```
epatters, line 250: It seems confusing that the top of the migration hierarchy is `AbstractDataMigration` and the top of the migration functor hierarchy is `MigrationFunctor`. Can we make these follow the same pattern? E.g., we could have `AbstractMigration` and `AbstractMigrationFunctor`.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -41,102 +41,107 @@ a "duc query" (disjoint union of conjunctive queries).
...
+on those schemas. This functor ``F`` may be partial, in which case
+it contains extra data describing how to use it to execute a migration. The
+migration may be a pullback data migration ([`DeltaMigration`](@ref)), specified by a
+functor ``D → C`` between the schemas, or a conjunctive, gluing,
+duc, or gluc (contravariant) data migration. Alternatively, ``F`` may
+be a schema functor specifying a ``Σ`` migration in the covariant direction.
 """
-const DeltaSchemaMigration{D<:FinCat,C<:FinCat} = FinFunctor{D,C}
+abstract type AbstractDataMigration{F<:FinDomFunctor} end
+func(M::AbstractDataMigration) = M.functor
```
epatters, line 54: I'd strongly prefer to rename `func` to `functor` because:

- `func` could mean function or functor
- `functor` is only two characters longer
- `functor` is already a generic function that Catlab exports, so won't have to add another, inconsistent name to the namespace

You'll need `import ...Syntax: functor` at the top of the file.

epatters (changes requested):
Thanks Kevin! This is a nice design. A few comments below.

epatters (approved):

PR #795, opened 2023-05-31
Remove acset transformation type parameters kept for backward compatibility

### LINKED ISSUE #755: Remove acset transformation type parameters kept for backward compatibility
In #754, we reintroduced two type parameters, `S` and `Comp`, for the abstract type `ACSetTransformation` to restore backwards compatibility. However, these only make sense for struct acset transformations (and the second one really shouldn't be there at all). They should be removed in the next breaking release.

The two other type parameters, `Dom` and `Codom`, can be kept.

epatters:
Closes #755.


epatters:
FYI @mehalter, this is something that was previously added for backwards compatibility with AlgPetri but is now being removed.

PR #796, opened 2023-06-01
Upgrade to ACSets.jl v0.2

epatters:
Taking into account the breaking change in https://github.com/AlgebraicJulia/ACSets.jl/pull/31.


epatters:
Note that benchmarks are actually passing now. It's just the comparison with the old benchmarks that is failing, a typical problem.

PR #806, opened 2023-06-14
Transformations between acsets involving only combinatorial data

kris-brown:
By considering CSetTransformations between ACSets (where naturality is **not** demanded on attributes) we can obtain two sorts of practical constructions: 

1. A kind of "labeled CSet" where we can think of the attributes as functioning as labels
2. A kind of limit construction for ACSets which acts purely on the combinatorial data. The apex can put whatever it wants for the attributes, but we can deliberately *pick* a representative of the isomorphism class to make it correspond to something meaningful (with valid ACSetTransformations), which Kevin has a nice categorical characterization of in an AlgebraicJulia [blog post](https://blog.algebraicjulia.org/post/2023/06/varacsets/).

Implementation-wise, this means we want an abstract type of `ACSetMorphism` which is implemented by `ACSetTransformation` and `CSetTransformation` (therefore, this is changing the meaning of `CSetTransformation`, which before was an `ACSetTransformation` that happened to be on a schema with no attributes). There is probably a lot of functionality which is written for `ACSetTransformation` which would straightforwardly work for `ACSetMorphism`, but the minimal required functionality (e.g. `dom`, `compose`) for getting limits to work is all that is generalized for now.

It may be the case that we want something in between, where a subset of the attributes are marked to be ignored. This would give us a kind of "LACSets" - but we would then lose the ability to do #2 above, which I see as the real value of this.

Edit: I also addressed a tiny bug in the HomSearch code where `monic` and `iso` fast failure constraints were not being checked for the combinatorial Attr data.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -23,13 +23,13 @@ using ACSets.DenseACSets: indices, unique_indices, attr_type, attrtype_type,
 using ...GATs, ...Graphs.BasicGraphs
 using ...Theories: ThCategory, Hom, Ob, Attr, AttrType
 import ...Theories: ob, hom, dom, codom, compose, ⋅, id,
-  meet, ∧, join, ∨, top, ⊤, bottom, ⊥
+  meet, ∧, join, ∨, top, ⊤, bottom, ⊥, ⊕, ⊗
+
 using ..FreeDiagrams, ..Limits, ..Subobjects, ..Sets, ..FinSets, ..FinCats
 using ..FinSets: VarFunction, LooseVarFunction, IdentityFunction, VarSet
 import ..Limits: limit, colimit, universal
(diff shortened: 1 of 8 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -666,10 +720,20 @@ end
...
+
+compose(α::CSetTransformation, β::CSetTransformation) =
+  CSetTransformation(map(compose, components(α), components(β)), dom(α), codom(β))
 
 # Limits and colimits
 #####################
 
+⊕(xs::AbstractVector{T}) where {S, T<:StructACSet{S}} = 
+  foldl(⊕, xs; init=apex(initial(T)))
+⊗(xs::AbstractVector{T}) where {S, T<:StructACSet{S}} =
```
epatters, line 734: These two new methods don't seem to be tested.

epatters (commented):
Very nice! I just left a few small comments.

epatters (approved):
Thanks Kris!

PR #808, opened 2023-06-16
Subobject with attribute variables

kris-brown:
Biheyting logic for acsets with variables

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -1447,6 +1447,13 @@ function hom(A::SubACSetComponentwise{T}) where T <: ACSet
   U = constructor(X)()
   hom_components = map(collect∘hom, components(A))
   copy_parts!(U, X, hom_components)
+  # Reindex attrvars in the domain
+  for at in attrtypes(acset_schema(X))
+    comp = Dict(v=>AttrVar(k) for (k,v) in enumerate(collect(hom(components(A)[at]))))
+    for atr in attrs(acset_schema(X), to=at, just_names=true)
+      U[atr] = [get(comp, x, x) for x in U[atr]]
+    end
+  end
   ACSetTransformation(hom_components, U, X)
 end
 
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -1446,14 +1446,9 @@ function hom(A::SubACSetComponentwise{T}) where T <: ACSet
   X = ob(A)
   U = constructor(X)()
   hom_components = map(collect∘hom, components(A))
-  copy_parts!(U, X, hom_components)
-  # Reindex attrvars in the domain
-  for at in attrtypes(acset_schema(X))
-    comp = Dict(v=>AttrVar(k) for (k,v) in enumerate(collect(hom(components(A)[at]))))
-    for atr in attrs(acset_schema(X), to=at, just_names=true)
-      U[atr] = [get(comp, x, x) for x in U[atr]]
-    end
-  end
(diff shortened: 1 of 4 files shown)


test/categorical_algebra/CSets.jl
```diff
@@ -811,4 +811,12 @@ clim = colimit(Span(f,g));
 @test apex(clim)[:weight] == [true,true,AttrVar(2),true]
 @test collect(legs(clim)[1][:Weight]) == [true,AttrVar(1)]
 
+# Subobjects with variables 
+
+X = @acset SetAttr{Bool} begin X=2;D=1;f=[true, AttrVar(1)] end 
+A = Subobject(X, X=[1])
+B = Subobject(X, X=[2], D=[1])
+@test A ∧ B |> force == ⊥(X) |> force
+@test A ∨ B |> force == ⊤(X) |> force
```
jpfairbanks, line 820: We need a test for a bigger schema like WeightedGraphs, and ReflexiveGraphs.

src/categorical_algebra/FinSets.jl
```diff
@@ -325,6 +325,7 @@ symbolic attributes. (Likewise, AttrVars will have to wrap Any rather than Int)
     return new(FinDomFunction(Vector{Union{AttrVar,T}}(f)), cod)
   end 
 end
+VarFunction(f::UnitRange,cod::Int) = VarFunction(FinFunction(f,cod))
```
epatters, line 328: This function is not covered and since

```julia
UnitRange{Int} <: AbstractVector{Int}
```

I figure it is not actually needed?

src/categorical_algebra/CSets.jl
```diff
@@ -1439,6 +1447,13 @@ function hom(A::SubACSetComponentwise{T}) where T <: ACSet
   U = constructor(X)()
   hom_components = map(collect∘hom, components(A))
   copy_parts!(U, X, hom_components)
+  # Reindex attrvars in the domain
```
epatters, line 1450: Does `copy_parts!` not support attribute variables? It seems like this logic shouldn't need to be here.

src/categorical_algebra/FinSets.jl
```diff
@@ -1464,4 +1465,31 @@ join(A::SubFinSet{Int}, B::SubFinSet{Int}, ::SubOpBoolean) =
 top(X::FinSet{Int}, ::SubOpBoolean) = SubFinSet(trues(length(X)))
 bottom(X::FinSet{Int}, ::SubOpBoolean) = SubFinSet(falses(length(X)))
 
+# SubVarSets
+
+""" Subset of a finite varset.
+"""
+const SubVarSet{T} = Subobject{VarSet{T}}
+
+Subobject(X::VarSet{T}, f::Union{AbstractVector, UnitRange}) where T =
```
epatters, line 1474: Looks like you have two copies of the same method, which is why this one isn't covered. Also, `AbstractVector` will probably suffice instead of `Union{AbstractVector,UnitRange}`.

src/categorical_algebra/CSets.jl
kris-brown: `copy_parts!` does copy attribute variables, but does so literally (it doesn't reindex them).

You're right that it should probably also handle the reindexing logic, but that would be a change to ACSets.jl.

epatters: OK, yes, that does sound like a change that should be made to ACSets. Can you create an issue upstream?

kris-brown: https://github.com/AlgebraicJulia/ACSets.jl/issues/38

epatters (approved):

PR #810, opened 2023-06-21
BUILD: fix BenchmarkCI assumption that `origin/master` exists

mehalter:
Not sure why the default `baseline` is not set to the default branch set in the git repo smh


epatters (approved):
Thanks for sorting out these issues, Micah!

PR #812, opened 2023-06-21
BUILD: actually fix review checklist

mehalter:
Debugged and validated solution on my own fork: https://github.com/mehalter/Catlab.jl/pull/1


epatters (approved):
Thanks for slogging through. Sounds like it has been a pain.

mehalter:
It was a bit painful 😅 I just tested it (#813) and it worked great!

PR #814, opened 2023-06-22
Optional visual trace of backtracking search tree

kris-brown:
This PR makes three improvements to homomorphism search:

1. The main one is that an optional `BacktrackingTree` data structure is added to the `BacktrackingState`. If `debug_homomorphisms(...)` is called rather than `homomorphisms(...)`, then a list of homomorphisms in addition to a   `BacktrackingTree` are returned as a pair.  This can be viewed in graphviz (see below).

2. `find_mrv_elem` picks which C-Set part to branch on, attempting to minimize the branching factor. It does this by finding, for each part, how many valid assignments there are. Previously it just counted how many valid assignments, rather than remembering what these were - this is fixed by just using `filter` rather than `count`. I would expect this to have a positive performance impact, but this hasn't been benchmarked.

3. When a successful assignment is found, there is a bit of logic to extract an `ACSetTransformation` from the `BacktrackingState`. This has now been factored into a function `get_hom`.


kris-brown:
Example of the search tree graphviz output (included the hover text)
Screenshot 2023-06-21 at 6 42 42 PM

kris-brown:
(TODO: Rebase and then tag Evan as a reviewer)

PR #816, opened 2023-06-26
Fix omitted case in `leftmost_arg`

### LINKED ISSUE #805: Inconsistency between composition operations
There seems to be a bug resulting in an inconsistency between composition written in classical and diagrammatic order.

See the reproduction example below:

```
# Tested with Catlab v0.15.0
using Catlab.Graphs, Catlab.Graphics
using Catlab.Theories, Catlab.CategoricalAlgebra, Catlab.Programs

g = path_graph(Graph, 3)

# Works fine!
h₁ = @migrate Graph g begin
  V => @cases (v₁::V; v₂::V)
  E => E
  src => v₁ ∘ src
  tgt => v₂ ∘ tgt
end

# MethodError: Cannot `convert` an object of type Nothing to an object of type Symbol
h₂ = @migrate Graph g begin
  V => @cases (v₁::V; v₂::V)
  E => E
  src => src ⋅ v₁
  tgt => tgt ⋅ v₂
end

# Works fine!
k₁ = @migrate Graph g begin
  V => V
  E => @join begin
    (e₁, e₂)::E
  end
  src => e₁ ⋅ src
  tgt => e₂ ⋅ tgt
end

# MethodError: Cannot `convert` an object of type Nothing to an object of type Symbol
k₂ = @migrate Graph g begin
  V => V
  E => @join begin
    (e₁, e₂)::E
  end
  src => src ∘ e₁
  tgt => tgt ∘ e₂
end
```

KevinArlin:
Resolves #805 .

The `leftmost_arg` function that's used to build `AST.Apply` and `AST.Coapply`s was missing a case for what to do when giving a composition operator that works in the opposite order of the target AST type.

### DIFF
--- a/src/programs/DiagrammaticPrograms.jl
+++ b/src/programs/DiagrammaticPrograms.jl
@@ -1105,7 +1105,7 @@ end
 function parse_mapping_ast(body, dom; kw...)
   parse_mapping_ast((rhs, _) -> parse_ob_ast(rhs), body, dom; kw...)
 end
-
+#ugh wtf is leftmost arg's even deal with \circ
 function parse_apply_ast(expr, X, target)
   y::Symbol, f = @match expr begin
     ::Symbol => (expr, nothing)
@@ -1166,6 +1166,8 @@ function reparse_arrows(expr)
 end
 
 """ Left-most argument plus remainder of left-associated binary operations.
+`ops` denotes the operations that won't need to be reversed for the desired
+parser output (AST.Apply or AST.Coapply).
 """
 function leftmost_arg(expr, ops; all_ops=nothing)
   isnothing(all_ops) && (all_ops = ops)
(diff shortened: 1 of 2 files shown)


src/programs/DiagrammaticPrograms.jl
```diff
@@ -1105,7 +1105,7 @@ end
 function parse_mapping_ast(body, dom; kw...)
   parse_mapping_ast((rhs, _) -> parse_ob_ast(rhs), body, dom; kw...)
 end
-
+#ugh wtf is leftmost arg's even deal with \circ
```
epatters, line 1108: This comment should be removed.

epatters (changes requested):
Thanks Kevin!

src/programs/DiagrammaticPrograms.jl
KevinArlin: Whoops :)

KevinArlin: OK, it's removed.

epatters (approved):

PR #817, opened 2023-06-27
More support for subacsets with variables

kris-brown:
This PR extends https://github.com/AlgebraicJulia/Catlab.jl/pull/808 to support negation / subtraction / implication for VarACSets.

Because monic VarFunctions are representable by FinFunctions, a lot of redundant code was eliminated by just treating the components of the AttrType parts the same as the Ob parts - the only care that's needed is to map over the components with `AttrVar` when constructing the ACSetTransformation components (and undoing this when going in the reverse direction).


epatters (approved):
Thanks Kris!

PR #818, opened 2023-06-28
Category of elements for general C-sets

kris-brown:
There was an unnecessarily strict typing on the the argument to `elements` which required it to be a `StructACSet` - this PR relaxes that constraint.


aaguinal:
Any chance we can bump this up in the merge queue? Thank you!

epatters (approved):
Thanks, looks great!

PR #821, opened 2023-06-30
Enumerating partial overlaps as an iterator

kris-brown:
Previously, the subobjects in order of decreasing size were computed as an iterator, but the partial overlaps between ACSets was computed all at once. However, for things like maximum common subacset, we are not interested in enumerating all possible overlaps, so this PR modifies the partial overlap computation to be an iterator.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -14,7 +14,7 @@ export ACSetTransformation, CSetTransformation, StructACSetTransformation,
 
 using Base.Iterators: flatten
 using Base.Meta: quot
-using DataStructures: BinaryHeap, OrderedDict
+using DataStructures: BinaryHeap, DefaultDict
 using MLStyle: @match
 using Random
 using Reexport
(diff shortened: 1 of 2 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -1734,6 +1734,28 @@ function Base.iterate(Sub::SubobjectIterator, state=SubobjectIteratorState())
...
+  top::ACSet
+  others::Vector{ACSet}
+  function OverlapIterator(Xs::Vector{T}) where T<:ACSet
+    t, o... = Xs[sortperm(total_parts.(Xs))]
+    new(t, o)
+  end
+end
+
+function Base.collect(itr::OverlapIterator)
+  res = []; for sp in itr push!(res, sp) end; res
```
epatters, line 1747: Is this method needed? Given that the iteration interface is implemented, I would expect `Base.collect` to already do this.

src/categorical_algebra/CSets.jl
```diff
@@ -1734,6 +1734,28 @@ function Base.iterate(Sub::SubobjectIterator, state=SubobjectIteratorState())
 end
 
 
+struct OverlapIterator
+  top::ACSet
+  others::Vector{ACSet}
+  function OverlapIterator(Xs::Vector{T}) where T<:ACSet
+    t, o... = Xs[sortperm(total_parts.(Xs))]
```
epatters, line 1741: If I understand correctly,

```julia
Xs[sortperm(total_parts.(Xs))]
```

can be simplified to

```julia
sort(Xs, by=total_parts)
```

src/categorical_algebra/CSets.jl
kris-brown: ah I just had to define Base.IteratorSize for OverlapIterator!

epatters (approved):
Looks good, thanks Kris!

PR #823, opened 2023-07-03
Data migrations with Julia functions on attributes

KevinArlin:
This PR upgrades the `@migration` macros with the ability to plug Julia code into appropriate locations. For instance:

- If ``a`` is an attribute of the target schema, one can write `a=> x->body`, where `body` may depend on `x` and all the names in the source schema. This allows the produced migration to mutate the attributes in a more-or-less arbitrary way. The generality of this mutation is illustrated in the tests by `TuringMachine`.
- Similarly, if ``h`` is a hom in the target schema, we can write ``h``'s image in terms of an anonymous function whose value depends on all the names in the source schema, but one whose values will be combinatorial data rather than attribute values.
- More generally, the above hold when a hom or an attribute is sent to a multi-line diagram hom, for each line of that hom independently.
- If an object of the target schema is sent to a conjunctive diagram, this diagram may include morphisms over expressions of the same form as in the previous bullet, allowing the migration to effectively filter the migrated acset via arbitrary code applied to its attributes.
- When an attribute ``a`` of the target schema has as domain an object sent to a conjunctive diagram containing an object ``y``, we may write `a=> (b(y)|>(x->body))` to disambiguate the input.
- The single-line `->` expressions can be replaced with blocks whose last line is such an expression.

The behavior is basically that the macro builds a `DataMigration` superficially just as it used to, but where some values of the `hom_map` of the underlying functor may be `AttrExpr{:nothing}`, and those and perhaps other homs have their values modified upon actually migrated using the Julia functions stored in the migration's `params`.

The most immediate remaining concern is that the code relies on changes in lots of places, at all three of the key steps of the migration macros (parsing into values from `DiagrammaticPrograms.AST`, then into actual `Diagram`s and `DiagramHom`s, then doing the actual migration), while also allowing for several different syntaxes, and so is not so clean.

The biggest large-scale concern is that we can't yet use multi-input Julia functions, as for instance you'd want to take a product of weighted graphs with the weights of a pair of edges given by the sum of the pair's member's weights. This is an actual semantic step beyond the model we've used so far, as the intermediate migration of something like `weight => (x,y) -> weight(x)+weight(y)` would need to send `Weight` to `Tuple{Weight,Weight}` or to `@cases W:Weight; U:Weight`, and this splitting would need to be collapsed upon the actual migration. None of this shifting of attribute types happens with examples we can do up till now. The `Tuple` solution seems easiest, but it does lead to the question of what happens if you have two different combinatorial types using the same attribute type a different number of times each. So I don't think there's an obvious choice of direction here.


test/programs/DiagrammaticPrograms.jl
```diff
@@ -577,4 +545,323 @@ F_src_vv, F_src_ev, F_src_ve = components(diagram_map(F_src))
...
+          L :: Len
+          (l:e→L) :: (x->len(x)^2)
+          (d:e→L) :: (x->sum((pos(src(x))-pos(tgt(x))).^2))
+      end
+      Pos => Pos
+      Len => Len
+      (src:E→V) => src(e)
+      (tgt:E→V) => tgt(e)
+      (pos:V→Pos) => pos
+      (len:E→Len) => (len(e)|>(x->2x))
```
epatters, line 678: Not high priority but consider supporting syntax `len(e) |> double` where `double` is a side function defined as `double(x) = 2x`.

test/programs/DiagrammaticPrograms.jl
```diff
@@ -577,4 +545,323 @@ F_src_vv, F_src_ev, F_src_ve = components(diagram_map(F_src))
...
+      e₁ => v₁ ∘ tgt
+      e₂ => v₂ ∘ tgt
+    end
+    pos => begin
+      v₁ => pos
+      v₂ => (pos|> (x-> -x)) 
+    end
+    len => (e₁ => len ; e₂ => len)
+  end
+  #=variant,TODO
```
epatters, line 720: TBD: either extend DSL to support this case or remove commented code.

test/programs/DiagrammaticPrograms.jl
```diff
@@ -577,4 +545,323 @@ F_src_vv, F_src_ev, F_src_ve = components(diagram_map(F_src))
...
+      e₁ => v₁ ∘ (e⋅tgt)
+      e₂ => v₂ ∘ (e⋅tgt)
+    end
+    pos => begin
+      v₁ => pos
+      v₂ => (pos|> (x-> -x)) 
+    end
+    len => (e₁ => e⋅len ; e₂ => e⋅len)
+  end
+#=not yet present
```
epatters, line 775: Remove from this PR and possibly revisit later.

test/programs/DiagrammaticPrograms.jl
```diff
@@ -577,4 +545,323 @@ F_src_vv, F_src_ev, F_src_ve = components(diagram_map(F_src))
...
+      src => src
+      tgt => tgt
+      pos => begin 
+              θ = π/5
+              M = [[cos(θ),sin(θ),0] [-sin(θ),cos(θ),0] [0,0,1]]
+              x -> M*pos(x)
+              end
+      len => len
+  end
+  A = migrate(MechLink,G,M)
```
epatters, line 600: `migrate` should be tested in `categorical_algebra/DataMigrations.jl` rather than here.

test/programs/DiagrammaticPrograms.jl
```diff
@@ -577,4 +545,323 @@ F_src_vv, F_src_ev, F_src_ve = components(diagram_map(F_src))
...
+              m
+    x -> x == p(1) ? μ(marks(x),state(1)) : marks(x)
+    end
+  end
+  run_TM(T,n) = n == 0 ? T : run_TM(migrate(TuringMachine,T,M),n-1)
+
+  @test begin subpart(run_TM(T,10),1:8,:marks) == Union{AttrVar,enums.BinaryMarks}[map(x->enums.O,1:6);enums.X;enums.Blank] end
+end
+
+@present SchAng(FreeSchema) begin
```
epatters, line 851: Make into example understandable without context from PTG.

test/categorical_algebra/Diagrams.jl
```diff
@@ -71,6 +71,18 @@ d = dom(f)
 @test codom(op(f)) == op(dom(f))
 @test op(g)⋅op(f) == op(f⋅g)
 
+# param_compose, should probably be in FinCats
```
epatters, line 74: Remove since not part of public interface and implicitly tested by data migration pipeline.

test/categorical_algebra/Diagrams.jl
```diff
@@ -71,6 +71,18 @@ d = dom(f)
...
+####################
+C = FinCat(parallel_arrows(Graph,2))
+D = FinCat(FinSet(2))
+F = FinDomFunctor([FinSet(2),TypeSet(Float64)],[FinDomFunction([1.0,2.3]),FinDomFunction([-π,π])],C)
+G = FinFunctor([1,1],D,C)
+H = FinFunctor([2,2],D,C)
+α = CategoricalAlgebra.FinCats.FinTransformationMap([1,FreeSchema.Attr{:nothing}([],[])],G,H)
```
epatters, line 81: Should create an issue to track how we can refactor this design in the future. One possibility is allowing functors and transformations to be partial by default (like we do for functions in acsets).

src/theories/Category.jl
```diff
@@ -44,6 +44,7 @@ symbol ``⋅`` (`\\cdot`) for composition in diagrammatic order.
 end
 
 # Convenience constructors
+compose(f) = f
```
epatters, line 47: Is this needed? Preferable to use `compose(::AbstractVector)` for programmatic use.

src/categorical_algebra/FinSets.jl
```diff
@@ -258,7 +258,11 @@ the form ``{1,...,n}``.
   func::V
   codom::Codom
 end
-
+function Base.convert(::Type{SetFunctionCallable},S::FinDomFunctionVector)
+  FinDomFunction((i->S.func[i]),
+                        FinSet(length(S.func)),
+                        S.codom)
+end
```
epatters, line 265: Shouldn't have this conversion. Instead, construct vectors with explicit types to avoid overly strong typing.

src/categorical_algebra/FinSets.jl
```diff
@@ -278,7 +282,10 @@ function Base.show(io::IO, f::FinDomFunctionVector)
   print(io, ")")
 end
 
-force(f::FinDomFunction{Int}) = FinDomFunctionVector(map(f, dom(f)), codom(f))
+function force(f::FinDomFunction{Int})
+  d = map(f, dom(f))
+  FinDomFunctionVector(Vector{eltype(codom(f))}(d), codom(f))
```
epatters, line 287: `FinDomFunctionVector(eltype(codom(f))[ f(x) for x in dom(f) ], codom(f))`.

Actually, instead of doing this in `force`, write `eltype(codom(f))[ f(x) for x in dom(f) ]` when you need the vector to have certain type.

src/categorical_algebra/FinCats.jl
```diff
@@ -67,6 +67,7 @@ Because object generators usually coincide with objects, the default method for
 function ob_generator end
 
 ob(C::FinCat, x) = ob_generator(C, x)
+hom(C::FinCat,x::Nothing) = x
```
epatters, line 70: Remove this line and two more below.

src/categorical_algebra/FinCats.jl
```diff
@@ -371,8 +374,11 @@ FinDomFunctor(ob_map, ::Nothing, dom::FinCat, codom::Cat) =
 function hom_map(F::FinDomFunctor, path::Path)
   C, D = dom(F), codom(F)
   path = decompose(C, path)
-  mapreduce(e -> hom_map(F, e), (gs...) -> compose(D, gs...),
-            edges(path), init=id(D, ob_map(F, src(path))))
+  #this is horrific, to be discussed
```
epatters, line 377: Good evidence that we should incorporate "null" morphisms in some more explicit way.

src/categorical_algebra/FinCats.jl
```diff
@@ -642,10 +652,10 @@ function Categories.do_composeH(F::FinDomFunctorMap, β::Transformation)
                        compose(F, G, strict=false), compose(F, H, strict=false))
 end
 
-function Categories.do_composeH(α::FinTransformationMap, H::Functor)
+function Categories.do_composeH(α::FinTransformationMap, H::Functor; params=[])
```
epatters, line 655: Looks like leftover from previous design. No `params` used below.

epatters (changes requested):
Thanks Kevin, this will be an awesome feature. First round of comments below.

src/theories/Category.jl
```diff
@@ -197,3 +199,34 @@ abstract type TightExpr{T} <: GATExpr{T} end
...
+zero maps and zero object in the codomain.
+"""
+@theory ThPtCategory{Ob,Hom,ZeroHom} <: ThCategory{Ob,Hom} begin
+  Z()::Ob
+  ZeroHom(dom::Ob,codom::Ob)::TYPE
+  z(A::Ob,B::Ob)::ZeroHom(A,B)
+  hom(x::ZeroHom(A,B))::Hom(A,B)⊣(A::Ob,B::Ob)
+  compose(x::ZeroHom(A,B),f::(B→C))::ZeroHom(A,C)⊣(A::Ob,B::Ob,C::Ob)
+  compose(g::(A→B),x::ZeroHom(B,C))::ZeroHom(A,C)⊣(A::Ob,B::Ob,C::Ob)
+  z₁==z₂⊣(A::Ob,B::Ob,z₁::ZeroHom(A,B),z₂::ZeroHom(A,B))
```
epatters, line 223: I think we can get rid of `ZeroHom`, have `zero(A::Ob, B::Ob)::Hom(A,B)`, and equation

```julia
compose(zero(A,B), f) == zero(A,C) ⊣ (A::Ob, B::Ob, C::Ob, f::Hom(B,C))
```

plus other equation. Will give `HomExpr{:zero}` type in syntax.

src/theories/Category.jl
```diff
@@ -197,3 +199,34 @@ abstract type TightExpr{T} <: GATExpr{T} end
...
+end
+
+abstract type ZeroHomExpr{T} <: CategoryExpr{T} end
+@syntax FreePtCategory{ObExpr,HomExpr,ZeroHomExpr} ThPtCategory begin
+  compose(f::Hom,g::Hom) = associate_unit(new(f,g; strict=true), id)
```
epatters, line 228:
```julia
# inside @syntax
compose(f::Hom,g::Hom) = associate_unit(normalize_zero(new(f,g; strict=true)), id)

# outside @syntax
normalize_zero(f::HomExpr{:zero}, g::HomExpr) = zero(dom(f), codom(g))
normalize_zero(f::HomExpr, f::HomExpr{:zero}) = ...
normalize_zero(f::HomExpr{:zero}, g::HomExpr{:zero}) = ...
```

src/gats/SyntaxSystems.jl
```diff
@@ -348,7 +347,11 @@ function generator_like(expr::GATExpr, value)::GATExpr
   invoke_term(syntax_module(expr), gat_typeof(expr),
               value, gat_type_args(expr)...)
 end
-
+"""As with generator_like, but change the syntax instead of the name."""
```
epatters, line 350: Can write `[generator_like](@ref)` in docstring to get hyperlink.

src/programs/DiagrammaticPrograms.jl
```diff
@@ -30,15 +31,34 @@ abstract type ObExpr end
...
+end
+
 @data DiagramExpr begin
-  ObOver(name::Symbol, over::Union{ObExpr,Nothing})
-  HomOver(name::Union{Symbol,Nothing}, src::Symbol, tgt::Symbol, over::HomExpr)
-  AssignLiteral(name::Symbol, value)
+  ObOver(name::Symbol, over::Union{ObExpr,Nothing}) #probably shouldn't be nothing
+  HomOver(name::Symbol, src::Symbol, tgt::Symbol, over::HomExpr)
+  AttrOver(name::Symbol, src::Symbol,tgt::Symbol,aux_func_def::Expr,mod::Module)#make JuliaCode
+  TwistedHomOver(lhs::HomOver,rhs::AttrOver)
```
epatters, line 56: Maybe something like `HomWithCodeCover`?

src/programs/DiagrammaticPrograms.jl
```diff
@@ -369,12 +389,36 @@ end
...
-  F = FinDomFunctor(d.ob_map, d.hom_map, d.shape, codom)
+  #if simple, we'll throw away the params and make
+  #sure the shape is based on non-pointed stuff
+  simple = isempty(d.params)
+  newShape = changeShape(simple,d.shape)
+  newCodom = changeShape(simple,codom)
+  F = FinDomFunctor(d.ob_map, d.hom_map, newShape, newCodom)
+  simple ? SimpleDiagram{T}(F) : QueryDiagram{T}(F, d.params)
+end
+function changeShape(simple::Bool,oldShape::FinCatPresentation)
```
epatters, line 406: Snake case not camel case for functions.

test/categorical_algebra/Diagrams.jl
KevinArlin: Welp

KevinArlin: Probably should've left this for later, but it's resolved!

src/categorical_algebra/Diagrams.jl
```diff
@@ -105,48 +107,64 @@ type `DiagramHom{id}` and `DiagramHom{op}` induce morphisms between colimits and
...
+    QueryDiagramHom{T,C,F,Φ,D,Params}(shape_map, diagram_map, precomposed_diagram,params)
+"""Convert the diagram category in which a diagram hom is being viewed."""
 DiagramHom{T}(f::DiagramHom) where T =
-  DiagramHom{T}(f.shape_map, f.diagram_map, f.precomposed_diagram)
+  DiagramHom{T}(f.shape_map, f.diagram_map, f.precomposed_diagram,params=get_params(f))
 
-DiagramHom{T}(ob_maps, hom_map, D::Diagram{T}, D′::Diagram{T}) where T =
-  DiagramHom{T}(ob_maps, hom_map, diagram(D), diagram(D′))
+DiagramHom{T}(ob_maps, hom_map, D::Diagram{T}, D′::Diagram{T};params=Dict()) where T =
+  DiagramHom{T}(ob_maps, hom_map, diagram(D), diagram(D′);params=params)
```
epatters, line 139: For this kind of keyword delegation, the recommended approach is

```julia
DiagramHom{T}(ob_maps, hom_map, D::Diagram{T}, D′::Diagram{T}; kw...) where T =
  DiagramHom{T}(ob_maps, hom_map, diagram(D), diagram(D′); kw...)
```

which saves on repetition and makes it easier to refactor/add keyword args later.

epatters (changes requested):
Thanks Kevin! Here's a batch of small questions/comments. I'll pick up with more reviewing later but I wanted to get some stuff down now.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -134,9 +137,35 @@ end
 
 # Delta migration
 #----------------
-#migrate(Y::ACSet,M::DeltaSchemaMigration) = migrate(FinDomFunctor(Y),M)
-#migrate(Y::FinDomFunctor,M::DeltaSchemaMigration) = compose(M,Y)
-#TODO: write compose
+function migrate(X::FinDomFunctor,M::DeltaSchemaMigration)
```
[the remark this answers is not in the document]
KevinArlin: This migrates a FinDomFunctor, rather than an acset. If you call `@migrate` with no explicit target schema then the other methods can't be used.

src/categorical_algebra/DataMigrations.jl
```diff
@@ -161,28 +190,55 @@ function migrate!(X::ACSet, Y::ACSet, FOb, FHom)
   migrate!(X, Y, M)
 end
 
+"""
+Get the names of the morphisms in a schema given either
+by a NamedGraph or by a presentation.
+"""
+get_homnames(S::FinCatGraph) = collect(subpart(graph(S),:,:ename))
```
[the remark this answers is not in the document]
epatters: You're planning to delete the stuff that you commented, right?

src/categorical_algebra/Diagrams.jl
```diff
@@ -105,48 +106,64 @@ type `DiagramHom{id}` and `DiagramHom{op}` induce morphisms between colimits and
...
+  params::Params
+end
+
 DiagramHom{T}(shape_map::F, diagram_map::Φ, precomposed_diagram::D) where
     {T,C,F<:FinFunctor,Φ<:FinTransformation,D<:Functor{<:FinCat,C}} =
-  DiagramHom{T,C,F,Φ,D}(shape_map, diagram_map, precomposed_diagram)
-
+  SimpleDiagramHom{T,C,F,Φ,D}(shape_map, diagram_map, precomposed_diagram)
+#makes sure you'll never accidentally build a querydiagramhom by building a diagramhom 
+#with no params.
```
epatters, line 127: This feels like a code smell to me. Usually you want the type of the output to depend only on the types of the inputs, even in degenerate cases where things are empty.

Is this being done for backwards compatibility?

src/categorical_algebra/Diagrams.jl
```diff
@@ -244,11 +266,68 @@ op(f::DiagramHom{op}) = DiagramHom{id}(op(shape_map(f)), op(diagram_map(f)),
...
                 compose(f.precomposed_diagram, F; kw...))
 end
 
+"""Whisker a partially natural transformation with a functor ``H``,
+given any needed parameters specifying the functions in ``H``'s codomain
+which the whiskered result should map to. Currently assumes
+the result will be a totally defined transformation.
+"""
+
+Base.haskey(v::AbstractVector,i) = 1 <= i <= length(v)
```
epatters, line 281: This is a [type piracy](https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy). We absolutely cannot define this method for abstract vectors.

Also, it's not covered, so presumably is not even being used?

src/categorical_algebra/Diagrams.jl
```diff
@@ -244,11 +266,68 @@ op(f::DiagramHom{op}) = DiagramHom{id}(op(shape_map(f)), op(diagram_map(f)),
...
+  whiskered = isempty(params) ? composeH(diagram_map(f), F; kw...) : param_compose(diagram_map(f),F;params=params)
+  DiagramHom{T}(shape_map(f), whiskered,
                 compose(f.precomposed_diagram, F; kw...))
 end
 
+"""Whisker a partially natural transformation with a functor ``H``,
+given any needed parameters specifying the functions in ``H``'s codomain
+which the whiskered result should map to. Currently assumes
+the result will be a totally defined transformation.
+"""
```
epatters, line 279: Doc string is attached to nothing.

src/categorical_algebra/Diagrams.jl
```diff
@@ -244,11 +266,68 @@ op(f::DiagramHom{op}) = DiagramHom{id}(op(shape_map(f)), op(diagram_map(f)),
...
+
+Base.haskey(v::AbstractVector,i) = 1 <= i <= length(v)
+function param_compose(α::FinTransformation, H::Functor; params=[])
```
epatters, line 282: Would be helpful to have a docstring for this function. I can't immediately tell what it's for besides it obviously having something to do with parameters.

src/categorical_algebra/FinCats.jl
```diff
@@ -104,17 +103,81 @@ is_discrete(C::FinCat) = isempty(hom_generators(C))
 """
 is_free(C::FinCat) = isempty(equations(C))
 
+"""Graph underlying a finitely presented category whose 
+object and
+hom generators are indexable, other
+than one explicitly generated by a graph.
+"""
+function graph(C::FinCat;maps::Bool=false,inv::Bool=false,named::Bool=false)
```
epatters, line 111: Rather than having all these keyword arguments, can we just return a named graph whose names are the object/hom *generators*? Then lookup and indexing will give you both the forward and backward mappings.

This does suggest that maybe thing being returned should not be called "named graph".

src/categorical_algebra/FinCats.jl
```diff
@@ -104,17 +103,81 @@ is_discrete(C::FinCat) = isempty(hom_generators(C))
...
+  s(f) = only(indexin([dom(f)],obgens))
+  t(f) = only(indexin([codom(f)],obgens))
+  add_edges!(g,Int[s(f) for f in homgens],Int[t(f) for f in homgens];ename=nameof.(homgens)) 
+  named || begin g′ = Graph(); copy_parts!(g′,g,(:V,:E)); g = g′ end
+  maps || return g
+  inv ? (g, Dict(obgens[i] => i for i in 1:length(obgens)),
+  Dict(homgens[i] => i for i in 1:length(homgens))) :
+  (g, Dict(pairs(obgens)), Dict(pairs(homgens)))
+end
+#probably move into Graphs
```
epatters, line 125: Agreed that this does not belong here.

src/categorical_algebra/FinCats.jl
```diff
@@ -104,17 +103,81 @@ is_discrete(C::FinCat) = isempty(hom_generators(C))
...
 hom_generators(C::OppositeFinCat) = hom_generators(C.cat)
 
 ob_generator(C::OppositeCat, x) = ob_generator(C.cat, x)
 hom_generator(C::OppositeCat, f) = hom_generator(C.cat, f)
 
+"""
+Reverses the domain and codomain of certain types in a presentation,
+for use in an opposite category.
+"""
+function Base.reverse(p::Presentation,types::Vector{Symbol})
```
epatters, line 168: Since this function is really about categories and uses the functions `dom` and `codom`, hence is not something that makes sense on a generic presentation, would suggest that this function not overload `Base.reverse` and be called something like `reverse_dom_codom`.

src/categorical_algebra/FinCats.jl
```diff
@@ -303,18 +367,39 @@ function FinCatPresentation(pres::Presentation{ThSchema})
...
+  Ob = Union{S.Ob, S.AttrType}
+  Hom = Union{S.Hom, S.Attr, S.AttrType}
+  FinCatPresentation{ThPtSchema,Ob,Hom}(pres)
+end
+"""
+The opposite of a FinCatPresentation whose syntax is not just FreeCategory
+is effectively interpreted as if it were, since the opposite may not 
+strictly speaking exist in another syntax.
+"""
+function hom_generators(C::OppositeCat{Ob,Hom,FinCatSize,Ct}) where {Ob,Hom,Ct<:FinCatPresentation}
```
epatters, line 381: I'm confused by why this is needed. If you use the categories API in the intended way, always interpreting Julia objects that represent objects and morphisms *relative to a category* (`Cat` instance), then taking the opposite category should never require munging the data of the original category.

src/categorical_algebra/FinCats.jl
```diff
@@ -303,18 +367,39 @@ function FinCatPresentation(pres::Presentation{ThSchema})
...
+  S = presentation(C).syntax
+  map(x->S.Hom(nameof(x),S.Ob(S.Ob,nameof(codom(x))),S.Ob(S.Ob,nameof(dom(x)))),hom_generators(C.cat))
+end
+"""
+Computes the graph generating a finitely
+presented category. Ignores any attribute
+side and any equations. Optionally returns
+the mappings from generators to their indices
+in the resulting graph.
+"""
```
epatters, line 391: Docstring in the wrong place.

src/categorical_algebra/FinCats.jl
```diff
@@ -716,11 +836,18 @@ end
 
 # Something like this should be built into Julia, but unfortunately isn't.
 
+"""
+Map two given functions across the respective keys and values of a dictionary.
+"""
```
epatters, line 841: Thanks for writing docstrings for these helper functions! Hopefully we can find a way to get rid of them one day.

src/categorical_algebra/Diagrams.jl
KevinArlin: No, just did it wrong, think I've got it right now.

src/categorical_algebra/Diagrams.jl
KevinArlin: It was mere lines away!

src/categorical_algebra/FinCats.jl
KevinArlin: I think the problem is that hom generators in a FinCatPresentation know their domain and codomain. Previously `hom_generators(op(C))` would return `hom_generators(C)`, which makes `dom` and `codom` be wrong.

src/categorical_algebra/Diagrams.jl
KevinArlin: Eek, my bad. Params are always a dict now so no problem.

src/categorical_algebra/FinCats.jl
KevinArlin: Hmmmmmmm, I'm pretty sure I got this working without returning maps at all but leaving the names as symbols.

src/categorical_algebra/FinCats.jl
KevinArlin: Alright, I moved it into BasicGraphs, non-exported, and imported it, DiagramGraph, and the four vertex/edge_name(d) methods as needed.

src/theories/Category.jl
```diff
@@ -197,3 +198,22 @@ abstract type TightExpr{T} <: GATExpr{T} end
   compose(f::Hom, g::Hom) = associate_unit(new(f,g; strict=true), id)
   transitive(t::Tight, u::Tight) = associate_unit(new(t,u; strict=true), reflexive)
 end
+
+"""
+Theory of a pointed category.
+We axiomatize a category equipped with zero morphisms.
```
epatters, line 204: Since this is going into `Catlab.Theories` I am going to bikeshed the terminology. The nLab suggests that a [pointed category](https://ncatlab.org/nlab/show/pointed+category) is a category with a zero object, whereas this is a category with [zero morphisms](https://ncatlab.org/nlab/show/zero+morphism), which is a category enriched in pointed sets.

We need a clear and not-to-long name for a "category enriched in pointed sets." `PointedSetCategory`?

epatters (commented):
OK, with this round of comments I am through the code itself! Now I just need to go through the tests.

src/theories/Category.jl
KevinArlin: Oh, yeah, I was initially providing the zero object as well as the zero morphisms but then didn't need them. I would prefer `Set_*-Category` in principle, but `PointedSet-Category` is probably clearer to more people. On the other hand perhaps I should just put the zero objects back in so that we can have object-partial functors too for later?

test/programs/DiagrammaticPrograms.jl
KevinArlin: Leaving this open for now.

test/programs/DiagrammaticPrograms.jl
```diff
@@ -140,26 +139,24 @@ d = @diagram C begin
...
+@test F_parsed.ob_map == F.ob_map
 
 d = @diagram SchGraph begin
   v => V
   (e1, e2) => E
   t: e1 → v => tgt
   s: e2 → v => src
 end
-@test diagram(d) == F
+@test diagram(d).ob_map == F.ob_map
```
epatters, line 159: I feel like we shouldn't be pulling out a field to do this test.

test/programs/DiagrammaticPrograms.jl
```diff
@@ -334,8 +315,10 @@ M′ = @migration SchGraph SchSet begin
   V => @unit
   E => X
 end
-# something weird about == for datamigrations
-@test functor(M′) == functor(M) && M′.params == M.params
+@test M′ isa DataMigrations.ConjSchemaMigration
+@test isempty(shape(ob_map(functor(M′), :V)))
+# XX:This one has vector maps in the empty shape_map which isn't right
```
epatters, line 320: Is this comment up to date? Is it pointing to a bug?

test/programs/DiagrammaticPrograms.jl
```diff
@@ -347,8 +330,9 @@ end
 F = functor(M)
 F_V = diagram(ob_map(F, :V))
 @test collect_ob(F_V) == fill(SchGraph[:V], 2)
+s = hom_generators(dom(F))[1]
 F_src = hom_map(F, :src)
-@test collect_ob(F_src) == [(1, SchGraph[:src]), (2, SchGraph[:src])]
+@test F_src.diagram_map.components == Dict(:v₂=>s,:v₁=>s)
```
epatters, line 335: `F_src.diagram_map.components` -> `components(diagram_map(F_src))`

test/runtests.jl
```diff
@@ -1,8 +1,8 @@
 using Test
 
-@testset "GATs" begin
-  include("gats/GATs.jl")
-end
+#@testset "GATs" begin
+# include("gats/GATs.jl")
+#end
```
epatters, line 5: Uh oh!

epatters (commented):
Final round of comments on the tests.

src/gats/Presentations.jl
```diff
@@ -44,6 +45,43 @@ function Base.:(==)(pres1::Presentation, pres2::Presentation)
     pres1.equations == pres2.equations
 end
 
+"""Since Presentations have a module field and modules cannot be deep-copied,
+we need a specialized method.
+"""
+function Base.deepcopy_internal(x::Presentation,stackdict::IdDict)
```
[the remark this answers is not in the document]
KevinArlin: Yeah, I just killed the `deepcopy` in `colimit_representables` and it doesn't seem to matter.

test/programs/DiagrammaticPrograms.jl
KevinArlin: Out of date!

test/runtests.jl
KevinArlin: lol sorry, 1.10 breaks a few tests in GATs at the moment

KevinArlin:
Ready for re-review (!!!!!) 

I should note that there aren't any tests for aggregation and such unintended applications yet.

epatters (approved):

PR #829, opened 2023-07-12
Better support for attributes in sigma migrations 

### LINKED ISSUE #825: chase with domain attributes
Currently, the chase tries to create a component for each object and attribute type in the collage of a functor between two schemas. Because we cannot have maps between attribute types, this will currently error. 

There is a workaround that doesn't require making this fundamental change to Catlab schemas, however. If we simply omit the maps between attribute types, the result of the sigma migration should still work fine, except there will be distinct attrvars for every attribute value. A simple `for` loop should be sufficient to then set values for all these attributes according to the naturality squares which were omitted during the chase.

KevinArlin:
Closes #825, allowing sigma-migrations along functors out of schemas with attrtypes, as long as acset morphisms are assumed to be tight.


src/categorical_algebra/DataMigrations.jl
```diff
@@ -335,9 +340,32 @@ function (M::SigmaMigrationFunctor)(d::ACSet; n=100)
...
+  for (k,kdom,kcod) in attrs(S)
+    f = hom_map(F,k)
+    #split f into its hom part and its attr part
+    f1,f2 = split_r(f)
+    f1 = hom_map(i2,f1)#,hom_map(i2,f2)
+    for i in parts(d,kdom)
+      oldval = subpart(d,i,k)
+      #Find where i goes under alpha, and then where that goes
+      #under the hom part of f, by walking the spans in rel_res.
+      j = rel_res[only(incident(rel_res,i,Symbol("src_α_$kdom"))),
```
kris-brown, line 354: use `src_a, tgt_a = add_srctgt(Symbol("α_$kdom"))` to abstract away the string hacking (and have one place to make a change if we want to change the convention)

(also below with `"src_$f1"`)

src/categorical_algebra/DataMigrations.jl
```diff
@@ -335,9 +340,32 @@ function (M::SigmaMigrationFunctor)(d::ACSet; n=100)
...
+                  Symbol("tgt_α_$kdom")] 
+      f1j = f1 isa GATExpr{:id} ? j :
+       rel_res[only(incident(rel_res,j,Symbol("src_$f1"))),
+              Symbol("tgt_$f1")]
+      res[f1j,first(f2)] = oldval
```
kris-brown, line 359: `nameof(f2)`

src/categorical_algebra/DataMigrations.jl
```diff
@@ -335,9 +340,32 @@ function (M::SigmaMigrationFunctor)(d::ACSet; n=100)
       res[domval,h] = codomval
     end
   end
+  #Go back and make sure attributes that ought to have 
+  #specific values because of d do have those values.
+  for (k,kdom,kcod) in attrs(S)
+    f = hom_map(F,k)
+    #split f into its hom part and its attr part
+    f1,f2 = split_r(f)
+    f1 = hom_map(i2,f1)#,hom_map(i2,f2)
```
kris-brown, line 349: Explain in the comment why f1 is hit with the iota but not f2

kris-brown (commented):
Looks great, just some small cosmetic fixes!

epatters:
@KevinArlin, did you forget to push a commit? I see that Kris' comments are resolved but not any changes.

KevinArlin:
> @KevinArlin, did you forget to push a commit? I see that Kris' comments are resolved but not any changes.

Thanks, yeah, looks like it.

src/categorical_algebra/Chase.jl
```diff
@@ -352,19 +352,20 @@ end
 A collage of a functor is a schema encoding the data of the functor
 It has the mapping data in addition to injections from the (co)domain.
 """
-function collage(F::FinFunctor)
+function collage(F::FinFunctor,obs=ob_generators(dom(F)))
```
epatters, line 355: IMO, it's usually makes for clearer code to use keyword arguments with a clear keyword rather than optional positional arguments. So this would be

```julia
function collage(F::FinFunctor; objects=ob_generators(dom(F))
```

src/categorical_algebra/DataMigrations.jl
```diff
@@ -335,9 +340,36 @@ function (M::SigmaMigrationFunctor)(d::ACSet; n=100)
...
+    end
+  end
   return res
 end 
-
+"""
+Split an n-fold composite (n may be 1) 
+Hom or Attr into its left n-1 and rightmost 1 components
+"""
+split_r(f) = f isa GATExpr{:compose} ? (compose(f.args[1:end-1]),f.args[end]) : (id(dom(f)),f)
```
epatters, line 372: Better to use the accessor functions for GAT exprs, so this would be:

```julia
head(f) == :compose ? (compose(args(f)[1:end-1]), last(f)) : (id(dom(f)), f)
```

epatters (changes requested):
Thanks Kevin! I left two more small comments.

Also, it's not essential, but it would be nice to have a unit test that's more independently meaningful than this PTG-inspired one.

epatters (approved):
Thanks Kevin!

PR #830, opened 2023-07-15
VM-based homomorphism search

olynch:


kris-brown:
Future work:

- Handle attributes
- Sample random codomains and optimize the compiled program  
- Convert program into an actual julia function rather than interpreting the VM
- Load instruction has `f` as a symbol, so dynamic dispatch used. This could be made static.

test/categorical_algebra/HomSearch.jl
```diff
@@ -148,6 +148,53 @@ end
 @test_throws ErrorException @acset_transformation g h begin V = [4,3,2,1]; E = [1,2,3,4] end
 
 
+# VM search
+#----------
+using Catlab.CategoricalAlgebra.HomSearch: compile_search
```
epatters, line 153: If we need to do this in both the benchmarks and the tests, perhaps we should be exporting a function named `compile_hom_search` or something similar. (I think `compile_search` is a little vague.)

epatters (changes requested):

epatters (approved):

PR #831, opened 2023-07-17
Support for mark-as-deleted in acsets

kris-brown:
This branch incorporates [mark as deleted](https://github.com/AlgebraicJulia/ACSets.jl/pull/32) into Catlab. As was the case with VarACSets, the priority is getting support for hom search. Colimits would be the next natural thing to do. In the course of working on this, the problem of Catlab having it's own version of everything in ACSets (FinFunction machinery vs column + index machinery) was very painful. 

There could be no difference in the data structure for a morphism between MaD ACSets if we just treat the data in position `i` of a FinFunction (when `i` has been deleted in the domain) as to-be-ignored.  

Constructing the ACSetTransformation data structure (e.g. as the result of a hom search) posed a challenge, as the data gets validated in many ways which do not work when parts have been deleted. 

There are many challenges in making colimits work. However, since AlgebraicRewriting has support for in-place colimits (for MaD ACSets), there is no need to integrate with the normal colimit infrastructure just yet.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -6,7 +6,7 @@ export ACSetMorphism,
   StructTightACSetTransformation, TightACSetTransformation,
   LooseACSetTransformation, SubACSet, SubCSet,
   components, type_components, force,
-  naturality_failures, show_naturality_failures, is_natural,
+  naturality_failures, show_naturality_failures, is_natural, in_bounds,
   abstract_attributes
 
 using Base.Iterators: flatten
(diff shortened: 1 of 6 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -643,18 +647,18 @@ function naturality_failures(X, Y, comps, type_comps; only_combinatorial=false)
   Fun = Union{SetFunction,VarFunction,LooseVarFunction}
   comps = Dict(a => isa(comps[a],Fun) ? comps[a] : FinDomFunction(comps[a])  
                for a in keys(comps))
-              
   type_comps = Dict(a => isa(type_comps[a], Fun) ? type_comps[a] : 
                         SetFunction(type_comps[a],TypeSet(X,a),TypeSet(Y,a)) 
                     for a in keys(type_comps))
-  α = merge(comps, only_combinatorial ? Dict() : type_comps)
```
epatters, line 650: Is `only_combinatorial` still supported as a keyword argument? I don't see it being used in the new code.

epatters (commented):
Looks good! Just one question below.

src/categorical_algebra/CSets.jl
kris-brown: The integration tests passed (other than AlgRewriting, which has a compatible PR waiting to be released once this one is merged).

`only_combinatorial` was relevant when we considered the subtle difference of `ACSetTransformations` and `CSetTransformations` which, this past year, came to have the meaning of two different ways of composing the same data (the latter *ignored* attribute data). However, this was a distinction we tried to make at the type level because GATlab didn't exist! Now we can simply have two different models of ThCategory for ACSetTransformations, one which ignores attributes and one which doesn't. So I think we can go ahead and remove this (currently unused) functionality.

epatters (approved):
Thanks Kris!

PR #835, opened 2023-07-27
Fix `abstract_attributes` function

kris-brown:
`abstract_attributes` is currently an unexported function. It takes an acset `X` and makes a map `A -> X` where `A` is 'all variables'. This is useful in AlgebraicRewriting. It might be more generally useful such that it should be an exported function. 

Regardless, the implementation in Catlab was broken and this PR fixes it and adds a test.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -1619,20 +1619,21 @@ subparts.
 """
 function abstract_attributes(X::ACSet)
   S = acset_schema(X)
-  A = copy(X); 
-  comps = Dict{Any,Any}(map(attrtypes(S)) do at
-    rem_parts!(A, at, parts(A,at))
-    comp = Union{AttrVar,attrtype_type(X,at)}[]
+  A = deepcopy(X);
+  comps = Dict{Any, Any}(map(attrtypes(S)) do at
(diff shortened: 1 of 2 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -1619,20 +1619,21 @@ subparts.
 """
 function abstract_attributes(X::ACSet)
   S = acset_schema(X)
-  A = copy(X); 
-  comps = Dict{Any,Any}(map(attrtypes(S)) do at
-    rem_parts!(A, at, parts(A,at))
-    comp = Union{AttrVar,attrtype_type(X,at)}[]
+  A = deepcopy(X);
```
epatters, line 1622: Thanks for the fix! I doubt this needs to be a deep copy since you're not mutating the attribute *values*, just the acset parts which are copied by a shallow copy.

kris-brown: ACSets.jl actually already does a `deepcopy` on the parts (and a shallow copy of the subparts), so switching this to `copy` does have the correct behavior.

epatters: Yay!

epatters (approved):

Issue #838, opened 2023-08-17
ACSetTransformation mapping attributes to nothing in StockFlow.jl no longer works after upgrading to Catlab 15


neonWhiteout:
@kris-brown 

Previously, we mapped attributes in an ACSetTransformation to nothing to create a natural transformation for a pullback, of the form `Position = x -> nothing`.  A particular example can be see in cell 11 [here](https://github.com/AlgebraicJulia/StockFlow.jl/blob/master/examples/full_fledged_schema_examples_new/stratification/sir_linear_stratification.ipynb).

This no longer works for version 15 of Catlab.

Will check [typed_product](https://github.com/AlgebraicJulia/AlgebraicPetri.jl/blob/main/src/TypedPetri.jl) in algebraic petri to see if it can be used for this.

neonWhiteout:
```julia
typed_WeightModel=ACSetTransformation(WeightModel, l_type,
  S = [s,s,s],
  SV = [N],
  LS = [lsn,lsn,lsn],   
  F = [f_birth, f_death, f_fstorder, f_death, f_fstorder, f_death, f_aging, f_aging, f_aging],    
  I = [i_birth, i_aging, i_fstorder, i_aging, i_fstorder, i_aging], 
  O = [o_death, o_fstorder, o_aging, o_death, o_fstorder, o_aging, o_death, o_aging],
  V = [v_birth, v_death, v_fstorder, v_death, v_fstorder, v_death, v_aging, v_aging, v_aging],
  LV = [lv_death1, lv_fstorder1, lv_death1, lv_fstorder1, lv_death1, lv_aging1, lv_aging1, lv_aging1],
  LSV = [lsv_birth1],
  P = [p_μ, p_δ, p_rfstOrder, p_rfstOrder, p_δ, p_rage],
  LPV = [lpv_birth2, lpv_death2, lpv_fstorder2, lpv_death2, lpv_fstorder2, lpv_death2, lpv_aging2, lpv_aging2, lpv_aging2],
  Name = name -> nothing, Op=op->nothing, Position=pos->nothing
);
```

```julia
MethodError: no method matching coerce_attrvar_component(::Symbol, ::var"#46#49", ::TypeSet{Symbol}, ::TypeSet{Nothing}, ::Int64, ::Int64)

Closest candidates are:
[...]

PR #839, opened 2023-08-22
Infer LooseACSetTransformation when a Julia function is passed as a component

kris-brown:
`ACSetTransformation(args...)` will try to deduce from the arguments whether it is a `Tight` or `Loose` transformation. It's a safe assumption that a Julia function for an attrtype means one wants a LooseACSetTransformation. This recovers pre-VarACSets behavior. Addresses https://github.com/AlgebraicJulia/Catlab.jl/issues/838

### DIFF
--- a/test/categorical_algebra/CSets.jl
+++ b/test/categorical_algebra/CSets.jl
@@ -310,6 +310,7 @@ uns = naturality_failures(β)
 
 # Loose morphisms.
 α = LooseACSetTransformation((V=[1,2], E=[1]), (Weight=x->x/2,), g, h)
+α = ACSetTransformation(g, h, V=[1,2], E=[1], Weight=x->x/2,)
 @test α isa LooseACSetTransformation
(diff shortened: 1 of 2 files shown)


neonWhiteout:
Thank you!

test/categorical_algebra/CSets.jl
```diff
@@ -310,6 +310,7 @@ uns = naturality_failures(β)
 
 # Loose morphisms.
 α = LooseACSetTransformation((V=[1,2], E=[1]), (Weight=x->x/2,), g, h)
+α = ACSetTransformation(g, h, V=[1,2], E=[1], Weight=x->x/2,)
```
epatters, line 313: This partially negates the previous test. Maybe we can have

```julia
α' = ACSetTransforamtion(...)
@test α' == α
```

epatters (commented):
Thanks Kris!

PR #847, opened 2023-09-07
GATlab integration

olynch:
This is the branch for the ongoing effort to refactor Catlab on top of GATlab.


docs/src/apis/gats.md
```diff
@@ -1,290 +0,0 @@
-# Theories, instances, and expressions
```
epatters, line 1: I know some of this is dated, but it is good content. Is it worth porting the relevant bits to the Gatlab docs? Or do you already have some comparable high-level exposition there?

docs/src/index.md
```diff
@@ -58,29 +58,7 @@ diagrams, Petri nets, and the like.
 
 ### What is a GAT?
 
-Generalized Algebraic Theories (GATs) are the backbone of Catlab so let's expand a bit on GATs and how they fit into the bigger picture of algebra.
```
epatters, line 61: Same question about this content. Also, the link that it is replaced with doesn't (yet?) work.

src/categorical_algebra/ACSetsGATsInterop.jl
```diff
@@ -22,6 +25,21 @@ function Schema(p::Presentation)
   BasicSchema(obs,homs,attrtypes,attrs)
 end
 
+function Schema(c::GATContext)
```
epatters, line 28: If codecov can be believed, this function isn't being run at all. Is it intended to be here?

src/graphs/GraphGenerators.jl
```diff
@@ -73,8 +72,6 @@ function parallel_arrows(::Type{T}, n::Int; V=(;), E=(;)) where T <: ACSet
   g
 end
 
-# Should this be exported from `BasicGraphs`?
-@generated is_directed(::Type{T}) where {S, T<:StructACSet{S}} = :inv ∉ hom(S)
```
epatters, line 77: Thanks for cleaning this up.

src/programs/ParseJuliaPrograms.jl
```diff
@@ -109,15 +109,15 @@ end
 """ Make a lookup table assigning names to generators or term constructors.
 """
 function make_lookup_table(pres::Presentation, syntax_module::Module, names)
-  theory = GATs.theory(syntax_module.theory())
-  terms = Set([ term.name for term in theory.terms ])
+  theory = syntax_module.THEORY_MODULE.Meta.theory
```
epatters, line 112: I wonder whether there should be an accessor function to get the GAT associated with the syntax module, not unlike what we had before, so that we don't have to expose the field `THEORY_MODULE` in all its ugly CAPS.

src/theories/Category.jl
```diff
@@ -102,45 +67,45 @@ end
...
-@theory ThCopresheaf{Ob,Hom,El} <: ThCategory{Ob,Hom} begin
+@theory ThCopresheaf <: ThCategory begin
   # copresheaf = object-indexed family
   El(ob::Ob)::TYPE
 
   # functoriality = covariant action
-  act(x::El(A), f::Hom(A,B))::El(B) ⊣ (A::Ob, B::Ob)
-  @op (⋅) := act
+  act(x::El(A), f::Hom(A,B))::El(B) ⊣ [A::Ob, B::Ob]
+  # @op (⋅) := act
```
epatters, line 76: Is this still unsupported? If so, that's fine but let's make sure to have an issue tracking it if we don't already.

src/theories/HigherCategory.jl
```diff
@@ -21,26 +20,26 @@ abstract type CellExpr{T} <: CategoryExpr{T} end
 
 """ Theory of *2-categories*
 """
-@signature ThCategory2{Ob,Hom,Hom2} <: ThCategory{Ob,Hom} begin
-  """ 2-morphism in a 2-category """
-  Hom2(dom::Hom(A,B), codom::Hom(A,B))::TYPE ⊣ (A::Ob, B::Ob)
+@signature ThCategory2 <: ThCategory begin
+  # """ 2-morphism in a 2-category """
```
epatters, line 24: Ditto for the docstrings: we should be able to capture docstrings, so let's make sure to have an issue tracking it if we can't do that yet.

src/theories/Limits.jl
```diff
@@ -56,32 +56,33 @@ ideas". Strictly speaking, this theory is not of a "finitely complete category"
...
-    ⊣ (A::Ob, B::Ob, f::(A → B), g::(A → B), eq::Equalizer(f,g)))
-  (incl(eq) == id(A)
-    ⊣ (A::Ob, B::Ob, f::(A → B), eq::Equalizer(f,f)))
-  (factorize(eq,h,eq_h) ⋅ incl(eq) == incl(eq_h) ⋅ h
-    ⊣ (A::Ob, B::Ob, C::Ob, f::(A → B), g::(A → B), h::(C → A),
-       eq::Equalizer(f,g), eq_h::Equalizer(h⋅f, h⋅g)))
-  (factorize(eq, k⋅incl(eq), eq_k) == k
-    ⊣ (A::Ob, B::Ob, D::Ob, f::(A → B), g::(A → B), eq::Equalizer(f,g),
-       k::(D → ob(eq)), eq_k::Equalizer(k⋅incl(eq)⋅f, k⋅incl(eq)⋅g)))
+  # (incl(eq)⋅f == incl(eq)⋅g
```
epatters, line 72: What happened to these axioms? I don't immediately see why they are different from others, hence would have broken.

src/theories/MonoidalAdditive.jl
```diff
@@ -150,65 +148,64 @@ end
 
 """ Theory of *monoidal categories with bidiagonals*, in additive notation
 
-Mathematically the same as [`ThMonoidalCategoryWithBidiagonals`](@ref) but
+Mathematically the same as (@ref) but
```
epatters, line 151: Something happened to this docstring

test/graphs/GraphAlgorithms.jl
```diff
@@ -72,4 +74,17 @@ ep = enumerate_paths(g)
...
+
+g = GraphSearching.tree([1, 1, 1, 2, 2])
+g′ = @acset Graph begin
+  V = 5
+  E = 4
+  src = [1, 1, 2, 2]
+  tgt = [2, 3, 4, 5]
+end
+
+@test is_isomorphic(g,g′)
```
epatters, line 88: Thanks for the extra tests!

test/graphs/GraphGenerators.jl
```diff
@@ -49,12 +49,12 @@ g = wheel_graph(Graph, n)
 @test (nv(g), ne(g)) == (n, 2(n-1))
 triangle = Graph(3)
 add_edges!(triangle, [1,2,1], [2,3,3])
-@test length(homomorphisms(triangle, g)) == n-1
+# @test length(homomorphisms(triangle, g)) == n-1
```
epatters, line 52: What's going on here? Seems like these tests should still work.

test/programs/DiagrammaticPrograms.jl
```diff
@@ -0,0 +1,816 @@
+module TestDiagrammaticPrograms
```
epatters, line 1: This whole file has been moved to DataMigrations.jl and seems to have been accidentally added here.

epatters (changes requested):
Huge milestone, congrats on getting here!

I've flagged a few things to look at it but it's generally easy to review a PR that is mostly deleting things :)

docs/src/apis/gats.md
olynch: Yes, I agree, this should be moved to the GATlab docs. We have some comparable high-level exposition, but we might be missing some things.

docs/src/index.md
olynch: Yeah, I'm not sure what's going on with GATlab docs.

src/categorical_algebra/ACSetsGATsInterop.jl
olynch: It makes sense as a function to have, so I think I'll just add a test for it.

src/programs/ParseJuliaPrograms.jl
olynch: Sounds reasonable, we could have a function `syntax_module.theory()` exactly like before.

src/theories/Category.jl
olynch: I'm not sure if we *want* to support this. The way aliases work right now is that they are handled at the scope level, not at the theory level. We'd have to move aliases to the theory level in order to support this, because when something is an alias for two different functions, we'd have to dispatch to resolve which one should be used for a particular invocation of the alias. This isn't used very much in Catlab, so might not be worth complicating the design over.

epatters: That seems like a reasonable position to take. In that case, though, if we're not planning to support it, then the theories in Catlab currently using aliases in this way should be refactored, i.e., choosing different aliases or rewriting the axioms to not use the aliases.

Another idea, which might be too crazy, is to reduce all aliases to term constructors supplemented with equality axioms. Arguably that is more uniform but I could imagine it causing computational annoyances, e.g., alias equalities junking up e-graphs.

src/theories/Limits.jl
olynch: Uncommenting them doesn't break anything, so I uncommented them.

test/graphs/GraphGenerators.jl
olynch: They do!

src/programs/ParseJuliaPrograms.jl
olynch: Here's the question though: should `syntax_module.theory()` return `syntax_module.THEORY_MODULE.Meta.theory`, `syntax_module.THEORY_MODULE` or `syntax_module.THEORY_MODULE.Meta.T`? The closest analogy to Catlab would be `syntax_module.THEORY_MODULE.Meta.T`, but that also seems like the least useful.

epatters (approved):

PR #850, opened 2023-09-16
Bug fix and tests for `to_graphviz` with `Diagram` objects

slwu89:
The method `to_graphviz_property_graph(d::Diagram; kw...)` would previous fail if given a diagram with anonymous ob/homs in the indexing category. It now checks if the graph associated to the indexing category is a `NamedGraph`, if so it will use the names in `vname`, and otherwise empty names are used. This PR also adds tests for both the named and anonymous case, as it seems `to_graphviz` for diagrams was not previously being tested.

### DIFF
--- a/src/graphics/GraphvizCategories.jl
+++ b/src/graphics/GraphvizCategories.jl
@@ -93,7 +93,7 @@ function to_graphviz_property_graph(d::Diagram; kw...)
   pg = to_graphviz_property_graph(g; kw...)
   for v in vertices(g)
     tᵥ = ob_map(d, v)
-    labels = has_vertex_names(g) ? [tᵥ] : [vertex_name(g,v), tᵥ]
+    labels = has_vertex_names(g) ? [vertex_name(g,v), tᵥ] : [tᵥ]
     set_vprop!(pg, v, :label, join(labels, ":"))
   end
   for e in edges(g)
--- a/src/graphs/BasicGraphs.jl
+++ b/src/graphs/BasicGraphs.jl
@@ -23,7 +23,8 @@ export HasVertices, HasGraph, AbstractGraph, Graph, SchGraph,
   add_dangling_edge!, add_dangling_edges!,
   AbstractLabeledGraph, LabeledGraph, SchLabeledGraph,
   AbstractWeightedGraph, WeightedGraph, SchWeightedGraph, weight,
-  AbstractSymmetricWeightedGraph, SymmetricWeightedGraph, SchSymmetricWeightedGraph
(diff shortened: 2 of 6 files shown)


slwu89:
@epatters I've rebased this after #858. Before I request a review, I have a question. As part of this, I am trying to extend the `graph` method defined in `FinCats` to accept a `Presentation` and make a `NamedGraph`. This will help simplify the graphviz code, and its also nice for other non graphics related programming. 

However, I cannot figure out why when I uncomment this line to import the method https://github.com/slwu89/Catlab.jl/blob/graphviz-diagram/src/graphs/NamedGraphs.jl#L14, Julia complains with the error `ERROR: LoadError: UndefVarError: `CategoricalAlgebra` not defined`. Do you have any idea how I can extend that method in the named graphs module?

epatters:
The reason is that CategoricalAlgebra is loaded in Catlab after Graphs, i.e., CategoricalAlgebra depends on Graphs, not the other around. So any interoperability layer between the two should go in CategoricalAlgebra.

In fact, I think the function you are looking for is already available on main: https://github.com/slwu89/Catlab.jl/blob/54b8e25ef13a7e125f36e687f02c58b9f1646cd1/src/categorical_algebra/FinCats.jl#L113

slwu89:
Ah, thanks @epatters for clearing that up!

Yes I did see that `graph` method, but I think it only works for a `FinCat`or a `FinCat` presented by an actual graph object. I was interested in implementing something that would work for a `Presentation`, like:
```julia
julia> graph(SchWeightedGraph)
ERROR: MethodError: no method matching graph(::Presentation{ThSchema, Symbol})
```

But I think that should wait for another PR and this one should just get those tests for visualizing a diagram in.

epatters:
IIRC, you can just wrap the Presentation in a FinCat, i.e., call `graph(FinCat(my_presentation))`.

src/graphics/GraphvizCategories.jl
```diff
@@ -93,7 +93,7 @@ function to_graphviz_property_graph(d::Diagram; kw...)
   pg = to_graphviz_property_graph(g; kw...)
   for v in vertices(g)
     tᵥ = ob_map(d, v)
-    labels = has_vertex_names(g) ? [tᵥ] : [vertex_name(g,v), tᵥ]
+    labels = has_vertex_names(g) ? [vertex_name(g,v), tᵥ] : [tᵥ]
```
epatters, line 96: Haha, this is why it's good to have tests. Thanks!

epatters (approved):
LGTM, thanks Sean!

PR #852, opened 2023-09-16
Add Sheaf implementation 

jpfairbanks:
This PR adds a module for sheaves, including abstract types. It has an example for free vector space as a sheaf over Diag(FinSet).


jpfairbanks:
@tylerhanks, can you review this? I want to add an example for induced subgraphs next.

benjaminmerlinbumpus (approved):

PR #854, opened 2023-09-22
Unbreak `makemap` from `attrMigr` branch

KevinArlin:
I made `make_map` force the output to a specific type while writing `attrMigr`, a functionality which can cause all kinds of problems and was only actually being used in one quite narrow case. This PR resets it to the original functionality of only forcing the type for empty outputs.


epatters (approved):
Thanks Kevin!

PR #856, opened 2023-09-28
Export named graphs from `Catlab.Graphs`

epatters:
CC @slwu89


slwu89:
Export all the things! Thanks. After you merge this I'll update my #850 before requesting review.

epatters:
Now that `vertex_name` and `edge_name` are part of the official API, we should update `GraphvizGraphs` and `GraphCategories` to use those by default, which will work for named graphs while preserving the old behavior unnamed graphs.

slwu89:
Ok, happy to do that, as it'll help figure out how to adjust #850.

slwu89:
@epatters I'm actually not sure what's the best way forward on updating `GraphvizGraphs` to use `vertex_name` and `edge_name`. Currently there's a lot of flexibility, as the argument `node_labels` can be boolean or a symbol referring to a subpart. For example, if we pass a `NamedGraph` to `to_graphviz` we have the option of printing the vertex ID, a subpart associated to vertices, or nothing. It would seem that using `vertex_name` in place of the method `node_label` for example, would mean we are forced to use the `vname` subpart.

PR #857, opened 2023-09-29
Make pointed-set categories and schemas actually pointed

KevinArlin:
I original left out zero objects in `PtdSetCategory` and `PtdSetSchema` but now I want the `Attr` side to be allowed to have a zero object so that I can pull attrs out of a hat that aren't in the source schema; I'm not sure whether I really need the zero object in the combinatorial side yet but why tempt fate?

### DIFF
--- a/src/theories/Category.jl
+++ b/src/theories/Category.jl
@@ -3,7 +3,7 @@ export ThCategory, FreeCategory, Ob, Hom, dom, codom, id, compose, ⋅,
   ThCopresheaf, FreeCopresheaf, El, ElExpr, ob, act,
   ThPresheaf, FreePresheaf, coact,
   ThMCategory, FreeMCategory, Tight, reflexive, transitive,
-  ThPointedSetCategory, FreePointedSetCategory, zeromap
+  ThPointedCategory, FreePointedCategory, zeromap, ZeroOb
 
 import Base: inv, show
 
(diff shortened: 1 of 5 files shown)


src/theories/Category.jl
```diff
@@ -200,21 +200,27 @@ abstract type TightExpr{T} <: GATExpr{T} end
...
-equivalently, a pointed-set enriched category in which no two nonzero maps 
+pointed category, 
+equivalently, a pointed category in which no two nonzero maps 
 compose to a zero map, is a good notion
-of a functor that's total on objects and partial on morphisms.
+of a partial functor.
 """
-@theory ThPointedSetCategory{Ob,Hom} <: ThCategory{Ob,Hom} begin
+@theory ThPointedCategory{Ob,Hom} <: ThCategory{Ob,Hom} begin
+  ZeroOb()::Ob
```
epatters, line 215: Our convention for term constructors in `Catlab.Theories` is that they are lowercase.

src/theories/Category.jl
```diff
@@ -200,21 +200,27 @@ abstract type TightExpr{T} <: GATExpr{T} end
...
   zeromap(A,B)::Hom(A,B)⊣(A::Ob,B::Ob)
   compose(zeromap(A,B),f::(B→C))==zeromap(A,C)⊣(A::Ob,B::Ob,C::Ob)
   compose(g::(A→B),zeromap(A,B))==zeromap(A,C)⊣(A::Ob,B::Ob,C::Ob)
+
+  #f == g ⊣ (A::Ob,f::Hom(A,ZeroOb()),g::Hom(A,ZeroOb()))
+  #f == g ⊣ (B::Ob,f::Hom(ZeroOb(),B),g::Hom(ZeroOb(),B))
```
epatters, line 221: Why have commented out axioms?

epatters (commented):
Thanks for the PR. With the idea that `Catlab.Theories` should contain a hierarchy of plausibly useful theories even if we don't immediately need them, I might suggest having:

1. `ThPointedSetCategory` for a category enriched in pointed sets
2. `ThPointedCategory` for a pointed category inherits from `ThPointedSetCategory`, adding the zero objects.

src/theories/Category.jl
jpfairbanks: Documentation on your documentation.

PR #868, opened 2023-12-02
Make sure you can add SetFunctionCallables to BipartiteDiagrams during limit computation

KevinArlin:


src/categorical_algebra/FinSets.jl
```diff
@@ -925,7 +929,7 @@ end
 ensure_type_set(s::FinSet) = TypeSet(eltype(s))
 ensure_type_set(s::TypeSet) = s
 ensure_type_set_codom(f::FinFunction) =
-  FinDomFunction([f(i) for i in dom(f)], dom(f), TypeSet(eltype(codom(f))))
+  FinDomFunction(f.func,dom(f), TypeSet(eltype(codom(f))))
```
epatters, line 932: This can't be correct since `FinFunction` is an abstract type and having a `func` field is not part of its interface.

epatters (changes requested):
Thanks for the fix!

src/categorical_algebra/FinSets.jl
KevinArlin: Well, every implementation so far has a `func`. How about we add it to the interface? It seems reasonable to assume a `FinFunction` contains some kind of function.

KevinArlin: The previous two versions of this just brutally cast to either a `SetFunctionCallable` or a `FinDomFunctionVector`, fwiw.

epatters: Granted that Julia makes it rather ill-specified what the interface is, the interface that `FinFunction` is *supposed* to have is that the domain `dom(f)` and codomain `codom(f)` are `FinSet`s and that `f` can be evaluated at every element in the domain. That is a very minimal interface, but the expression `[f(i) for i in dom(f)]` is compatible with it. Actually, now that I think it about, that expression is already available as `collect(f)`.

So I would say that

```julia
FinDomFunction(collect(f), dom(f), TypeSet(eltype(codom(f))))
```

is reasonable since the function is going to be evaluated at every point anyway while computing the limit.

KevinArlin:
OK, this should be ready now.

epatters (approved):
Thanks Kevin!

PR #873, opened 2023-12-11
Constrain acset hom search to specified subsets

kris-brown:
This constraint allows one to say, e.g. when searching for maps from the path graph with two edges into some big graph, that V2 *must* be sent to either vertex 2, 5, or 6 in the codomain, and that E1 *must* be sent to edge 3 or 7 in the codomain, etc. I.e. we have the option to restrict valid assignments to an arbitrary subset of the codomain for each part in the domain.

This is a required feature for my implementation of [incremental hom search](https://github.com/AlgebraicJulia/AlgebraicRewriting.jl/pull/36).

This is technically a generalization of the `initial` keyword argument, which would be `valid` assigning a singleton set to some part of the domain. However, the way that case is treated (immediately beginning hom search with that assignment) is very different from the general case (a check when making an assignment arbitrarily deep in the tree), so it's helpful to be able to declare these separately.

### DIFF
--- a/src/categorical_algebra/HomSearch.jl
+++ b/src/categorical_algebra/HomSearch.jl
@@ -77,11 +77,11 @@ default, a backtracking search algorithm is used ([`BacktrackingSearch`](@ref)).
 Use the keyword argument error_failures = true to get errors explaining 
 any immediate inconsistencies in specified initial data.
 
-The keyword `valid` accepts a Dict{Ob->Dict{Int->Union{Nothing, Set{Int}}}}
-For each part of the domain, we have the option to give a constraint stating 
-which parts of the codomain are allowable values for assignment. E.g. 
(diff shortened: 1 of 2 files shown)


epatters (commented):
Thanks Kris! Since in the end the implementation just adds an extra predicate to check, it might be cleaner and slightly more flexible to call the keyword argument `predicates` and have it take a dict of dicts of predicates (boolean-valued Julia functions). Then to recover the current behavior, you pass the lambda `x -> x ∈ my_set`. What do you think?

kris-brown:
Yeah I like this idea a lot!

epatters:
I agree that in practice this will at most save you a coercion to a set since the input data is finitary, but it might still be nice to not insist on a single data type. Your call.

kris-brown:
The `Set` coercion is allowing the user to pass in any `AbstractVector` at least, right? Or they could write a `Set(::MyStruct)` method.

epatters (approved):

PR #881, opened 2024-01-11
Adding missing `compose` and `id` methods for pointed set schemas

KevinArlin:
Just needed to expand a signature for the method of `id` relevant to attrtypes to make sure everything in a category presented by a schema with zero maps still has an identity morphism.


epatters (approved):
Thanks! I'll merge once the tests pass.

PR #883, opened 2024-01-23
Rename `Slices` to `SliceCategories` to avoid name conflict

KevinArlin:
Julia complains all the time now about Catlab and Base both exporting the word `Slices`. Can we just change the name? This would be a breaking change, not sure whether there's a better approach but I think leaving it as is isn't a serious option in that the attempted export wouldn't actually work if you tried to use `Slices` unqualified.

### DIFF
--- a/src/categorical_algebra/CategoricalAlgebra.jl
+++ b/src/categorical_algebra/CategoricalAlgebra.jl
@@ -21,7 +21,7 @@ include("CatElements.jl")
 include("Chase.jl")
 include("FunctorialDataMigrations.jl")
 include("StructuredCospans.jl")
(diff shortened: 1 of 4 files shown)


epatters:
Yes, we should change the name. I would suggest renaming the module to `SliceCategories`.

I'm not sure why this PR deletes the modules.

KevinArlin:
I think I just didn't add the renamed files on the first commit?

epatters:
Thanks! Can you please rebase and force-push so we get a clean commit?

KevinArlin:
Yep, done.

epatters (approved):
Great, thanks Kevin!

PR #887, opened 2024-02-16
Enumerate subobjects of acsets with attribute variables

kris-brown:
Previously, subobject enumeration would error if performed on an ACSet with AttrVars. 

This PR implements the following behavior: when a subobject is taken such that certain AttrVars are no longer referred to, we remove those AttrVars from the subobject, too.


epatters (approved):
Thanks Kris!

PR #890, opened 2024-02-23
Fixed typo.  "commute must" -> "must commute".

neonWhiteout:
Those words don't commute.
Minor text edit, so not that big a deal.


epatters:
> Those words don't commute.

Lol! Thanks for the fix.

PR #900, opened 2024-04-19
Smoothing interaction between VarSets and actual sets and functions

KevinDCarlson:
The changes below are to enable smoother data migrations of acsets with some `AttrVar` attribute values as requested here: https://github.com/AlgebraicJulia/DataMigrations.jl/issues/159


kris-brown (approved):
These seem like reasonable casting functions! LGTM

KevinDCarlson:
@epatters Could you merge this and do a patch release?

epatters (changes requested):
Though the additions are simple, if they are to be part of the public API of Catlab, then they should have tests, which is what will ensure that they remain stable over time. I think the important thing to add tests for is the conversion between SetOb/FinSets and VarSets.

epatters (approved):
Thanks Kevin! I'll merge once the CI finishes.

PR #901, opened 2024-04-23
Missing ACSetsGATsInterop methods

kris-brown:
The current ACSetsGATsInterop file does not indicate how `subpart` called with a specific part (e.g. `Int`, `Vector{Int}`, `Colon`) behaves with respect to `id` or `generator` GATExprs. This PR adds this in.

This is necessary for compatibility with https://github.com/AlgebraicJulia/ACSets.jl/pull/130


epatters (approved):
Thanks for fixing this up, Kris!

Issue #903, opened 2024-04-26
`SigmaMigrationFunctor` method uses an undefined variable


neonWhiteout:
SigmaMigrationFunctor(f,T′,T) = SigmaMigrationFunctor(f,d,T()) on line 193 of FunctorialDataMigrations.jl

epatters:
Thanks for the report. @kris-brown, could you take a look at this one?

kris-brown:
Hi @neonWhiteout, can you share what context you called this in? This method seems quite extraneous (no type information on the parameters) and the tests pass without it, so I'm leaning towards fixing this bug by just deleting the method. 

You could also test your code works in the branch here, which has the method deleted: https://github.com/AlgebraicJulia/Catlab.jl/pull/906

PR #906, opened 2024-04-28
Delete extraneous SigmaMigration method

kris-brown:
See https://github.com/AlgebraicJulia/Catlab.jl/issues/903


epatters (approved):

epatters:
I will go ahead and merge since we haven't heard back from the OP. If problems persist we can revisit.

PR #909, opened 2024-05-19
Fix Heyting subtraction for ACSets with variables

kris-brown:
The subtraction operation was not brought up to date with AttrVar refactor, so a minor tweak is needed.

It uses `all_subparts`, which gets all the other parts that a part is referencing; however, that method searched over `homs` and not `arrows`. When this change is made, the correct behavior is recovered.

The relevance of this is that complement is used in this PR https://github.com/AlgebraicJulia/AlgebraicRewriting.jl/pull/62


epatters (approved):
Thanks Kris! Glad somebody is using this stuff :)

PR #911, opened 2024-05-25
CompatHelper: bump compat for Convex in [weakdeps] to 0.16, (keep existing compat)

algebraicjuliabot:
This pull request changes the compat entry for the `Convex` package from `0.15` to `0.15, 0.16`.
This keeps the compat entries for earlier versions.

Note: I have not tested your package with this new compat entry.
It is your responsibility to make sure that your package tests pass before you merge this pull request.


epatters:
Looks like it's actually incompatible and a fix will be needed.

PR #914, opened 2024-06-11
Code simplification from equations in schema

kris-brown:
After https://github.com/AlgebraicJulia/ACSets.jl/pull/142 is merged and a new version of ACSets is released, we can simplify code that requires equations (`yoneda`, `FinDomFunctor`, `representable`, `subobject_classifier`, `internal_hom`), which would normally require passing in a `Presentation` object in addition to an ACSet / ACSet type.


epatters (approved):
Thanks Kris, this is a nice cleanup!

PR #915, opened 2024-06-21
Redesign ACSetFunctors when there are AttrVars

KevinDCarlson:
Modifications to make attrvars play nicer with data migrations. The key change in CSets is to redefine `ACSetFunctor`s, which are used critically in data migrations to compose with ordinary functors representing the migrations themselves, to again always take values in `FinSetInt`s and `FinDomFunction`s when possible. Previously, the `ACSetFunctor` of an `ACSet` containing any `Attrvar`s would take values entirely in `VarSet`s, which both confused the difference between actual `Attrvar`s and combinatorial data and interfaced badly with the `Limit`s API. 

To make this work I also defined how to compose `FinFunction`s with `VarFunction`s on the left.


epatters (approved):
Thanks Kevin! This change makes good sense to me.

I'll merge once the conflict is fixed.

KevinDCarlson:
Hm, dunno what that was about but I've fixed it.

PR #917, opened 2024-06-26
Fix constructor of acset from `FinDomFunctor` in varacset case

KevinDCarlson:
The constructor `(T)(::FinDomFunctor) where T<:ACSet` did not previously add any parts at `AttrType`s, which is necessary in case where some of the input's attributes take `AttrVar` values. We now add such `AttrVar`s, only when they're definitely needed, to avoid dangling `AttrVar`s, which are quite dangerous.


epatters (approved):
Thanks Kevin!

PR #919, opened 2024-06-28
Fix construction of varacset from `FinDomFunctor`

KevinDCarlson:
`(::Type{ACS})(F::FinDomFunctor)` would previously give back an `ACS{Any}` if `ob_map(F,:someAttr)` was a `VarSet` and an `ACS{Union{AttrVar,T}}` if `ob_map(F,:someAttr)` was `TypeSet{Union{AttrVar,T}}`.


epatters:
Thanks for the fix, Kevin!

Confused about what is going on the CI here: the tests haven't run for the latest commit.

epatters (approved):

PR #921, opened 2024-07-01
Integer-typed UWDs

samuelcohen1:
Adds additional compatibility to typed UWDs to allow for Int types and Expr types. (Currently, only Symbol types are supported.)

### DIFF
--- a/src/programs/RelationalPrograms.jl
+++ b/src/programs/RelationalPrograms.jl
@@ -145,7 +145,7 @@ function parse_relation_diagram(head::Expr, body::Expr)
     Expr(:where, expr, context) => (expr, parse_relation_context(context)...)
     _ => (head, nothing, nothing)
   end
-  var_types = if isnothing(all_types)    # Untyped case.
+  var_types = if isnothing(all_types)    # Untyped case
     vars -> length(vars)
   elseif typeof(all_types[1]) <: Int     # Int typed case
     var_type_map = Dict{Symbol,Int}(zip(all_vars, all_types))
@@ -255,6 +255,7 @@ function parse_relation_inferred_args(args)
       Expr(:kw, name::Symbol, var::Symbol) => (name => var)
       Expr(:(=), name::Symbol, var::Symbol) => (name => var)
       var::Symbol => var
+      Expr(:(::), _, _) => error("All variable types must be included in the where clause and not in the argument list")
       _ => error("Expected name as positional or keyword argument")
     end
   end
--- a/src/programs/RelationalPrograms.jl
+++ b/src/programs/RelationalPrograms.jl
@@ -149,17 +149,12 @@ function parse_relation_diagram(head::Expr, body::Expr)
   var_types = if isnothing(all_types) # Untyped case.
     vars -> length(vars)
   elseif typeof(all_types[1]) <: Int # Int typed case
-    println("int typed case")
     var_type_map = Dict{Symbol,Int}(zip(all_vars, all_types))
     vars -> getindex.(Ref(var_type_map), vars)
   elseif typeof(all_types[1]) <: Expr # Int typed case
-    println("expr typed case")
     var_type_map = Dict{Symbol,Expr}(zip(all_vars, all_types))
     vars -> getindex.(Ref(var_type_map), vars)
   else # Symbol typed case.
-    println("symbol typed case")
-    println(typeof(all_types))
-    println(all_types)
     var_type_map = Dict{Symbol,Symbol}(zip(all_vars, all_types))
     vars -> getindex.(Ref(var_type_map), vars)
 
@@ -199,7 +194,6 @@ function parse_relation_diagram(head::Expr, body::Expr)
 end
 
 function parse_relation_context(context)
-  println("here")
   terms = @match context begin
     Expr(:tuple) => return (Symbol[], nothing)
     Expr(:tuple, terms...) => terms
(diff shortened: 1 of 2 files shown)


src/programs/RelationalPrograms.jl
```diff
@@ -145,9 +145,15 @@ function parse_relation_diagram(head::Expr, body::Expr)
     Expr(:where, expr, context) => (expr, parse_relation_context(context)...)
     _ => (head, nothing, nothing)
   end
-  var_types = if isnothing(all_types) # Untyped case.
+  var_types = if isnothing(all_types)    # Untyped case
     vars -> length(vars)
-  else # Typed case.
+  elseif typeof(all_types[1]) <: Int     # Int typed case
```
jpfairbanks, line 150: you should use `eltype(all_types) <: Integer` here. Int is a concrete type so it only has itself as subtype. Int32 and Int64 are both subtypes of Integer. The `eltype` function get the element type of the array, which is more reliable than the type of the first element.

src/programs/RelationalPrograms.jl
```diff
@@ -145,9 +145,15 @@ function parse_relation_diagram(head::Expr, body::Expr)
...
+    var_type_map = Dict{Symbol,Int}(zip(all_vars, all_types))
+    vars -> getindex.(Ref(var_type_map), vars)
+  elseif typeof(all_types[1]) <: Expr    # Expr typed case
```
jpfairbanks, line 153: eltype here too. but I'm thinking you might get a Union{Symbol, Expr} or something like that.

src/programs/RelationalPrograms.jl
```diff
@@ -194,19 +200,26 @@ function parse_relation_context(context)
...
     (vars, nothing)
   elseif vars isa AbstractVector{Pair{Symbol,Symbol}}
     (first.(vars), last.(vars))
+  elseif vars isa AbstractVector{Pair{Symbol, Int}}
+    (first.(vars), last.(vars))
+  elseif vars isa AbstractVector{Pair{Symbol, Expr}}
+    (first.(vars), last.(vars))
   else
-    error("Context $context mixes typed and untyped variables")
+    error("Context $context mixes variable types")
```
jpfairbanks, line 218: Only risk I see here is that when you have expressions, you might want to treat a symbol as the expression containing that symbol. Like how you might want to treat an Int as the list containing only that Int in some contexts. So there might be a valid input that comes out as `AbstractVector{Pair{Symbol, Union{Symbol,Expr}}}`

test/programs/RelationalPrograms.jl
```diff
@@ -66,6 +69,47 @@ set_junction!(d, [1,4,2,4,3,4])
...
+#------
+
+parsed = @relation (x,y,z) where (x::n(1), y::n(2), z::n(3), w::n(4)) begin
+  R(x,w)
+  S(y,w)
+  T(z,w)
+end
+
+# Testing doesn't work right now because Julia doesn't like calling n as a function
+# d = RelationDiagram([Expr(n(1)), Expr(n(1)), Expr(n(1))])
```
jpfairbanks, line 103: replace `n(1)` with `:(n(1))`.

test/programs/RelationalPrograms.jl
```diff
@@ -66,6 +69,47 @@ set_junction!(d, [1,4,2,4,3,4])
...
+# add_box!(d, [:n(1),:n(4)], name=:R)
+# add_box!(d, [:n(2),:n(4)], name=:S)
+# add_box!(d, [:n(3),:n(4)], name=:T)
+# add_junctions!(d, [:n(1),:n(2),:n(3),:n(4)], variable=[:x,:y,:z,:w])
+# set_junction!(d, [1,4,2,4,3,4])
+# set_junction!(d, [1,2,3], outer=true)
+# @test parsed == d
+
```
jpfairbanks, line 111: Mixed Symbol and expression case.

```julia
parsed = @relation (x,y,z) where (x::X, y::n(2), z::n(3)+X, w::n(4)) begin
  R(x,w)
  S(y,w)
  T(z,w)
end
```
This will probably error because of the overly strict typing above.

Mixed symbol, number, expression case:


```julia
parsed = @relation (x,y,z) where (x::X, y::n(2), z::n(3)+X, w::4) begin
  R(x,w)
  S(y,w)
  T(z,w)
end
```

This will definitely error because of the way you are checking element types. Basically the expression case should be handling any subtype of `Union{Symbol, Int, Expr}` as the element type. And then just think of :foo or :(4) as the expression containing a symbol or number.

jpfairbanks (changes requested):
Looks good. I think you just need to change some of the type constraints to handle some tests I suggested.

jpfairbanks (changes requested):

src/programs/RelationalPrograms.jl
```diff
@@ -242,6 +255,7 @@ function parse_relation_inferred_args(args)
       Expr(:kw, name::Symbol, var::Symbol) => (name => var)
       Expr(:(=), name::Symbol, var::Symbol) => (name => var)
       var::Symbol => var
+      Expr(:(::), _, _) => error("All variable types must be included in the where clause and not in the argument list")
```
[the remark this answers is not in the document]
samuelcohen1: Just a better error message for a usage pattern that was already broken.

jpfairbanks (approved):
LGTM! Go @samuelcohen1

jpfairbanks:
This is mergable as soon as tests pass!

epatters (approved):
Thanks all for this feature! I'm happy to see this merged once the CI passes.

Issue #913, opened 2024-06-05
Parameters in queries for attributes that are unions


slwu89:
When an `AttrType` is being modeled with some union type, queries with a parameter for something pointing to that attribute type won't work (errors). I am not sure if its possible to make them work given how the spans are set up in `oapply`, but I am documenting it here.

```julia
using Catlab, DataFrames

@present TestSch(FreeSchema) begin
    X::Ob
    A::AttrType
    a::Attr(X,A)
end

@acset_type TestData(TestSch)

data = @acset TestData{Union{Int,Symbol}} begin
    X=5
    a=[1,2,3,:four,:five]
end

testquery = @relation (x=xid, a=attr) begin
    X(_id=xid, a=attr)
end

query(data, testquery)
query(data, testquery, (attr=3, ))
```

slwu89:
The error is as follows. I guess a possible fix would be to do something here https://github.com/AlgebraicJulia/Catlab.jl/blob/6c7c935f749775f017ba6220df655f1407280984/src/wiring_diagrams/Algebras.jl#L169-L179 to make sure the type sets are correct.

```julia
julia> query(data, testquery, (attr=3, ))
ERROR: Feet of spans are not equal: TypeSet(Int64) != TypeSet(Union{Int64, Symbol})
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] oapply(composite::Catlab.Programs.RelationalPrograms.UntypedNamedRelationDiagram{…}, spans::Vector{…}; Ob::Type, Hom::Type, return_limit::Bool)
   @ Catlab.WiringDiagrams.WiringDiagramAlgebras ~/.julia/packages/Catlab/YRhzN/src/wiring_diagrams/Algebras.jl:54
 [3] query(X::TestData{…}, diagram::Catlab.Programs.RelationalPrograms.UntypedNamedRelationDiagram{…}, params::@NamedTuple{…}; table_type::Type)
   @ Catlab.WiringDiagrams.WiringDiagramAlgebras ~/.julia/packages/Catlab/YRhzN/src/wiring_diagrams/Algebras.jl:182
 [4] query(X::TestData{…}, diagram::Catlab.Programs.RelationalPrograms.UntypedNamedRelationDiagram{…}, params::@NamedTuple{…})
[...]

PR #923, opened 2024-07-09
Fix for queries with attributes using union types

slwu89:
Addresses #913, happy for comments.


epatters:
Thanks for the fix, @slwu89!

This feels like more work than it ought to be, but I can't see a shorter fix. Union types in Julia are a frequent source of bugs in my experience.

epatters (approved):

PR #924, opened 2024-07-10
Bugfix for complement for VarACSets

kris-brown:
A recent change which makes iterating over VarSets return AttrVars rather than just integers introduced a bug in the complement operation for VarACSets (which gets used in AlgebraicRewriting/AlgebraicABMs). This fixes that and adds a test which calls complement on a VarACSet.


epatters (approved):
Thanks for the fix!

PR #926, opened 2024-07-15
Require `homomorphism` to be unique

kris-brown:
This PR involves refactoring the `backtracking_search` interface, which now accepts `take`, `max`, and `filter` keyword arguments.

- `take` = number of homomorphisms requested (stop the search process early if this number is reached)
- `max` = throw an error if we find more than this many morphisms (e.g. set `max=1`  if one expects 0 or 1 morphism)
- `filter` = only consider morphisms which meet some criteria, expressed as a Julia function of type `ACSetTransformation -> Bool`

This allows `homomorphism` to be redefined as:

```julia
function homomorphism(X::ACSet, Y::ACSet, alg::BacktrackingSearch; any=false, kw...)
  res = homomorphisms(X, Y, alg; Dict((any ? :take : :max) => 1)..., kw...)
  isempty(res) ? nothing : only(res)
end
```

The old behavior is recovered via `any=true`, but this should not be default as it leads to endless user confusion. Thus there is a version bump for this breaking change.

The postprocessing logic for solutions to backtracking search (introduced recently in https://github.com/AlgebraicJulia/Catlab.jl/pull/918) was simplified and moved into its own function to not clutter `backtracking_search`.

Note: `filter` is needed for certain AlgebraicRewriting features, such as trying to get a single (but arbitrary) homomorphism subject to certain arbitrary constraints, such as the dangling condition, NACs, etc. 

Also note: there was an unrelated bug in `coerce_component` for MarkAsDeleted ACSets which was unearthed in this process.

### DIFF
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -453,8 +453,12 @@ end
   
 function coerce_component(ob::Symbol, f::FinFunction{Int,Int},
                           dom_size::Int, codom_size::Int; kw...)
-  length(dom(f)) == dom_size || error("Domain error in component $ob")
-  # length(codom(f)) == codom_size || error("Codomain error in component $ob") # codom size is now Maxpart not nparts
+  if haskey(kw, :dom_parts)
+    !any(i -> f.func[i] == 0, kw[:dom_parts]) # check domain of mark as deleted
+  else                         
+    length(dom(f)) == dom_size # check domain of dense parts
+  end || error("Domain error in component $ob")
+  # length(codom(f)) == codom_size || error("Codomain error in component $ob")
   return f 
 end
 
--- a/src/categorical_algebra/CSets.jl
+++ b/src/categorical_algebra/CSets.jl
@@ -1123,9 +1123,10 @@ end
 const SubCSet{S} = Subobject{<:StructCSet{S}}
 const SubACSet{S} = Subobject{<:StructACSet{S}}
 
-# Componentwise subobjects
+# Componentwise subobjects: coerce VarFunctions to FinFunctions
 components(A::SubACSet{S}) where S = 
-  NamedTuple(k => Subobject(vs) for (k,vs) in pairs(components(hom(A)))
+  NamedTuple(k => Subobject(k ∈ ob(S) ? vs : FinFunction(vs)) for (k,vs) in 
+             pairs(components(hom(A)))
 )
 
 force(A::SubACSet) = Subobject(force(hom(A)))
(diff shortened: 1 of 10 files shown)


src/categorical_algebra/CSets.jl
```diff
@@ -453,8 +453,12 @@ end
...
 function coerce_component(ob::Symbol, f::FinFunction{Int,Int},
                           dom_size::Int, codom_size::Int; kw...)
-  length(dom(f)) == dom_size || error("Domain error in component $ob")
-  # length(codom(f)) == codom_size || error("Codomain error in component $ob") # codom size is now Maxpart not nparts
+  if haskey(kw, :dom_parts)
+    !any(i -> f(i) == 0, kw[:dom_parts]) # check domain of mark as deleted
+  else                         
+    length(dom(f)) == dom_size # check domain of dense parts
+  end || error("Domain error in component $ob")
+  # length(codom(f)) == codom_size || error("Codomain error in component $ob")
```
epatters, line 461: Looks like this code has been dead for some time now. Can we just delete it?

src/categorical_algebra/HomSearch.jl
```diff
@@ -115,10 +116,29 @@ homomorphisms exist, it is exactly as expensive.
 homomorphisms(X::ACSet, Y::ACSet; alg=BacktrackingSearch(), kw...) =
   homomorphisms(X, Y, alg; kw...)
 
-function homomorphisms(X::ACSet, Y::ACSet, alg::BacktrackingSearch; kw...) 
+"""
```
epatters, line 119: Thanks for documenting this method! Can we add a line the beginning, before documenting the optional arguments, stating what the method does?

src/categorical_algebra/HomSearch.jl
```diff
@@ -509,6 +506,48 @@ unassign_elem!(state::BacktrackingState{<:DynamicACSet}, depth, c, x) =
...
+iterate through this product at the very end rather than having the backtracking
+search navigate the product space. Currently, this is only done with assignments
+for floating attribute variables, but in principle this could be applied in the
+future to, e.g., free-floating vertices of a graph or other coproducts of 
+representables.
+
+This function takes a result assignment from backtracking search and returns an
+iterator of the implicit set of homomorphisms that it specifies.
+"""
+function postprocess_res(dom, codom, assgn, monic)
```
epatters, line 522: Would suggest a slightly more informative/verbose name, e.g., `postprocess_search_results`.

epatters (changes requested):
Thanks Kris, the new behavior should be much more in line with users' expectations.

src/categorical_algebra/CSets.jl
kris-brown: Yeah, I'll add a comment that (for whatever reason) we check domains but not codomains in `coerce_component`

epatters (approved):

PR #932, opened 2024-08-16
Loosen type in `to_graphviz_property_graph`

slwu89:
The type in `to_graphviz_property_graph` is too narrow, the function calls down to the `PropertyGraph` constructor which is defined for type `HasGraph`.


epatters (approved):
Thanks Sean!

PR #933, opened 2024-08-23
Cm/buildkite

quffaro:
Changing pipeline.yml

### DIFF
--- a/.buildkite/jobscript.sh
+++ b/.buildkite/jobscript.sh
@@ -1,0 +1,9 @@
+pwd; hostname; date
+
+module load julia
+
+echo "Running Tests..."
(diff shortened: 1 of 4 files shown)


quffaro:
@jpfairbanks the Buildkite CI now passes here with the new dynamic pipeline and build matrix features. The former lets us make organization-wide changes to the build pipeline, while the build matrix lets us test on different combinations of environment variables, which for now is just the Julia version.

This branch allows our CI/CD to use HPC to build/test.

jpfairbanks:
Tick the task list items off that are true and LGTM

epatters:
So is the plan to make the benchmarks CI action (which is currently very slow) run in Buildkite? I don't mind merging this PR as long as the CI passes but I don't know what it's actually doing.

PR #941, opened 2024-10-11
Fix empty collection typing bug

KevinDCarlson:
Commit 3f95da1 introduced a regression for constructing Yoneda in the case of a schema that has an `AttrTypes` with no `Attrs` pointing to it, which surprisingly enough is sometimes useful (in case you want to migrate from a schema with `AttrTypes` to one without for migrations including code.) The problem was that `union` has no method for empty input since it can't decide on the output type, because typing Julia collections in empty cases is our entire job here. I fixed it.


epatters:
Tests are failing for an unrelated reason (#942).

epatters (approved):

PR #944, opened 2024-10-16
WIP: Cscaff/peg uwd

jpfairbanks:
This PR migrates the UWD grammar from SyntacticModels.jl to upstream Catlab.


jpfairbanks:
@cscaff, this branch has conflicts with main. Please fix them so that we can merge this.

cscaff:
> @cscaff, this branch has conflicts with main. Please fix them so that we can merge this.

Sounds good. I'll get on that and update you.

jpfairbanks:
It looks like the build failed on linux. It can't find the ADTs.jl file, I know that paths are case sensitive on linux, but not windows and mac, so that might be the problem.

quffaro:
That's the issue. Changing to the include statement in the main file to ```include("ADTs/ADTs.jl")``` will resolve this issue. I don't have permission to push the fix @cscaff but it's a small change.

cscaff:
Ok, I think I've resolved it. I made that change in the main src and test files. Looks like everything still runs on my end on Windows 10.

test/runtests.jl
quffaro, line 1: this change was not necessary because it correctly specified the path of the file, but the folders are inconsistently capitalized.

quffaro (approved):

PR #947, opened 2024-10-22
Floating viewer for Graphviz output

quffaro:
Since I mainly use Vim to develop, I don't have an embedded file viewer like VSCode so I wrote this snippet to visualize graphs in a floating window. But I'm opening this draft PR to open the discussion whether it should be upstreamed, and if so, what technical considerations.

I'm aware that `isdefined(Main, :VSCodeServer)` allows us to check if the REPL is being run in VSCode (and likewise for Juypter and IJulia), but I didn't write a validation to check that as I assumed a user would call this function in order to visualize the graph in a floating window.

1. Should `run` statements be used in code?
2. What other considerations besides operating system should a viewer check for? `feh` for example is not an image viewer everyone uses.
3. How can this code be generalized?
4. How can it be tested?

### DIFF
--- a/src/graphics/Graphviz.jl
+++ b/src/graphics/Graphviz.jl
@@ -7,7 +7,7 @@ References:
 """
 module Graphviz
 export Expression, Statement, Attributes, Graph, Digraph, Subgraph,
-  Node, NodeID, Edge, Label, pprint, run_graphviz
+  Node, NodeID, Edge, Label, pprint, run_graphviz, view_graphviz
 
 using DataStructures: OrderedDict
 using StructEquality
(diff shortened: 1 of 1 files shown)


src/graphics/Graphviz.jl
```diff
@@ -275,4 +275,18 @@ end
 
 indent(io::IO, n::Int) = print(io, " "^n)
 
+function to_graphviz_with_viewer(g::Graph; path::String="", kw...)
```
epatters, line 278: Thanks Matt! I would suggest calling this function `view_graphviz` since the input is already converted to a Graphviz graph.

src/graphics/Graphviz.jl
```diff
@@ -275,4 +275,18 @@ end
...
+    filepath = isempty(path) ? "$(tempname()).png" : path
+    open(filepath, "w") do io
+        run_graphviz(io, to_graphviz(g), format="png")
```
epatters, line 281: Relatedly, I think this should just be 

```julia
run_graphviz(io, g, format="png")
```

src/graphics/Graphviz.jl
```diff
@@ -275,4 +275,18 @@ end
...
+    end
+    if Sys.islinux()
+        run(`feh $filepath`, wait=false)
```
epatters, line 284: Perhaps there is no standard way to do this, but on my system it's `xdg-open`. I don't have `feh`.

epatters (changes requested):
Thanks Matt! A few comments below.

epatters (approved):
Looks great, thanks!

PR #949, opened 2024-10-24
Incorporating GATlab more deeply

kris-brown:
This is a big refactor - right now just a draft. The goal is to rely less on type dispatch and use explicit models from GATlab. This will ultimately address a number of persistent problems we've seen in Catlab over the years, including issues with empty lists and VarSets vs FinSets. It will still be convenient at times to use a Julia `abstract_type` to handle informal interfaces, but for things like different implementations of FinSets, different implementations of FinCats, etc. we will use a data structure with a `impl` field and a `model` field which controls how the implementation satisfies the desired interface.

It is a bit difficult to track the changes because this is also an appropriate time to restructure the files. I'll do my best to summarize them here (eventually). One idea I have is that the `CategoricalAlgebra` code that is category-agnostic could come first, and then subfolders related to **Set**-like categories and **C-Set**-like categories could come afterwards.

The goal is for only minor breaking changes, keeping the user-experience as unchanged as possible despite big changes under the hood.

The current status is that that (Fin)Sets/(Fin)Functions/FreeDiagrams/Limits have all been refactored to rely less on type dispatch. Next is FinCats, and after that it remains to be seen how much change the CSet code requires.


jpfairbanks:
@kris-brown what is the timeline around this? I could see how keeping this up to date with any Catlab changes is going to be a nightmare.

kris-brown:
I'm working full speed on it now, I expect to be ready for review next week (really depends on how many dependent packages on Catlab use implementation details that I wasn't expecting though)

jpfairbanks:
Ok, let's block the devs meeting next week to talk about this and you can have some kind of document that describes the changes that we can go through together. Rather than just using traditional code review. Because this is basically doing a systematic change across the whole codebase

kris-brown:
Comments from devs meeting:

- BasicSets could be a top-level folder (parallel to CategoricalAlgebra)
- We might be able to remove type parameters even from implementation structs (to make `@instance` work we'll need to put function calls in the signatures)

kris-brown:
An update before I go back and refactor an earlier part of this PR.

- I've gotten to C-Sets. Trying to make it work has made me realize I didn't go far enough in GATlabification earlier. For context, for various datatypes in Catlab, there've been three options for what to do: 1.) Make an explicit GAT for the type's interface. Allow there to be many difficult implementations of it and only interact with the type through its interface. For example, `ThFinSet`  (`FinSetInt` , `TabularSet`, `FinSetHash` ) and `ThCategory` (`OpCategory` , `TypeCat` ). 2.) Make an abstract type which has an informal interface. Managed via dispatch. For example:  `DiagramImpl` (`DiscreteDiagram` , `Span` , `BipartiteFreeDiagram` ) and `ColimitImpl` (`ColimitCocone` , `CompositePushout` , `ACSetColimit` ). 3.) Just have a struct (e.g. `Span` ).
- The problem I've encountered relates to computing (co)limits of ACSets. The problem arises from the fact that dispatch is used to address the fact that limits and colimits are handled in a category-specific way.
- If I have `limit(d::DiscreteDiagram, c::Category) = ...` , then the value of `c`  determines the implementation of what I'm supposed to do. However, there is no way to do this with just the official interface of a category. So, we illegally use the fact that every category has *some* implementation and write this method as `limit(d, c.impl)` . Once we do that, it's up to us to define limit for this pair (of `DiagramImpl` and `ThCategory` impl), otherwise face a `MethodNotFound` error.  For example, `struct SetC end` is a particular implementaion of `ThCategory` . There are many `limit` methods where the second parameter is `::SetC` . 
- The interface of an `ACSetCategory` is: 1.) the interface of a profunctors (i.e. two copies of `ThCategory` + heteromorphisms) and 2.) a means of extracting ob/hom/attr/attrtypes from an `ACSet`. This is meant to be very flexible because we might want to work flexibly with, e.g., **C-Set**, **C-Par**, **ACSet** (loose or tight), **VarACSet**, or **VarACSet** with a certain unusual notion of 'product'.
- For example, the vanilla ACSet category has for the domain category FinSets for `Ob` and FinFunctions for `Hom` . For heteromorphisms, SetFunctions for `Attr`. In the discrete codomain category, Julia types for `AttrTypes` . In Implementing `ThACSetCategory` , we might say `compose(f::Hom,g::hom) = compose[SetC()](f,g)` to reuse the code we already wrote for how to compose morphisms in **Set**.
- If we want to compute limits pointwise, we loop over `Ob` and `AttrType` and compute limits in either the domain or codomain category. How is one supposed to get the domain category from a `ThACSetCategory` implementation? Given that we've provided composition / id methods as part of providing that interface, we can make a `struct DomACSetCategory`  which wraps a `ACSetCategory` and show how we can implement ThCategory (this is also the sort of thing we could do via a pullback along a theory monomorphism). But, how do we know that this resulting category IS **Set**, the category which we defined limits and colimits for above?  Julia dispatch can't figure out how to use the `limit` methods defined for `SetC`. The fact that the domain category is **Set** is due to what the code of the implementation of `ThACSetCategory` is, which isn't transparent to Julia, especially at the type level.
- So what is the solution? A bandaid fix would be to change `ThACSetCategory`  to just have a `dom_cat()::Cat`  and `cod_cat()::Cat`  method, so that one can, e.g., plug in `SetC`  into the definition of `CSetCat` . 
- But to address the root of the problem we should never dispatch on an implementation. Rather we need, e.g. `ThCategoryLimits`. This is the sort of thing I wanted to wait on doing until we had [sugar for "structs" in GATlab officially supported](https://github.com/AlgebraicJulia/GATlab.jl/blob/main/src/nonstdlib/theories/Pushouts.jl). But it's hopefully not as bad as I'd thought if we use the trick that has been working out well where we have some GAT types be forced to be certain Julia types.

kris-brown:
Consider this GAT which is meant to abstract the common ways `DiscreteDiagram`, `ParallelHoms`, `ComposableHoms`, (Multi)(Co)Spans, `BipartiteFreeDiagram` and `FreeDiagram` are all used.

```julia 
"""
This is simpler than a `FinDomFunctor` because the domain is not a FinCat but
rather a finite Graph. We also don't worry about the category structure of the
codomain (e.g. homs need not specify dom/cod, no composition structure). 
To get a real diagram out of this, one must supply a Category with the
appropriate Ob/Hom types which are supertypes of the Ob/Hom types of the 
FreeDiagram.
"""
@theory ThFreeDiagram begin 
  V::TYPE; E::TYPE; Ob::TYPE; Hom::TYPE; FinSet::TYPE;
  src(h::E)::V;
  tgt(h::E)::V;
  obset()::FinSet
  homset()::FinSet
  obmap(v::V)::Ob
  hommap(e::E)::Hom
end
```

For example, a `Span{Ob,Hom}(f,g)` will always have `V -> FinSet(3), E -> FinSet(2)` with `src(::Int)=1` and `tgt(n)=n+1`. And `hommap(1) = f` and `hommap(2)=g`.

By not presupposing the category structure of the codomain of this diagram, this means we don't have to carry around a bunch of methods for how to do identities/composition -- this is good; the point of a FreeDiagram is that we can really just treat it like a piece of concrete *data*. We can simply say "here is a pair `f`+`g` I want to treat as a span" and defer anything that depends on what category we take `f` and `g` to live in until later, such as when we compute the pushout.

However, there is one wrinkle in this, which is that the `dom`/`codom` structure is also something that depends on what category we interpret `f` and `g` as living in (in this new refactor, that is no longer determined by what type `f` and `g` are). The problem is that, although `hommap` in the example above was determined by `f` and `g`, it's not the case that `obmap` is determined by `f` and `g`. The way Catlab currently handles this is effectively implementing `obmap(3)`  with `codom(tgt(hommap(1)))` and so on. But here it's clear that `src` is relying on type-dispatch.

An obvious situation where this is a problem is if we want to have a span in a category of 'bare' morphisms (where the morphisms don't carry full info about their domain and codomain). Or, in the case that brought this issue to my attention, the category `SkelFinSet` has `FinSetInt` as its objects and `FinFunctions` as its morphisms. 

> [!NOTE]  
> `struct FinSetInt i::Int end` is a particular struct which implements `ThFinSet`, whereas a `FinSet` is a wrapper for *any* stuct which implements `ThFinSet`.

So the domain of a `FinFunction` is a `FinSet` which is not a `FinSetInt` (neither a super nor subtype). There's no way for a span made up of two FinFunctions to know how to get the apex which needs to be a `FinSetInt`. 

It seems the easiest way around this issue would be to just add the extra data to the various structs, e.g.: 

```julia 
@struct_hash_equal struct Multispan{Ob,Hom,Legs<:AbstractVector{Hom}}
  apex::Ob
  legs::AbstractVector{Hom}
  feet::AbstractVector{Ob} # NEW: we add this data, which will sometimes be redundant
end
```

PR #954, opened 2025-01-22
Fix bugs in `Base.in` methods and empty comprehension

samuelsonric:
- The `in` functions were not really causing problems, since Julia was defaulting to iterating over the container. However, it was creating method ambiguities (I caught it with Aqua).
- In the function `make_homomorphism` the comprehension `[row[i+=1] for _ in parts(X,c)]` is creating a vector of type `Any` if `parts(X, c)` is empty.


epatters (approved):
Thanks Richard! I don't know how we got all those `Base.in` methods backward.

PR #959, opened 2025-02-07
Removing ADTs and Parser Exports Due to Too Much Internals Exposure

### LINKED ISSUE #958: RelationTerm/Parser Exports Removal
RelationTerm and Parser core elements are exported, however, this exposes too much of internal Catlab elements to the common user. These exports should be removed, and imports should be used when they are needed with Catlab source code.

cscaff:
This PR resolves #958. In my last Catlab PR, I added the ADTs and Parser modules to Catlab. The end goal was to provide the user with access to the relation string macro and AlgebraicJulia developers with PEG rules when they explicitly import them in other packages. However, many unneeded items were exported, leaving basic users with too much access to internals. This PR modifies this such that basic users can only access the Relation string macro with a typical ```Using Catlab```. Otherwise, they will need to explicitly call ```Using Catlab.Parsers``` to gain access to specific PEG-building constructs. Minor fixes were made on top of this. For instance, the "collect()" function within ParserCore.jl was renamed to create clarity between the ParserCore feature and the Base Julia function.


lukem12345:
@cscaff The GitHub actions should run fine now. It looks like you will want to add `PEG` to the `test/Project.toml` and reconfigure the exports some more.

src/wiring_diagrams/Algebras.jl
```diff
@@ -25,6 +25,11 @@ function oapply(composite::UndirectedWiringDiagram, hom_map::AbstractDict,
   oapply(composite, homs, obs; kwargs...)
 end
 
+(uwd::UndirectedWiringDiagram)(hom_map::AbstractDict,
```
epatters, line 28: I'm confused by this diff. Weren't these methods already added in #962? You might need to do a fresh rebase off main.

epatters (changes requested):

src/wiring_diagrams/Algebras.jl
cscaff: I think I messed up my rebase when I tried squashing my commits. I don't know how I missed all the new file changes. Sorry! Will rebase now.

cscaff: Commits should match now!

epatters (approved):
Thanks!

PR #963, opened 2025-03-05
Splitting and Joining Basic Schemas

quffaro:


TheCedarPrince:
Just for the conversation reference around this PR:

> Is it possible to modularize the creation of presentations? I am curious if I can have smaller defined chunks of an ACSet presentation such that I can work on a large ACSet in a piecewise manner. I think Sean did something like this with Individuals.jl if I am not mistaken? - @TheCedarPrince 

> we should have a way to break it down into more manageable parts. we can then patch them together. so similarly, we expect to parameterize the vertex of the edge-labeled graph representing relations between ACSets by Schema @quffaro 

Exciting developments! I wonder if this could be used within CatColab to have reusable ACSet presentation components - like building blocks? - in collaborative modeling.

slwu89:
This is pretty neat! Even though this is super draft work, I'll just make some random comments. Maybe you already know all this.

- Because the inheritance for presentations is seperate from the inheritance for acset types, you can do `@acset_type EdgeLabeledGraph(SchEdgeLabeledGraph) <: AbstractLabeledGraph` and then calling `to_graphviz(n, node_labels=:tablename, edge_labels=:edgelabel)` will let you get the nice graph below, for example: image
- I guess so every attribute doesn't get sent to the same attribute type when using `join`, you'd need another attribute in `SchTable` for `Column` giving the name of the `AttrType` in the joined schema it will be sent to?
- I think `using Catlab` brings in all modules and the entire `ACSets` library now, so it alone is sufficient.
- You can wrap the `BasicSchema` like `to_graphviz(Presentation(SchAnew))` to use all the Catlab methods on it

PR #974, opened 2025-09-29
Add Reflexive Edges to graph

quffaro:
Small PR which allows users to add reflexive edges to a graph

### DIFF
--- a/src/graphs/BasicGraphs.jl
+++ b/src/graphs/BasicGraphs.jl
@@ -379,18 +379,12 @@ end
 
 function convert(::Type{AbstractReflexiveGraph}, g::Graph)
   X = ReflexiveGraph()
-  vs = add_parts!(X, :V, nv(g))
-  for e in edges(g)
-    s, t = src(g, e), tgt(g, e)
-    e′ = add_part!(X, :E, src=s, tgt=t)
-    s != t && continue
-    # if the edge is reflexive and there is no entry for that vertex in `refl`, add it. If there is already a reflexive edge there, then the underlying graph is invalid.
-    refl(X, s) == 0 ? set_subpart!(X, s, :refl, e′) : error("Vertex $s has mulitple loops, and therefore cannot have a distinguished reflexive edge.")
-  end
-  # add the rest
-  refls = add_loops!(X; idempotent=true)
-  for e in refls
-    set_subpart!(X, src(X, e), :refl, e)
+  copy_parts!(X, g)
+  add_loops!(X; idempotent=true)
+  for v in vertices(X) 
+    e = incident(X, v, :src) ∩ incident(X, v, :tgt)
+    length(e) > 1 && error("Vertex $v has multiple loops, and therefore cannot have a distinguished reflexive edge.")
+    set_subpart!(X, v, :refl, only(e)) 
   end
   return X
 end


src/graphs/BasicGraphs.jl
```diff
@@ -159,6 +159,28 @@ function add_edges!(g::HasGraph, srcs::AbstractVector{Int},
   add_parts!(g, :E, n, (src=srcs, tgt=tgts, kw...))
 end
 
+""" Add reflexive edges to a graph
+
+Setting the keyword `idempotent::Bool=true` will skip adding reflexive edges to vertices which already have them
+"""
+function add_reflexives!(g::HasGraph; idempotent::Bool=false)
```
kris-brown, line 166: Should it throw an error (perhaps only if `idempotent=true`) and the input graph has *more* than one loop on a given vertex?

src/graphs/BasicGraphs.jl
```diff
@@ -159,6 +159,28 @@ function add_edges!(g::HasGraph, srcs::AbstractVector{Int},
...
+    if idempotent
+        out = Int[]
+        for v in vertices(g)
+            if isempty(incident(g, v, :src) ∩ incident(g, v, :tgt))
+                e′ = add_edge!(g, v, v)
```
kris-brown, line 171: to be consistent with the rest of this file, these tabs should be just two spaces

src/graphs/BasicGraphs.jl
```diff
@@ -346,6 +368,24 @@ function rem_vertices!(g::AbstractReflexiveGraph, vs; keep_edges::Bool=false)
   rem_parts!(g, :V, vs)
 end
 
+function convert(::Type{AbstractReflexiveGraph}, g::Graph)
+    X = ReflexiveGraph()
+    vs = add_parts!(X, :V, nv(g))
+    es = for e in edges(g)
+        let src = g[e, :src], tgt = g[e, :tgt]
```
kris-brown, line 375: wow never seen this syntax before! is the idea that you want to use `src`/`tgt` as variable names and need this to avoid conflict with the surrounding namespace?

kris-brown (commented):
Thanks Matt! Some small comments and one bigger picture point: in order for there to not be arbitrary behavior, I think `add_reflexives!` should throw an error if there is more than one loop on a given vertex, since otherwise `convert` will pick an edge at random (well, the 'last' edge in the arbitrary order) to make *the* reflexive edge. 

(If your existing use case would have issues with that, maybe we'd have to relax that.)

src/graphs/BasicGraphs.jl
quffaro: I agree, I've added an error in the `idempotent=true` case.

src/graphs/BasicGraphs.jl
```diff
@@ -346,6 +372,23 @@ function rem_vertices!(g::AbstractReflexiveGraph, vs; keep_edges::Bool=false)
   rem_parts!(g, :V, vs)
 end
 
+function convert(::Type{AbstractReflexiveGraph}, g::Graph)
+  X = ReflexiveGraph()
+  vs = add_parts!(X, :V, nv(g))
+  es = for e in edges(g)
```
kris-brown, line 378: sorry I don't follow what this block is doing. You don't seem to use `es`. Also why do you need to add anything in this block if `add_reflexives!` is doing all the edge-adding that you need?

src/graphs/BasicGraphs.jl
```diff
@@ -159,6 +159,32 @@ function add_edges!(g::HasGraph, srcs::AbstractVector{Int},
...
+  if idempotent
+    out = Int[]
+    for v in vertices(g)
+      es = incident(g, v, :src) ∩ incident(g, v, :tgt)
+      if isempty(es)
+        e′ = add_edge!(g, v, v)
+        push!(out, e′)
+      end
+      if length(es) > 1
+        error("Vertex $v has more than one reflexive loop. Set `idempotent=false` to bypass this error by allowing multiple reflexive edges")
```
kris-brown, line 176: Hm the issue isn't that there are multiple reflexive edges (there can only be one in a reflexive graph) but that it gets picked arbitrarily by `convert`. But the error message is here rather than in `convert` so you'd have to phrase it differently. Maybe just say "multiple loops" rather than multiple reflexive edges?

src/graphs/BasicGraphs.jl
kris-brown: Maybe also add some comment in the docstring about how this adds *loops* to a graph. These loops can be taken to be reflexive edges in any abstract reflexive graph canonically if there's exactly one loop per vertex (guaranteed by `idempotent=true`, at the cost of throwing an error if there's more than one).

src/graphs/BasicGraphs.jl
quffaro: Thanks for catching the `es`, that block used to be a `map` for some reason. The `for` loop is supposed to construct the graph `g` underlying the new reflexive graph, while also accounting for any existing loops.

I renamed `add_reflexives!` to `add_loops!` to relieve it from thinking about reflexive graphs, punting the `error` message in `add_loops!` to `convert.` If the underlying graph already has multiple loops, then its automatically invalid. `add_loops!` will just add the rest.

quffaro: I prefer that; it's more idiomatic to Catlab.

PR #986, opened 2025-10-29
`@acset_colim` variants: exposing name mapping, and morphisms 

kris-brown:
This draft introduces two new macros related to `@acset_colim`. The first, `@named_acset_colim`, takes in an extra parameter which will be assigned a NamedTuple relating the names of the generators to their respective elements (`Pair{Symbol, Int}`).

```julia 
acs = @named_acset_colim y_Graph MyDomain begin
  v::V; e::E; src(e)==v
end

@test acs[MyDomain.e[2],:src] == MyDomain.v[2]
```

The second `@acset_transformation_colim`, accepts two `@acset_colim`-style expressions as well as a sequence of pairs (which initializing a hom search) as well as keyword arguments for hom-search.

```julia
# sending path graph •→•→• to •→•↺↺. 
f = @acset_transformation_colim y_Graph #=domain=# begin
  (vstart,vend)::V; (e1,e2)::E; 
  src(e1)==vstart; tgt(e1)==src(e2); tgt(e2)==vend
end #=codomain =# begin
  v::V; (ea,eb,ec)::E; 
  src(ea)==v; tgt(ea)==src(eb) == tgt(eb) == src(ec) == tgt(ec)
end #=mapping=# begin vstart=>v; end #=kw=# (any=true, monic=[:E])
```

In order for this to truly work, the data of `yoneda` is actually not enough. We are missing the data of, in a representable `Foo`, which element is the representing element. For example, the representable edge in a reflexive graph has three edges in it. Only one of them is the representing element. Presently, for simplicity so that the same data can be plugged into `@acset_colim` and these macros, we throw an error if there is any ambiguity in what the representing element is.

### DIFF
--- a/src/categorical_algebra/pointwise/datamigrations/Yoneda.jl
+++ b/src/categorical_algebra/pointwise/datamigrations/Yoneda.jl
@@ -271,20 +271,18 @@ end
 
 """
 ```
-Dom = @named_acset_colim yGraph MyDomain begin
+names, Dom = @named_acset_colim yGraph begin
   v::V; e::E; src(e)==v
 end
 ```
 
-This would be have like `@acset_colim` and assign to `Dom` the walking edge 
+This would behave like `@acset_colim` and assign to `Dom` the walking edge 
 ACSet with src=1 and tgt=2. However, unlike `@acset_colim`, this also assigns 
-to a variable MyDomain the NamedTuple `(v=(:V, 1),e=(:E, 1))`.
+to a variable `names` the NamedTuple `(v=(:V, 1),e=(:E, 1))`.
 """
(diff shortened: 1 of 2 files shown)


wwaites:
Moving some discussion from the Zulip here. Having some experience with these new macros now, I think that it might be unwise to have `@named_acset_colimit` implemented the way it is because it is not referentially transparent -- the names appear effectively by side-effect. This more functional (and simpler!) implementation might be better :

```julia
macro named_acset_colim(yon, body)
  quote 
    colimit_representables($(parse_diagram_data(body, __module__)), $(esc(yon)))
  end
end
```

